Issue #8, opened 2025-05-19
Oracle abstraction for evaluation


HacksonClark:
Allow oracle definitions via polling, alert-based triggers, or user-defined logic. Support both generic (e.g., HTTP heartbeat) and custom (e.g., training convergence). Right now, we're also entirely dependent on the `submit` API to evaluate agents. The agent has to call that API, and the user has to implement the `eval` function in the problem class. I think we should abstract this into a separate Oracle so that we can support other kinds of evaluation (like alert based evaluation).

HacksonClark:
One note: we shouldn't remove the submit API, since. it is essential for tasks like anomaly detection.

HacksonClark:
I think we should have separate oracles for each task (detection, localization, etc).

HacksonClark:
An important note: I'm going to be completely changing how problems are structured. Mitigation problems will be evaluated end to end (detection, localization?, then mitigation). I'm thinking of removing RCA, I feel our current implementation just doesn't make a lot of sense, and mitigation and RCA are decoupled.

yimingsu01:
I think it's intrinsically hard to design an one-size-fits-all oracle abstraction for different types of application and problems.

yimingsu01:
For the classic problems, like detection localization mitigation, oracle abstractions are rather mature I think.

HacksonClark:
@yimingsu01 I think the custom oracles will only be used for mitigation. Either using a custom function or alerts or something else. I'm not super dead set on this so we can discuss more.

yimingsu01:
I agree

yimingsu01:
I think the oracle should do a continuous evaluation on the agent. E.g., in a mitigation task, we can continuously score the agent to see how close it gets to the correct mitigation action

yimingsu01:
This is especially important when we start to do creative stuff, such as chaining multiple faults together

HacksonClark:
@yimingsu01 that's the current plan! I'm using a polling style Oracle that will evaluate the mitigation tasks.

yimingsu01:
I'm SUPER excited to design new problems. 

I think what will really set us apart is to have challenging, high-fidelity, and a wide-range of problems. 

Stuff like:
1. chaining faults
2. designing problems that are reasonable (i.e., will happen in real-life) with real-world systems (e.g, zookeeper) and non-symptomatic faults (e.g,. bad IO from khaos)
3. problems that touch different aspects of the system: networking, configuration, management operations
4. etc
[...]

Issue #4, opened 2025-05-19
Remove ACI


HacksonClark:
The agent-cloud interface needs to be removed, can be replaced by #3. The benchmark shouldn't force any particular tools and should be decoupled. With MCP, they could use our tools if they chose, but we shouldn't be coupled to things like get_action. AIOpsLab *technically* didn't force this since you could just use other tools in addition to the provided ones, but I think the ACI is just not worthwhile at this point. We shouldn't be reliant on `get_action` or any AIOpsLab functions for testing agents.

HacksonClark:
This is dependent on #8 as we can't fully remove the ACI until we have the new oracles.

Issue #1, opened 2025-05-19
Renaming


HacksonClark:
This is a hard fork of AIOpsLab, so we still have all the old assets, so we should go change the name everywhere from AIOpsLab to OpsArena. Also, we should remove all the Microsoft copywrite. This is an xlab project now :)

Also, I want to rename the orchestrator to 'conductor' since the orchestrator is a big part of the AIOpsLab paper and I want us to differentiate from it.

tianyin:
Indeed very good first issue for whoever wants to help.

HacksonClark:
I will say, I don't like the term agentops at all anymore since there's a large open-source project called agentops already: https://www.agentops.ai/

tianyin:
lol change to whatever you want. "Ops" is too genetic. perhaps say SRE-arena?

HacksonClark:
@tianyin Yeah, I was thinking SREArena as well.

yimingsu01:
That sounds like SREBench a lot, which exists already

HacksonClark:
@yimingsu01 I think that's kinda the point since we're scooping.

There's a couple SREBench projects around but they're all trash anyways.

yimingsu01:
ok

iamyeezuschrist:
>>> Also, we should remove all the Microsoft copywrite.

am i right in thinking files like this shall just be rm -rfed? like code of conduct//contributing.md? 

❯ cat CODE_OF_CONDUCT.md
# Microsoft Open Source Code of Conduct

thought it doesn't really make sense to keep microsoft rudiments, yet probably you'd like some versions of this files for the repo?

HacksonClark:
@iamyeezuschrist Yep, you're right. For things like CONTRIBUTING we can just remove the MS stuff or delete it entirely.

tianyin:
@iamyeezuschrist yeah we should remove the Microsoft stuff and just put a note somewhere, at least before we open the repo to public.

But we literally don't have anything at this moment, so it's not the most urgent item.

PR #26, opened 2025-05-27
Add new oracles

HacksonClark:
This change is massive, to add the new oracles we need to completely change how problems are created and ran. I will update later with the exact implementation.

Basically, the new style of evaluation is that problems will not be separated by type (detection, mitigation). The problem is responsible just for injecting the fault.

Every problem will be ran twice, once faulty and no-op once. The first thing to be evaluated is anomaly detection, if the agent answers correctly it will proceed to localization. After localization, it will try mitigation. However, it will try mitigation regardless of localization as mitigation is not dependent on localization.

For evaluation, we abtract the `eval` functions into oracles. So, there will be a detection oracle, a localization oracle, and a mitigation oracle (to start, they mitigation oracle will just use eval, but in the future we want to support different ones like alerts).


HacksonClark:
@yimingsu01 Don't approve yet, I haven't done the bulk of my changes. This is just a draft PR that I will keep pushing to.

HacksonClark:
This will close #8, #1, and #4

HacksonClark:
This PR is massive, it closes 3 issues and completely changes the way the code is structured. It might make sense to meet and review. Let me know if there are any questions..

yimingsu01:
Ok, 475 files, here we go

HacksonClark:
You definitely don't need to read all the files. If you want I could just meet with you and explain the new changes.

________________________________

yimingsu01:
@HacksonClark It's no worries, also my chance to learn. I forgot to tell you that you don't need to port the baseline agents: they have to be re-written to communicate to the MCP servers anyway.

PR #27, opened 2025-05-27
Added example MCP server and client

yimingsu01:
This PR adds an MCP server and client demo.

To run this demo, run this command:
```bash
python3 ./mcp-server/test_client.py ./mcp-server/test_server.py
```

This runs the MCP server and the client locally. They communicate through stdio. 

Make sure you have a working `.env` file with these values:
```
### Tools ###
PROVIDER_TOOLS=""
MODEL_TOOLS=""
URL_TOOLS=""
API_VERSION_TOOLS=""
API_KEY_TOOLS="<api_key>"
```

The demo is tested with openai gpt-4o. We are using `litellm` so it should also support other models. Please test with other models to see if tool calling is different.

This demo mainly demonstrates the tooling ability of an MCP server. You can ask the llm `what tools do you have` and it should list all the tools it has access to. Currently, I only implemented the `surround(character, main_body)` tool, which surround a text with some characters. You can see how the LLM returns a function calling response, then the client invokes the MCP server endpoint.

I'm moving my focus to trying out `langgraph` as the next agent framework to better figure out how to use the `resource` and `prompt` features in the MCP server. @HacksonClark please don't delete this branch yet.

### DIFF
--- a/mcp-server/test_client.py
+++ b/mcp-server/test_client.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import json
+import sys
 from contextlib import AsyncExitStack
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


HacksonClark:
Looks great, I'll review it tomorrow most likely.

PR #38, opened 2025-06-02
Update workload duration and workload type

### LINKED ISSUE #37: Jaeger does not show all available traces/services
@HacksonClark 
During testing MCP server, we found that Jaeger might not show all available services and traces when a problem is just freshly started. Though this is transient. We need to look deeper into the workload generator and Jaeger.

HacksonClark:
Closes #37


HacksonClark:
@daklqw Check out what I did in this PR, which the duration change won't matter (since we're switching to a constant load generator) I think we should use the mixed workloads so that the jaeger traces cover the full topology of the services.

Issue #65, opened 2025-06-03
Randomly select targeted service


HacksonClark:
This builds on #14 so read that issue as well.

Basically, for some faults we can apply them to any service (like target port misconfiguration). However, we currently only select one service to inject the fault on. However, it would be great if we randomly selected a service so the fault manifestation isn't the same every single time.

hieue:
@hieue

hieue:
@HacksonClark About this problem for some of the generalized problem we can do this across app as well, do you think it's worth implementing that?

HacksonClark:
@hieue Yes, that would be awesome!

Saadmrp1038:
Hi @HacksonClark. @hieue and I chatted a bit. Wanted a little clarification on one thing. For generalized faults (not service dependent) targeting services that are not called by the workload may result in a scenario where the `MitigationOracle` fails but the `WorkloadOracle` passes. Is this an acceptable scenario when doing random service selection?

---
#### readiness_probe_misconfig

App: `astronomy-shop`
Service: `frontend`

```bash
== Mitigation Evaluation ==
⚠️ Container frontend is not ready
[❌] Workload entry at 1751789071.149572 failed with log: {"current_response_time_percentiles":{"response_time_percentile_0.5":7,"response_time_percentile_0.9...
```

App: `astronomy-shop`
Service: `jaeger`

```bash
== Mitigation Evaluation ==
⚠️ Container jaeger is not ready
[✅] Successfully collected 2 workload entries.
```

HacksonClark:
@Saadmrp1038 I think the workload oracle should only be used *additionally* to a mitigation oracle not instead of it, as the mitigation oracle can tell us if the root cause was really fixed.

Saadmrp1038:
Understood, I was giving too much importance to the `WorkloadOracle`.

hieue:
I think I'm gonna try a different approach. The current one's a little too disruptive to our current design.

hieue:
I'll wait on #182 to be merged before retrying this issue.

HacksonClark:
This can be part of #191

Issue #63, opened 2025-06-03
Resource Requests Too Small


HacksonClark:
Do the opposite of #56

Issue #56, opened 2025-06-03
Resource Requests Too Large


HacksonClark:
- Fault: Set CPU/memory requests that no node can satisfy.

Example:
    ```yaml
    resources:
      requests:
        memory: "64Gi"
    ```
    
- Effect: Stuck `Pending`, error like:
    
    ```
    0/3 nodes available: Insufficient memory
    ```
    
- Fix: Lower the resource requests.

Issue #42, opened 2025-06-03
Missing Service Definition


HacksonClark:
- Fault: Remove the `Service` object entirely for a targeted service.
- Effect: Other services can't resolve the target via DNS (e.g., `cartservice` cannot reach `redis`).
- Fix: Recreate the `Service` with correct selectors and ports.

PR #93, opened 2025-06-10
Add #42, #56, #63

daklqw:
- #42
- #56 
- #63

Problems:
- Localization: I let `faulty_service` be an argument. But the faulty components obviously are not only `[faulty_service]`. 
  - Some problems like `revoke_auth_mongodb-1` and `storage_user_unregistered-1` also have such problem.
  - Also we need #65.
  - Need futher discuss on it. (Like auto-detect faulty pods and add them to the answer list)
- In `inject_virtual.py:237`, I used a `sleep(30)`, which is in the same way as `redeploy_without_PV`.


HacksonClark:
Do you want to work on #65 next?

Some problems already have faulty_service as a variable, also faulty_service is a list so we can pass multiple values there.

ChuanweiQu:
Indeed awesome! 👍

PR #97, opened 2025-06-11
Optimize noop evaluation

### LINKED ISSUE #68: Move noop out of problem registry
The noop (no operation, we do nothing). problems are currently in the problem registry. However, we now run the noop problems every time after we finish running a faulty problem (to ensure good anomaly detection coverage).

So, we don't want noop in the normal problem registry. We should move them out of registry, but they still need to run from the conductor.

### LINKED ISSUE #72: Optimize noop evaluation
When running a problem, we currently run the problem (with the fault) and then run the noop evaluation afterwards. We tear down and redeploy the application for both problems, which takes time. 

However, it makes more sense to run noop evaluation first (since there is no fault), then we can proceed with the faulty problem without having to tear it down.

HacksonClark:
Closes #72 

It also closes #68 as no_op is no longer its own problem, but is evaluated along side the problem.

Basically, when you start a problem here is the workflow:

Setup infra -> Setup application -> Evaluate noop (no fault is injected yet) -> inject fault -> evaluate fault problem.


HacksonClark:
@daklqw Are you able to approve PRs?

PR #104, opened 2025-06-13
Agent text editing interface

### LINKED ISSUE #88: Text Editing Agent Interface
#85 needs a text editing agent interface to enable the agent to debug programs.

Previous work SWE-Agent already have a text editing interface, this issue ports them to MCP servers and `langgraph` tools.

yimingsu01:
This PR closes #88. 

This PR provides a basic text editing interface for the agent to prepare for the upcoming software fault problems, where the agent needs to debug a buggy application and redeploy it. 

We have these text editing tools:
1. `window_file`: this is not a tool but the file system interface that the tools rely on. It allows the tools to open a file and only show a window (10 lines) of a file to avoid overwhelming the agent context. 
2. `open_file`: opens a file with a path and line number. The opened file is recorded in the agent state for other editing tools.
3. `goto_line`: go to a specific line in the file and prints the file content with a 10-line window.
4. `create`: creates a file with a given path.
5. `edit`: search for a string and replace occurrences of it.
6. `insert`: inserts a string at a given location, or at the end of the file.

They serve as the most simple way the agent can interact with the tool. Currently, they are implemented as `langgraph` tools, which implies that they are still local tools. The next PR will be implementing a "compile & deploy" tool that allows an application to be compiled and deployed, then the tools will be converted into MCP tools and deployed with the benchmark.

### DIFF
--- a/tests/file_editing/test_edit_1.yaml
+++ b/tests/file_editing/test_edit_1.yaml
@@ -1,0 +1,13 @@
+tool_calls:
+  - name: "open_file"
+    path: "tests/file_editing/example.txt"
+    line_number: "1"
+  - name: "edit"
+    search: "hello world"
+    replace: "world hello world"
+user_inputs:
+  - "open tests/file_editing/example.txt at line 1"
+  - "replace hello world with world hello world"
+expected_curr_file: "tests/file_editing/example.txt"
+expected_curr_line: "1"
+expected_output: "world hello world"


HacksonClark:
How should I test this PR?

yimingsu01:
You can start the agent and ask the agent to use these tools to modify a file on your machine, such as:
```
open /abs/path/to/file.py at line 1
use the edit tool to search for "foo" and replace with "bar"
goto line 100
```

HacksonClark:
@yimingsu01 is it possible to have a `cli.py` for the agent tools? I feel like there has to be something in the community already for that.

Either way I'll test with the agent.

yimingsu01:
@HacksonClark Yeah, it's the `k8s_agent.py` at the moment. It allows you to put in NL query for the tools.

yimingsu01:
@HacksonClark You can also go to `tests/test_file_editing_tool.py` and uncomment some of the tests to run, this asks you to have a valid openai key in your .env. Some tests will fail, but it doesn't mean the tool is not working, but that the test is too strict or not setup well.

HacksonClark:
@yimingsu01 I mean a tool where we can call the functions ourselves without any LLM.

yimingsu01:
@HacksonClark you would have to compose the LLM response by yourself, which is quite tedious

HacksonClark:
@yimingsu01 sounds like that could be an interesting tool🤔

yimingsu01:
Yeah we can make it into a tool, I just didn't do it for this PR.

yimingsu01:
@HacksonClark After battling one of the most elusive bugs I have ever encountered for 4 hours, I added some reproducible tests. Run them by running this at the root of the repo:
```bash
pytest tests/file_editing/test_file_editing_tool.py
```

No LLM call or keys required.

HacksonClark:
That sounds awesome! I'm curious to hear about this bug.

yimingsu01:
TLDR: if you forge two LLM responses with the same response ID, `langgraph` silently overwrites the first one with the other, so that it looks to you there are two LLM responses, but there really is only one. No error messages whatsoever.

yimingsu01:
@HacksonClark Added some more tests. Feel free to run them.

Issue #64, opened 2025-06-03
Cleanup on exit


HacksonClark:
We have a function implemented in `conductor.py` so that when the user Ctrl+C's or the application exits, the problem "cleanup" code still runs (deleting the applications, openebs, etc). It's currently a TODO in https://github.com/xlab-uiuc/SREArena/blob/main/srearena/conductor/conductor.py

```python
def exit_cleanup_fault(prob):
    print("Recovering fault before exit...")
    prob.recover_fault()
    # TODO: Clean up everything else too
```

hieue:
@hieue

HacksonClark:
@hieue Make sure you assign yourself to the issue, also ping me and/or Yiming not just yourself.

HacksonClark:
Closed by #89

PR #121, opened 2025-06-17
[#64] handle sigint while in user input loop

hieue:
Added additional logic to cleanup when CTRL+C or "exit" when in user input loop.

https://github.com/xlab-uiuc/SREArena/issues/64


HacksonClark:
This works great! Just need to add the config example file back and then I'll approve and merge.

hieue:
Done!

PR #124, opened 2025-06-19
[#65] Randomly targeting service

hieue:
- Created Randomizer class to randomly select and instantiate apps, as well as polling for available services and choosing one to target for fault injection.
- Rewrite problem base class in order for `select_service` to work (services can only be polled after the app is deployed and stable).
- Applied Randomizer to generalizable problems. Non-generalizable problems (such as those targeting mongodb services) are kept the same, I only rewrote them to keep consistency with the base class.

Resolves [#65]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -749,3 +749,4 @@ context.json
 # Ignore provisioner ssh keys
 provisioner_ssh_key
 provisioner_ssh_key.pub
+agent_graph.png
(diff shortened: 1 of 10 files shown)


hieue:
Hey @HacksonClark, this is my idea for randomly targeting a service. Is this the correct approach or should I hardcode a list of services for relating to each app?

HacksonClark:
@hieue This approach is reasonable.

PR #141, opened 2025-06-28
Integrate Khaos

### LINKED ISSUE #2: Integrate Khaos
We need to integrate [khaos](https://github.com/xlab-uiuc/khaos) (the eBPF fault injector project) into SREArena. I think it makes the most sense to deploy it as a DaemonSet so it can be easily deployed on all the nodes in the cluster. However, khaos isn't ready to be a DaemonSet yet, so we will be dependent on [this issue](https://github.com/xlab-uiuc/khaos/issues/2) getting fixed first.

We also will need the web server issue from khaos to be done as well.

### LINKED ISSUE #19: Node Interface
In order to integrate [khaos](https://github.com/xlab-uiuc/khaos) we need an interface for SREArena to be able to communicate with the nodes to perform fault injection on them. I think as part of the push-button project #16 we will need to deploy khaos on all the nodes after provisioning.

HacksonClark:
Closes #2 
Closes #19

### DIFF
--- a/srearena/conductor/problems/registry.py
+++ b/srearena/conductor/problems/registry.py
@@ -6,6 +6,12 @@ from srearena.conductor.problems.auth_miss_mongodb import MongoDBAuthMissing
 from srearena.conductor.problems.cart_service_failure import CartServiceFailure
 from srearena.conductor.problems.configmap_drift import ConfigMapDrift
 from srearena.conductor.problems.container_kill import ChaosMeshContainerKill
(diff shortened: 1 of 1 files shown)


HacksonClark:
I only have one problem drafted as an example just to demonstrate that khaos is launching. We will need to do some testing with each fault to see what kinds of failure we can cause.

Also, I think I'll refactor khaos to work on a broader level than a given pid, or maybe we should just inject into all services on a node instead of just a particular service. But the point of this PR was to just get it running.

Issue #58, opened 2025-06-03
Rolling Update Misconfigured


HacksonClark:
- Fault: Set `maxUnavailable: 100%` with no readiness probe.
    
- Effect: All pods get killed before replacement is ready → outage.
    
- Fix: Add readiness probe or lower `maxUnavailable`.

HacksonClark:
Closed by https://github.com/SREGym/SREGym/pull/164

PR #164, opened 2025-07-09
implement rolling update misconfiguration fault

DayyanALi:
#58

### DIFF
--- a/srearena/conductor/oracles/rolling_update_misconfiguration_mitigation.py
+++ b/srearena/conductor/oracles/rolling_update_misconfiguration_mitigation.py
@@ -1,4 +1,6 @@
+import json
 import yaml
+import tempfile
 from srearena.conductor.oracles.base import Oracle
 
 class RollingUpdateMitigationOracle(Oracle):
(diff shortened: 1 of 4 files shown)


DayyanALi:
I have implemented the fault by setting maxUnavailable to 100% in the rolling update strategy of the deployment, without including a readiness probe. During rollout, this configuration causes a brief period where no pods are ready — resulting in a temporary outage.

Additionally, I have implemented a mitigation oracle that checks whether the deployment has been updated to include either a readiness probe or a safer rolling update configuration (i.e., maxUnavailable set to less than 100%).

However, I noticed that the service seems to recover almost immediately after the rollout, and I’m unsure if that still qualifies as a proper demonstration of the outage. I'd appreciate clarification on whether this behavior is acceptable or if further delay on the outage is needed.

HacksonClark:
@DayyanALi Hmm, the idea is always that the fault persists. Otherwise, it's very difficult for the agent to do anything as by the time it observes the system the fault has gone away and the system is fine. I'll take a look at this problem and write out some ideas I may have on how to get it to be persistent.

DayyanALi:
Hmm right so initially i was thinking I can deploy a custom service that sleeps for some period before becoming ready. This would simulate the effect of an outage. This may cause the replacement pod takes time to become ready, allowing the fault to persist long enough. Let me know if this approach sounds good to you.

HacksonClark:
@DayyanALi That sounds great. Is there a way where the deployment issue could persist permanently until fixed?

DayyanALi:
@HacksonClark I made the container sleep for 100 seconds using the postStart lifecycle hook. This causes the new pods to get stuck in the ContainerCreating state during a rollout, which introduces a noticeable delay and simulates a temporary outage effectively.
Initially, I used sleep infinity to make the fault persist permanently. However, I noticed that when attempting to delete the deployment or pods, they would get stuck in the Terminating state because the container never exited or responded to termination signals. To avoid this cleanup issue I changed the sleep to 100 seconds. Let me know if this approach looks good to you, or if you have any alternative suggestions.

HacksonClark:
@DayyanALi Why can't we edit to to terminate as well so it doesn't get stuck?

DayyanALi:
That might be because using sleep creates a process that doesn't handle termination signals, making it unresponsive to Kubernetes shutdown requests. I can try an alternative approache like using signal traps.

HacksonClark:
@DayyanALi That sounds great!

DayyanALi:
@HacksonClark I’ve addressed the issue by using initContainer. Also i made some changes in its evaluation oracle in which it checks if at least one pod remains running during a rollout, then the check passes. If all pods drop out, it fails. Hopefully this works fine.

PR #165, opened 2025-07-09
I98: Clean up cli.py 

### LINKED ISSUE #98: Clean up cli.py
`cli.py` still has some jank. For instance, after every action you take the environment replies with "please take the next action" which is a holdover from AIOpsLab orchestrator where it would "guide" the agent.

Also, there's autocomplete functionality for the problem selection. However, it persists after you start a problem which is very annoying when trying to run commands. It should go away after the `start` command is ran.

Saadmrp1038:
Resolves #98

### DIFF
--- a/srearena/service/shell.py
+++ b/srearena/service/shell.py
@@ -48,7 +48,6 @@ class Shell:
                 return error_message
             else:
                 output_message = out.stdout.decode("utf-8")
-                # print(output_message)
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
@HacksonClark can u check now?

HacksonClark:
@Saadmrp1038 Will do!

PR #166, opened 2025-07-09
[#140, #126] Add deploy option in cli.py and fix console double printing

### LINKED ISSUE #126: Console double printing
The console currently double prints the results of any command executed on the shell. 

Image

### LINKED ISSUE #140: Add `deploy` function to the cli
One useful feature of SREArena outside the benchmark itself is that we've made the popular benchmark microservice projects (DeathStarBench, TrainTicket, etc) super easy to deploy. Many other folks in the systems community use these microservices for more than just SRE related tasks.

I figure we can add a deploy function to cli (much like start) where instead of starting a problem it deploys a selected application.

hieue:
Resolves #140, resolves #126 

# Contributions

- Rewrote `Conductor` to expose `deploy_app` and `undeploy_app`, which are then called in `cli.py`. User can deploy an app by typing `deploy <app_name>` or `undeploy <app_name>`. Available apps are compiled in `AppRegistry`.
- Added `list` option to list apps currently deployed on SREArena.
- Updated `WordCompleter` to include options to deploy, undeploy, list, options and exit.
- Fixed double printing which was caused by a simple print statement inside `Shell.exec`.

### DIFF
--- a/srearena/service/apps/registry.py
+++ b/srearena/service/apps/registry.py
@@ -14,7 +14,7 @@ class AppRegistry:
     def __init__(self):
         self.APP_REGISTRY = {
             "Astronomy Shop": AstronomyShop,
-            "Flight Ticket": FlightTicket,
+            # "Flight Ticket": FlightTicket,
             "Hotel Reservation": HotelReservation,
             "Social Network": SocialNetwork,
             # "Train Ticket": TrainTicket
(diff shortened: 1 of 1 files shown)


HacksonClark:
This is a great start! But it isn't quite what I had in mind. I don't want the command line to pop up after deploying the application. We only want to deploy the application.

Since we have a deploy button, we should also have a `undeploy` function that the user can call through cli.py to remove the app when they are finished.

hieue:
Ah I see, should've clarified with you beforehand. Moreover so I understand that since you can de-deploy the app do you want the user to be able to start another app/problem while the current app is deployed?

HacksonClark:
@hieue Yeah, it basically just uses the deployment functionally in SREArena *and nothing else*. These applications are common in systems research, so we can make this into an easy tool for other researchers who use these applications.

hieue:
@HacksonClark And would you want the apps to be shut down once SREArena is exited? Or should they persist over to the next session and the user can undeploy the app whenever?

HacksonClark:
@hieue they should persist. Once they press deploy, we run out deployment functions then exit, leaving the application running.

PR #184, opened 2025-07-24
Fix chaos mesh issues related to removing duration

### LINKED ISSUE #183: Chaos mesh having argument issues
```TypeError: SymptomFaultInjector.inject_pod_kill() takes 2 positional arguments but 3 were given```

This is because I removed duration from the chaos mesh problems (since we want them to be persistent). The code error is fixed, but for `pod_kill` it seems to not be working. Currently investigating this and will push a PR that fixes all these issues.

HacksonClark:
Closes #183

### DIFF
--- a/srearena/conductor/problems/network_loss.py
+++ b/srearena/conductor/problems/network_loss.py
@@ -28,11 +28,7 @@ class ChaosMeshNetworkLoss(Problem):
     @mark_fault_injected
     def inject_fault(self):
         print("== Fault Injection ==")
-        self.injector._inject(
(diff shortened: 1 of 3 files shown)


HacksonClark:
@yimingsu01 The bug is fixed. However, I thought there was an issue with certain faults firing, namely pod_kill. Pod kill only kills the pod one time, and when k8s redeploys it works just fine. I'm debating just removing this problem since I feel like we're not really measuring anything with it..

yimingsu01:
I can just disable that problem for now.

PR #192, opened 2025-07-26
feat: Reusing resources

### LINKED ISSUE #128: Reuse Resources
We can save a lot of time on our benchmark runs if we can reuse OpenEBS and Prometheus in between runs. But we need to figure out how to effectively "clean" data from both so that it is not persisted between problem runs. Then, we don't have to redeploy for each problem.

saleha-muzammil:
Resolves #128

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,3 +16,5 @@ SMTP_PORT="587"
 SMTP_USERNAME="your.email@gmail.com"
 SMTP_PASSWORD="your-app-password"
 
+REUSE_PROMETHEUS=true
+REUSE_OPENEBS=true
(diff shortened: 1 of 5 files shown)


HacksonClark:
This PR will have to be updated when #203 is merged as the logic has changed a bit.

HacksonClark:
Basically, once that PR is merged you can put this cleanup logic in the `undeploy_app` method and things should work!

HacksonClark:
I'm going to close this, we will be switching how we're deploying the apps so that way they're all deployed at once. We might borrow some code from here. Thanks for the great work @saleha-muzammil!

PR #195, opened 2025-07-30
[#57] Persistent Volume Affinity Violation 

### LINKED ISSUE #57: Persistent Volume Affinity Violation
- Fault: PVC bound to node A, pod gets scheduled to node B.
    
- Effect: Pod stuck due to **volume node affinity mismatch**.
    
- Fix: Either delete/rebind PVC, or constrain pod to node A.

hieue:
Resolves #57 

# Contributions
- Create a PV `temp-pv` with nodeAffinity bound to node 1, then create a PVC `temp-pvc` bound to that PV.
- Set targeted service's `nodeSelector` to node 2 and add the PVC to the service's volumes.

# Details
- Because the PVC needs to be mounted but the pod can't access the PV as it's located on a different node, the pod ends up stuck in `NodeAffinity` status (or more like pods because the controller is continuously spinning up new pods but failing because of the pod affinity violation). 
- The fix is to simply patch the serivce's deployment's `nodeSelector` to node 1, after which all pods will be running.
```
SREArena> kubectl patch deployment user -n test-hotel-reservation \
  --type=merge \
  --patch '{"spec": {"template": {"spec": {"nodeSelector": {"kubernetes.io/hostname": "<node name>"}}}}}'

SREArena> kubectl rollout restart deployment user -n test-hotel-reservation
```
(Note: node name is the one as appears in the `kubectl get pods`)

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -12,10 +12,10 @@ from rich.markdown import Markdown
 from rich.panel import Panel
 
 from srearena.conductor import Conductor, exit_cleanup_fault
-from srearena.service.shell import Shell
-from srearena.service.kubectl import KubeCtl
-from srearena.service.telemetry.prometheus import Prometheus
 from srearena.service.apps.registry import AppRegistry
+from srearena.service.kubectl import KubeCtl
+from srearena.service.shell import Shell
+from srearena.service.telemetry.prometheus import Prometheus
 from srearena.utils.sigint_aware_section import SigintAwareSection
 
 WELCOME = """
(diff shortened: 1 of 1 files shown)


HacksonClark:
@hieue seems you forgot to push the problem class. I added one, it seems like you were initially targeting hotel-reservation, I used social network and it seems to work, but somehow user service on the hotel reservation did not. But please update with what you have then this is ready to merge.

hieue:
Hey @HacksonClark, I tried to make it more flexible by allowing you to type in what app you want to run it on. I think it didn't work on hotel reservation because it's called "user" there and "user-service" on social network. But generally I think for this fault any service will work.

HacksonClark:
Please setup precommit btw

HacksonClark:
> Hey @HacksonClark, I tried to make it more flexible by allowing you to type in what app you want to run it on. I think it didn't work on hotel reservation because it's called "user" there and "user-service" on social network. But generally I think for this fault any service will work.

I adjusted for this when I tried hotel reservation and it still didn't work.

PR #198, opened 2025-08-03
fix the bug in problem TaintNoToleration

Jackcuii:
When I labeled the TaintNoToleration problem, I found that the implementation of the problem has some faults.
1. the codes use 'control-plane' or 'master' in the `metadata.name` to identify the control node (to select a non-control-plane node). But when deploying the problem on the remote cluster, the node name is actually like `node0.testproblems2.*********-pg0.wisc.cloudlab.us`. It makes our codes actually taint the control-plane node. So using the label to select is a better choice.
2. When we use a cluster having more than 2 nodes, the pod can be assigned to a third node so that the availability of the system will not be affected. SREArena seems to **assume** that the cluster **always** encounters failure after injecting the problem, but actually not. (expected detection result is always 'Yes', and the Agent can actually get a credit by do nothing in this testcase.) It is indeed a flaw in the config, but I don't think the agents are able to detect the potential flaws up till now.

**This PR fixes the problem 1 above.**

### DIFF
--- a/srearena/conductor/problems/taint_no_toleration.py
+++ b/srearena/conductor/problems/taint_no_toleration.py
@@ -15,8 +15,8 @@ class TaintNoToleration(Problem):
         self.namespace = self.app.namespace
         self.kubectl = KubeCtl()
 
-        # ── pick a real worker node dynamically ─────────────────────────
-        self.faulty_node = self._pick_worker_node()
+        # ── pick all real worker nodes dynamically ───────────────────────
+        self.faulty_nodes = self._pick_worker_nodes()
         self.faulty_service = "user-service"
 
         super().__init__(app=self.app, namespace=self.namespace)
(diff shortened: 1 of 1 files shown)


yimingsu01:
@Jackcuii Interesting observation Bohan. Can you write a list of steps to reproduce the bug?

Jackcuii:
> @Jackcuii Interesting observation Bohan. Can you write a list of steps to reproduce the bug?

Sure~

Jackcuii:
Hi @yimingsu01 @HacksonClark 
1. We can use the remote cluster on cloudlab (e.g. with 3 nodes), so that the node name will be the domain name.
2. then `start taint_no_toleration_social_network`
3. step to the Detection step. The fault is injected to the cluster now.
4. then run ` kubectl get node -n test-social-network -o yaml`. we can find the taint is on the control-plane node. (usually it is node0) so that the `user-service` (default faulty service) runs well. (Problem 1)
5. When fixing this with this PR, run the same Step 1-4, we taint the node1. But we can find the `user-service` is assigned to the node2 and also runs well (it should be run well anyway).
We can get the result `'Detection': {'accuracy': 0.0, 'success': False}` for all the detection answer are set to `Yes`. So it's wrong here. Maybe we should taint all the worker nodes. (Problem 2)

HacksonClark:
I'll take a look tomorrow. I think I know what the issue is.

HacksonClark:
@Jackcuii Great find! I updated it to taint all the worker nodes and the fault is working.

PR #199, opened 2025-08-04
Add TiDB Application

### LINKED ISSUE #142: TiDB Application
We have some interesting misoperation problems that rely on TiDB, however, we don't have an application that uses TiDB, so, our misoperation problems won't really work.

We could either (1) create a small application that uses TiDB, or replace a database component from one of our current applications with TiDB.

For the application, it can be as simple as a Python web server that just queries the DB and returns different results. The database can be populated with whatever you'd like. We also need a load generator that automatically sends requests to the web server to simulate traffic. It should be instrumented for logs, traces, and metrics.

I would recommend starting with a simple Python app that can be containerized and deployed on Kubernetes that uses TiDB. Once we have that working, we can inject our misoperation faults on that application.

### LINKED ISSUE #196: Incorporate the TIDB Faults
Currently, [line 210](https://github.com/xlab-uiuc/SREArena/blob/5f010115f0f6744a7d534d2bb411ab50ba8d4f9d/srearena/conductor/problems/registry.py#L210) to [line 219](https://github.com/xlab-uiuc/SREArena/blob/5f010115f0f6744a7d534d2bb411ab50ba8d4f9d/srearena/conductor/problems/registry.py#L219) are commented out as we didn't have an app to consume the TiDB data. Now @lilygn made an app to consume TiDB data so we can incorporate the problems back in.

This issue tracks the progress of merging those misoperation problems back into the benchmark.

HacksonClark:
Adds liily's TiDB application and TiDB faults

Closes #196
Closes #142


HacksonClark:
@lilygn Going to need some updates to FleetCast, it seems to use TiDB cloud, however, we want to use a database deployed with TiDB operator. You can see how we were deploying the operator before in `srearena/service/tidb_cluster_operator.py`. You should be able to reuse lots of that code.

Once that is done, the next step will be a load generator that calls the frontend with different requests providing coverage of the app.

After that, we can reintroduce the tidb operator issues that are currently commented out in `registry.py`.

lilygn:
@HacksonClark I created a script that sets up tidb cluster and tidb operator such that they are configured to my app. You run the deploy fleet cast and my tidb_create.py, which successfully connects the backend to the cluster.

lilygn:
@HacksonClark Do you want me to make the load generator, or are you planning to?

HacksonClark:
@lilygn That would be great if you could!

lilygn:
@HacksonClark My server already does that. Do you want me to isolate the load generator feature outside of the server, or for create_workload and start workload, can I just write NoOp. I schedule runs for the workload using my server.

HacksonClark:
@lilygn It doesn't matter where the load generator is as long as there is load flowing through the system.

lilygn:
1) Was trying to fix the "yes" VS yes submit bug. It would not let me type in yes. 
2)  tidb_prometheus does not deploy prometheus. It patches the deployment such that it is annotated for prometheus. 

Example:
-     "prometheus.io/scrape": "true",
        "prometheus.io/path": "/metrics",
        "prometheus.io/port": FLEETCAST_METRICS_PORT,

yimingsu01:
> 1. Was trying to fix the "yes" VS yes submit bug. It would not let me type in yes.
> 
>     2. tidb_prometheus does not deploy prometheus. It patches the deployment such that it is annotated for prometheus.
> 
> 
> Example:
> 
>     * ```
>       "prometheus.io/scrape": "true",
>         "prometheus.io/path": "/metrics",
>         "prometheus.io/port": FLEETCAST_METRICS_PORT,
>       ```

Got it. Good observation that the submission code is buggy. The team is aware of the bug. 

For 2, can you merge that patching logic into your own application? Technically, `tidb_prometheus` is not an application in the benchmark.

PR #205, opened 2025-08-06
Etcd snapshot

777lefty:
Build an etcd snapshot module, which can execute etcd snapshot backup and restoration in all 3 scenarios wherever SREArena runs. Initial test results show that applying snapshot strategy can reduce deployment time by around 26% for helm-installed apps, while it shows no strength against redeployment when it comes to non-helm apps. Therefore, we decided to apply the module in a branch style: The deployment strategy (snapshot or redeployment) depends on the app.

Here are the main modifications:
1) Added `cluster_snapshot.py` module, where it reads from `config.yml` and `inventory.yml` to detect which scenario SREArena is running in, and accordingly chooses to execute backup and restoration 1) via ssh, 2) locally or 3) in kind docker.
2) Modified `conductor.py`, where it detects the apps name, and if it's non-helm app, then redeployment is executed, otherwise it takes a snapshot of etcd if the app is first deployed, then it restores the cluster states after fault recovery.
3) Modified `deploy` function in apps, where a newly added `kubectl.check_if_ready` function is applied to detect whether the app has been deployed, if so, it skips deployment. (Only for helm-installed)
4) Added `helm_install` attribute to all apps to indicate whether it's helm installed. (Except non-finished apps)
5) Modified `remote_setup_controller_worker.yml` by adding one step for control node to install etcd-client (etcdctl).

### DIFF
--- a/srearena/conductor/conductor.py
+++ b/srearena/conductor/conductor.py
@@ -152,14 +152,6 @@ class Conductor:
         return "[✅] Problem completed."
 
     def load_workers(self):
-        # with open(path, "r") as f:
-        #     inventory = yaml.safe_load(f)
-        # control_host = list(inventory["all"]["children"]["control_nodes"]["hosts"].values())[0]["ansible_host"]
-        # worker_hosts = [
-        #     v["ansible_host"]
-        #     for v in inventory["all"]["children"]["worker_nodes"]["hosts"].values()
-        # ]
-        # return control_host, worker_hosts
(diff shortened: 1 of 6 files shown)


jq-huang:
I resolved conflicts locally and pushed again, please review them

HacksonClark:
I don't think we will use this given the performance profiling, but I hope you learned a lot from doing it @777lefty 

Thanks for all the great work!

Issue #182, opened 2025-07-24
Add `agnostic` label to problems interface


HacksonClark:
From our problem [spreadsheet](https://docs.google.com/spreadsheets/d/1FGIeLNcKsHjrZGQ_VJcQRGl6oTmYyzjW0_ve5tfM_eg/edit?gid=0#gid=0) we have them labelled as agnostic or not in the last column. Though, the column isn't filled out.

Agnostic refers to a fault that is application agnostic, meaning we can inject it on any application. In our base problem object, I would like to add an agnostic label and update the problems to use this label. This is important for a future feature.

hieue:
Does agnostic include services as well? Or just apps?

HacksonClark:
@hieue just problems

hieue:
I mean as in these faults can be applied to any service or just apps? Because I think they might be useful for #65

HacksonClark:
Any service. They are not tied to a specific application.

HacksonClark:
Closing as this will be part of #191 and BYOA, which is a future feature that is currently unplanned. But we will revisit this later.

PR #203, opened 2025-08-06
Benchmark driver

HacksonClark:
Closes #194 
Closes #70 
Closes #96 (though this may need an MCP component to go along with the http).
Closes #200 
Closes #210 
Closes #213
Closes #214 
Closes #215 

The new driver of the benchmark will be `main.py`, which will run through all the problems and start the http api server (and the mcp server as well).

PR #219, opened 2025-08-14
Add agnostic property

Jackcuii:
Add agnostic property to the registry. #182 

Add property to the problem seems not good, due to some of the values in the registry are lambda wrapper. So we can not know the `agnostic` before we instantiate a problem. (But when we init a problem, app has been created.)
So I add the property to the registry.

Though some of the problems are agnostic, additional specifications are needed for BYOA app. (including deployment ID, service ID, container ID, ports, env_var names, api path...) details are listed here: [agnostic_details_of_srearena_problems](https://docs.google.com/spreadsheets/d/1bMgaiYJAWP1aRtnkQZ-GZvtbKFTlp-I-PJdbRkPZH1s/edit?gid=0#gid=0)
(Most of the specifications can be automatized, though.l)

Maybe we need a separate **register procedure** for the BOYA application, deploy it, go through it and find spec for the agnostic problems.

### DIFF
--- a/srearena/conductor/problems/registry.py
+++ b/srearena/conductor/problems/registry.py
@@ -60,28 +60,28 @@ from srearena.conductor.problems.wrong_service_selector import WrongServiceSelec
 class ProblemRegistry:
     def __init__(self):
         self.PROBLEM_REGISTRY = {
-            "k8s_target_port-misconfig": {"creator": lambda: K8STargetPortMisconfig(faulty_service="user-service"), "agnostic": False},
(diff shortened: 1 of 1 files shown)


HacksonClark:
@Jackcuii Great work so far! I'll have to dive deeper into this after #203

Jackcuii:
> @Jackcuii Great work so far! I'll have to dive deeper into this after #203

Thanks :cowboy_hat_face:  For now, should I explore the automation settings for the BOYA app, or some other jobs for me?  :D

HacksonClark:
@Jackcuii Sure, that sounds great!

Jackcuii:
Maybe this PR should be closed since we do not need this feature.

Issue #176, opened 2025-07-21
Add NTAM localization oracle


HacksonClark:
ITBench has NTAM (normalized topology aware metric) for localization evaluation. It is not a trivial addition as the metric requires us to know the topology. Please read the ITBench paper to get an understanding of the metric. It would be a great addition to SREArena.

hieue:
Since the topology will consist of all entities (including things like ConfigMap, Endpoint, etc.) are these resources eligible to be ground truths as well? Or are ground truths strictly confined to services/ deployments?

HacksonClark:
@hieue Let's just stick with services for now.

hieue:
Imo, if it's just services, NTAM wouldn't be a super useful metric. We can just instruct the agent to trace back the fault (say due to a misconfigured PVC-PV pair) to a service to do an exact match by having it trace pvc -> pod -> replicaset -> deployment. Horizontally, the services aren't really connected to each other through resources, so you can't really estimate how a fault that happened in "user" would be close to that in "consul". If the ground truths are expanded to include stuff like PVCs, I think it'd be a more meaningful metric because we can use NTAM to gauge how close the agent is to the root of the problem. For example, if the root cause is a faulty PV, the agent would get a high score for submitting the associated PVC, and a low score for submitting an unrelated service. 

If I'm misunderstanding the point of this metric, please correct me. This stuff is a little vague to me right now.

HacksonClark:
@hieue the dependencies come from the call graph from traces. It's based on the topology of the application.

See this as an example: https://opentelemetry.io/docs/demo/architecture/

hieue:
Ah ok that's a lot clearer. What I envisioned from the ITBench paper was mostly resources within a namespace. I'll look into the traces.

HacksonClark:
No worries. One thing that would be interesting is having a tool that can grab the topology. In ITBench I believe they just wait for 5 minutes and use the graph from Jaeger (but this makes me kinda sad). I feel we can do better. I have some ideas on this (I was thinking of doing a research project related to it in 538) so if you're interested do let me know. Though I will be preoccupied with #204 so I won't really be able to help you hack code.

hieue:
If it helps build this feature better I'll take a look.

HacksonClark:
We're killing NTAM since IBM doesn't even use it anymore.

PR #220, opened 2025-08-16
[#176] Normalized Topology Aware Metric

hieue:
WIP state

Formula to be added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -337,6 +337,7 @@ data
 taskweaver_proj
 trace_output/
 metrics_output/
+topology_graph/
 
 # Ignore telemetry data directories
 srearena/observer/telemetry_data_*/
--- a/srearena/conductor/oracles/localization.py
+++ b/srearena/conductor/oracles/localization.py
@@ -1,15 +1,83 @@
+import os
+import networkx as nx
+import csv
+import math
+
 from srearena.conductor.oracles.base import Oracle
 from srearena.conductor.oracles.utils import is_exact_match, is_subset
-
+from srearena.observer import root_path
 
 class LocalizationOracle(Oracle):
     def __init__(self, problem, expected: list[str]):
         super().__init__(problem)
         self.expected = expected
 
+    def _fetch_topology_map(self):
+        file_path = root_path / "topology_graph" / f"{self.problem.namespace}.csv"
+        if not os.path.exists(file_path):
+            print(f"No snapshot of the topology of {self.problem.namespace} currently exists.")
+
+        self.topology_graph = nx.Graph()
+
+        with open(file_path, "r", newline="") as f:
+            reader = csv.reader(f)
+            next(reader)
+            for row in reader:
+                if len(row) >= 2:
+                    u, v = row[0], row[1]
+                    self.topology_graph.add_edge(u, v, weight=1)
+    
+    def ntam_score(self, predicted_nodes, ground_truths, alpha=1.0, beta=1.0, gamma=1.0, omega=0.7, tau=0.8):
+        """
+        Compute Normalized Topology-Aware Match (NTAM) score.
+
+        Parameters:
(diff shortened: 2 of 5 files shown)


HacksonClark:
This is awesome so far @hieue! Great work!

hieue:
@HacksonClark Here is what I currently have for NTAM localization oracle. I'm marking as draft cause I'm not sure how to handle fetching the topology map just yet.

As of now I'm basically assuming there will be a csv containing the edges inside `srearena/observer/topology_graph`. This graph by generated by collecting Jaeger traces through running `srearena/observer/trace_api.py`, but this runs into the problem of having to keep an app up for 30 minutes before running it which you don't like. I read all the related work you recommended but most of them seems to be using traces anyways. If you think there's no better solution I'll implement an option for the user to generate the graph the first time (aka sitting still for 30 minutes), and have a static graph saved in the repo for subsequent runs.

HacksonClark:
@hieue I think static graph is good for now, I believe other projects do similar. It would be an interesting research problem finding a good and fast way to get the topology.

HacksonClark:
See how we can get this graph from jaeger using the API:
image

All of our apps have jaeger, so we could use this for now.

To see the graph, run an astronomy shop problem, port-forward like so:
```kubectl --namespace astronomy-shop port-forward svc/frontend-proxy 8080:8080```

and then you can go to this url: http://localhost:8080/jaeger/ui/

I believe we could just save a graph with each application (until we find a better solution).

According to GPT you can do it like so:
```
That view is Jaeger’s **Dependencies graph**.
You can get it in two ways:

1. **In the UI**
   Open Jaeger → select the **Dependencies** tab → pick a time range.
   (If you use Elasticsearch/Cassandra storage, make sure the **jaeger-spark-dependencies** job is running so the deps table/index is populated.)

2. **Via the API (to fetch the DAG programmatically)**
   Jaeger Query exposes a JSON endpoint:

```bash
# last 24 hours
END_TS=$(date +%s000)                 # epoch millis now
LOOKBACK=$((24*60*60*1000))           # 24h in millis

curl "http://<jaeger-query-host>:16686/api/dependencies?endTs=${END_TS}&lookback=${LOOKBACK}"
```

The response contains `nodes` (services) and `links` (edges with call counts).
You can feed that JSON into Graphviz/NetworkX/etc. to render the DAG like in your screenshot.


Let me know if you need any help with this.

hieue:
How should we store the graphs? I generated them into csv files for now. I'm not sure if I should just push them into the repo or zip it and give a separate download link

HacksonClark:
@hieue I think pickling them would be fine. We should probably use some package like networkx to handle the graphs.

HacksonClark:
@hieue I see the topology mapper is in now. What is our current status? What is outstanding for us to do?

hieue:
Right now I'm finetuning the scoring function. I wanna revise it because it doesn't currently consider a few scenarios, as I tried it out and very different solutions ended up with very similar scores. I do need a little bit of clarification:

1. Are we planning on evaluating services many-to-many? As in there are many services that can be faulty and the solution submitted will be a list of services (evaluating the whole propagation chain, which I believe is what ITBench meant), or will it just be comparing a single service to a single root cause 
2. Say there's a path from A->B but not the other way around. Assuming the agent predicts B when A is the root cause, is the topological distance between B and A zero or dist(A->B)?

As for 1, if we are evaluating many-to-many, I think current approach is ineffective if all the predicted services are all close to a single ground truth but not the other ones. Think gt = [A, B, C], pred = [A, A1, A2] where A, A1, A2 are neighbors and far from B and C, which current code will give a high score, so I'm just thinking of a different approach that will ensure the whole propagation chain is considered.

HacksonClark:
I'm closing since ITBench doesn't even use NTAM anymore. However, I do think our code for grabbing the graphs might prove useful in the future.

PR #222, opened 2025-08-17
Port Stratus into Langgraph (Eval Ready)

yimingsu01:
This PR completes the porting of Stratus into Langgraph, plus a complete working eval pipeline.

To run the eval pipeline, make sure you have a valid kubeconfig.

# Reproduce eval
Currently, for eval purpose, the agents are kicked started by the benchmark when problem setup are ready. Uncomment `# await stratus_driver()` on [line 43](https://github.com/xlab-uiuc/SREArena/blob/c4b6c61c30eaf898168d07b8cb4624544b68fe89/main.py#L43) in `main.py` to start the agent. Note that currently the agent only supports running standard problems (i.e., problems with all 4 phases), but supporting custom problems requires no more effort than configuration changes.

After uncommenting, you only need to run [`main.py`](https://github.com/xlab-uiuc/SREArena/blob/c4b6c61c30eaf898168d07b8cb4624544b68fe89/main.py) at the root of the repo. For each problem, after the benchmark sets up the problem, it kicks start the agent. The rest is waiting. 

When the agent finishes a problem, it saves an output csv containing some useful agent information (e.g., token usages). 

# Running standalone agent
[`driver.py`](https://github.com/xlab-uiuc/SREArena/blob/stratus_eval/clients/stratus/stratus_agent/driver/driver.py) describes the workflow of Stratus. If you would like to start the benchmark and run the agent separately, simply run [`driver.py`](https://github.com/xlab-uiuc/SREArena/blob/stratus_eval/clients/stratus/stratus_agent/driver/driver.py) to start the Stratus workflow.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -34,31 +34,12 @@ MCP_SERVER_URL=http://127.0.0.1:${MCP_SERVER_PORT}
 
 # Agent LLM Config
 # Maximum number of retries for API calls
-MAX_RETRIES=5
+LLM_QUERY_MAX_RETRIES=5
 # Initial delay in seconds for retries
-INIT_RETRY_DELAY=1
+LLM_QUERY_INIT_RETRY_DELAY=1
 PROVIDER="openai"
 WATSONX_API_KEY=""
 PROVIDER_TOOLS="openai"
 MODEL_TOOLS="gpt-4o-mini"
 URL_TOOLS="https://api.openai.com/v1"
 API_KEY_TOOLS=""
-
-# Agent Workflow Config
-
-SUMMARY_FREQUENCY=3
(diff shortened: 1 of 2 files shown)


HacksonClark:
Do you think there's some "kickoff agent" interface we should develop as part of the "agent registration"?

yimingsu01:
> Do you think there's some "kickoff agent" interface we should develop as part of the "agent registration"?

Yes, that would be nice, makes stuff simpler for agent devs.

HacksonClark:
I think as part of registration we ask the user for a "kickoff" command which we just run as a subprocess. What do you think?

Issue #228, opened 2025-08-25
Some bugs in EnvVariableLeak Problem.


Jackcuii:
About #60 #136 : 

Hi Jackson @HacksonClark 😃 

When I make the env var leak problem agnostic, I found the problem has something wrong.
(please correct me if I am wrong) According to my understanding, this means the environmental variable is lost.

1. The configMap can be used to mount config files (with `configMap` field [Doc](https://kubernetes.io/docs/concepts/configuration/configmap/#using-configmaps-as-files-from-a-pod)) or introduce environmental variables (with `configMapRef` field [Doc](https://kubernetes.io/docs/concepts/configuration/configmap/#using-configmaps-as-files-from-a-pod)). However, the impl of this problem upon HotelRes and SocialNet both delete the configmap for the first use. 

e.g.
```
...
                    {
                        "configMap": {
                            "defaultMode": 420,
                            "name": "mongo-geo-script"
                        },
                        "name": "init-script"
                    },
                    ...
    },
```
`inject_virtual.py:`
```
if self.namespace == "social-network":
                configmap_name = "media-mongodb"
            elif self.namespace == "hotel-reservation":
                configmap_name = "mongo-geo-script"
            else:
                raise ValueError(f"Unknown namespace: {self.namespace}")
```

2. The impl of this problem is to delete a ConfigMap, the crash seems not to be caused by 'missing env var' but 'can not find ConfigMap 

Thanks for discussions with @777lefty

jq-huang:
Totally agree with Bohan, we didn't find any signs that indicate deleting the configmap would impact environment variables regarding the pods in the deployments. Please correct us if we're wrong.

HacksonClark:
I think we could just update this problem to be "missing configmap" and then make something better to remove an env var. What do you folks think? @Jackcuii @777lefty

jq-huang:
A natural design would be locating a configmap with `envFrom.configMapRef` or `env[].valueFrom.configMapKeyRef` field, delete or modify that field, and update configmap, like what `OtelFaultInjector` did.

HacksonClark:
@777lefty That sounds good, can you do that fix?

jq-huang:
@HacksonClark Of course, I can do that tomorrow

jq-huang:
[...]

Issue #233, opened 2025-09-02
Wrk2 requests fail on healthy app


777lefty:
While running `k8s_target_port-misconfig` problem, I noticed that the cluster doesn't manifest any sign of crash or failure after fault injection, as all the resources are in `running` state with no error. So it seems that the only way to diagnose the problem is to watch the output log of the `wrk2` job that keeps continuously sending request to the `social network` app. If there is failure in the level of application, the result of the workload test should reflect it.

However, when I observed the output log of the workload ***before*** fault injection, the log indicates that all the requests fail. Below is the result log of the very first round of workload test right after the app was deployed and in `running` state:
```
Running wrk2 on round #0                                                                                                                                                                                      
Running 10s test @ http://10.105.8.245:8080/wrk2-api/post/compose                                                                                                                                             
  3 threads and 3 connections                                                                                                                                                                                 
                                                                                                                                                                                                              
  Thread calibration: mean lat.: 1.739ms, rate sampling interval: 10ms                                                                                                                                        
  Thread calibration: mean lat.: 1.708ms, rate sampling interval: 10ms                                                                                                                                        
  Thread calibration: mean lat.: 1.665ms, rate sampling interval: 10ms                                                                                                                                        
                                                                                                                                                                                                              
-----------------------------------------------------------------------                                                                                                                                       
Test Results @ http://10.105.8.245:8080/wrk2-api/post/compose                                                                                                                                                 
  Thread Stats   Avg      Stdev     99%   +/- Stdev                                                                                                                                                           
    Latency     -nanus    -nanus   0.00us    0.00%                                                                                                                                                            
    Req/Sec       -nan      -nan   0.00      0.00%                                                                                                                                                            
  Latency Distribution (HdrHistogram - Recorded Latency)                                                                                                                                                      
 50.000%    0.00us                                                                                                                                                                                            
 75.000%    0.00us                                                                                                                                                                                            
 90.000%    0.00us                                                                                                                                                                                            
 99.000%    0.00us                                                                                                                                                                                            
 99.900%    0.00us                                                                                                                                                                                            
 99.990%    0.00us                                                                                                                                                                                            
 99.999%    0.00us                                                                                                                                                                                            
100.000%    0.00us                                                                                                                                                                                            
                                                                                                                                                                                                              
  Detailed Percentile spectrum:                                                                                                                                                                               
       Value   Percentile   TotalCount 1/(1-Percentile)                                                                                                                                                       
                                                                                                                                                                                                              
[...]

PR #234, opened 2025-09-03
Developed missing_env_variable problem for astronomy-shop

777lefty:
Developed missing_env_variable problem for astronomy-shop, with new fault injector and recovery, and mitigation oracle

### DIFF
--- a/srearena/utils/decorators.py
+++ b/srearena/utils/decorators.py
@@ -1,7 +1,15 @@
-# Set the fault_injected flag when inject_fault is called
 def mark_fault_injected(method):
     def wrapper(self, *args, **kwargs):
-        result = method(self, *args, **kwargs)
+        try:
+            result = method(self, *args, **kwargs)
+        except Exception as e:
+            if method.__name__ == "inject_fault":
+                # We exit if there's an error during fault injection, warning if in recovery
+                raise
+            else:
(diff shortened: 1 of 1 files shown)


HacksonClark:
@777lefty Looks great! Will review soon.

jq-huang:
@HacksonClark I renamed the `env_variable_leak` problem to be `missing_configmap`, so this would close issue #228.
Besides, the problem's interface is designed to be app agnostic, but since missing environment variable usually won't cause cluster crash, it can only diagnosed by observing workload. Due to the `wrk2` issue I mentioned in issue #233, I just developed `missing_env_variable_astronomy_shop`, as `locust workload` for `Astronomy Shop` seems to be correct.

HacksonClark:
Closes #228 

We can investigate the wrk2 issue in that other issue and submit a PR. I think I'll take a look at that tomorrow.

jq-huang:
@HacksonClark I see, I reproduced the bug, and it can be reproduced on your `incorrect_port_assignment` problem as well, since these two problems are highly similar.
I probed into this bug and figured out how it happens, I'll open an issue for it

jq-huang:
@HacksonClark #235 for the bug you detected, the bug should be elsewhere as well.

jq-huang:
The latest commit should close #235

HacksonClark:
@777lefty Excellent!

Issue #239, opened 2025-09-04
Multiple independent failures


HacksonClark:
From the gdoc:
```
3.2 Multiple Independent Failures
Pending, let me first dive into correlated problems.
Target scenario: multiple microservice applications in one cluster, each application experiencing failures.
Should we confine that each fault we inject into each microservice application should not cascade to other applications? No, try to inject faults that have varying degrees of cascading effects. 
```

So this issue will require some changes to the app deployment process, where we switch from deploying one application at a time to all of the applications running at the same time. This may take some work as our interfaces currently assume just a single application.

yimingsu01:
@HacksonClark My current experiment is to create a `MultipleIndependentFailures` problem which deploys a `CompositeApp`. `MultipleIndependentFailure` is a wrapper class that enables the user to compose different existing problems together and deploy/inject at the same time. `CompositeApp` is then a wrapper class that wraps around all $\mu$service applications and deploy/inject/cleanup at the same time.

HacksonClark:
@yimingsu01 yep, i was thinking similar things.

yimingsu01:
In terms of failure model, I would need to do some case studies first, but one key point is how the relationship between the injected faults affect how failures manifest.

PR #243, opened 2025-09-06
Implementing Multiple Independent Failures (aka Concurrent faults)

yimingsu01:
# Syopsis
This PR enables SREArena to simulate multiple independent failures in a k8s cluster.

The implementation allows composing different existing problems into one meta problem, such that multiple problems will be running at the same time on the same cluster, simulating a k8s cluster where multiple deployed applications are failing due to independent failures.

# Changes
Introduced `MultipleIndependentFailures` class in `srearena.conductor.problems.multiple_failures.py`. It acts as a wrapper problem over existing problems to support methods called by the conductor.

To define a multiple independent failure scenario, define a `MultipleIndependentFailures` object with your chosen problems. E.g.,

```python
MultipleIndependentFailures(
    problems=[
        K8STargetPortMisconfig(faulty_service="user-service"),
        MongoDBRevokeAuth(faulty_service="mongodb-geo"),
        MongoDBRevokeAuth(faulty_service="mongodb-rate"),
    ]
),
```

In terms of problem semantics, `MultipleIndependentFailures` inherits from does not differ from existing problems. 

Introduced `CompositeApp` class in `srearena.service.apps.composite_app.py`. Similarly, this is only a wrapper around existing applications to allow the conductor deploy multiple applications at the same time.

# Example problems
I included some example problems as showcases of `MultipleIndependentFailures`, but the combinations are unlimited as far as I know. I also added "mega problems": problems with the same applications to registry as examples.

# Known issues
1. `PersistentVolumeAffinityViolation()` has trouble recovering when running together with:
```python
MongoDBAuthMissing(),  # SocialNet
ScalePodSocialNet(),  # SocialNet
AssignNonExistentNode(),  # SocialNet
PodAntiAffinityDeadlock(),  # SocialNet
TaintNoToleration(),  # SocialNet
```
It works fine when running by itself. The root cause seems to be that the recovery CR does not represent the actual recovered desired state but still the faulty one.

2. Fault injection/recovery's time scales with # of problems 
Some fault injection implementation needs to `wait` for fault propagation (e.g., `WrongDNSPolicy` waits until the dns propagates to pods). We can optimize this through running them in parallel using `thread` or `multiprocessing`.

3. Certain faults affect how other problem injects or recovers
For example, `TaintNoToleration` can affect how pods from different problems  recover as it does not allow pod scheduling to happen at all.

### DIFF
--- a/srearena/conductor/oracles/missing_env_variable_mitigation.py
+++ b/srearena/conductor/oracles/missing_env_variable_mitigation.py
@@ -1,0 +1,71 @@
+from srearena.conductor.oracles.base import Oracle
+
+
+class MissingEnvVariableMitigationOracle(Oracle):
+
+    def evaluate(self) -> dict:
+        print("== Mitigation Evaluation ==")
+
+        kubectl = self.problem.kubectl
+        namespace = self.problem.namespace
+        results = {}
+
+        faulty_service = self.problem.faulty_service
+        env_var = self.problem.env_var
+        env_var_value = self.problem.env_var_value
+
+        all_normal = False
+        # check if deployment exists
+        try:
+            deployment = kubectl.get_deployment(faulty_service, namespace)
+            env_var_found = False
+
+            # check if env var exists in deployment
+            for container in deployment.spec.template.spec.containers:
+                if hasattr(container, 'env') and container.env:
+                    for env in container.env:
+                        if env.name == env_var and env.value == env_var_value:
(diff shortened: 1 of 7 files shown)


yimingsu01:
@HacksonClark feel free to checkout this branch and run `main.py` directly. It should run the only one problem, a `MultipleIndependentFailures` problem composed of 3 problems, each problem uses a different app.

HacksonClark:
Closes #239

PR #246, opened 2025-09-08
[#190] Adding task selection

### LINKED ISSUE #190: Allow task selection in problems
@yimingsu01 is requesting the ability (as an option) to evaluate just one task (like detection or mitigation, etc) to make evaluation easier. I think we can just add this is a config option.

hieue:
Resolves #190 

Adds task selection with `srearena/conductor/tasklist.yml`. Description on implementation in README.


HacksonClark:
I seem to be having an issue with using the http api running the problem with main:
```
curl -X POST "http://localhost:8000/submit" \
  -H "Content-Type: application/json" \
  -d '{"solution": "Yes"}'
```


I got this response:
```
{"detail":"Grading error: Error parsing response: Error parsing response: Unsupported AST node type: <class 'ast.Name'>"}%   
```

Not sure if the bug is here or elsewhere.

PR #253, opened 2025-09-15
first version of transient issues implemented with chaos-mesh

777lefty:
Implemented a `TransientIssueGenerator` class that uses chaos-mesh to continuously inject self-cleaning chaos experiment as self-healing transient issues randomly. The generator supports 8 universal chaos experiments that are classified into two types and can be injected within 5 levels of pod scopes based on user choice.

The generator can be started after fault injection and stopped in cleanup phase.

### DIFF
--- a/srearena/generators/noise/transient_issues/chaos_injector.py
+++ b/srearena/generators/noise/transient_issues/chaos_injector.py
@@ -7,7 +7,8 @@ from kubernetes import client
 from srearena.service.helm import Helm
 from srearena.service.kubectl import KubeCtl
 
-class ChaosInjector():
+
+class ChaosInjector:
     def __init__(self, namespace: str):
         self.namespace = namespace
         self.kubectl = KubeCtl()
@@ -42,8 +43,11 @@ class ChaosInjector():
         release_exists = Helm.exists_release(chaos_configs["release_name"], chaos_configs["namespace"])
         if not release_exists:
             Helm.install(**chaos_configs)
+            self.kubectl.wait_for_ready("chaos-mesh")
         else:
-            print(f"[ChaosInjector] Helm release '{chaos_configs['release_name']}' already exists in namespace '{chaos_configs['namespace']}', skipping install.")
+            print(
+                f"[ChaosInjector] Helm release '{chaos_configs['release_name']}' already exists in namespace '{chaos_configs['namespace']}', skipping install."
+            )
 
     def create_chaos_experiment(self, experiment_yaml: dict, experiment_name: str):
         try:


jq-huang:
An interactive interface to enable and configure transient issues are added, if the modification to enable this in `cli.py` is considered unnecessary, please just drop this feature

jq-huang:
Hi @HacksonClark, so you basically mean that the configuration should be done within a yaml file instead of interaction in the command line style, right? That means if the user want to change the configuration of noise, they should just modify the yaml on them own, is that correct? Btw, where do you find it a good location to put the configuration file for the noise, how about `srearena/generators/noise/configuration.yaml`?

HacksonClark:
@777lefty Yes, we should use a yaml file and just have default values there. If the user wants to change them, they will just change that file on their own.

For the location of the file, what you mentioned sounds reasonable.

jq-huang:
Hi @HacksonClark, two new commits dealing with the reviews

HacksonClark:
@777lefty somehow getting this issue just starting cli.py:
```
python cli.py
  File "/Users/jackson/Desktop/SREArena/cli.py", line 21, in <module>
    from srearena.conductor.conductor import Conductor
  File "/Users/jackson/Desktop/SREArena/srearena/conductor/__init__.py", line 1, in <module>
    from .conductor import Conductor
  File "/Users/jackson/Desktop/SREArena/srearena/conductor/conductor.py", line 14, in <module>
    from srearena.generators.noise.transient_issues.transient_issues import TransientIssuesGenerator, FaultType, PodScope
ModuleNotFoundError: No module named 'srearena.generators.noise.transient_issues.transient_issues'; 'srearena.generators.noise.transient_issues' is not a package
```

Is there something not committed?

jq-huang:
> @777lefty somehow getting this issue just starting cli.py:
> 
> ```
> python cli.py
> Traceback (most recent call last):
>   File "/Users/jackson/Desktop/SREArena/cli.py", line 21, in <module>
>     from srearena.conductor.conductor import Conductor
>   File "/Users/jackson/Desktop/SREArena/srearena/conductor/__init__.py", line 1, in <module>
>     from .conductor import Conductor
>   File "/Users/jackson/Desktop/SREArena/srearena/conductor/conductor.py", line 14, in <module>
>     from srearena.generators.noise.transient_issues.transient_issues import TransientIssuesGenerator, FaultType, PodScope
> ModuleNotFoundError: No module named 'srearena.generators.noise.transient_issues.transient_issues'; 'srearena.generators.noise.transient_issues' is not a package
> ```
> 
> Is there something not committed?

Some deletion weren't committed, just committed them, should be okay now

jq-huang:
Closes #238

PR #255, opened 2025-09-16
Metastable failures

777lefty:
Added two more metastable failures on `blueprint-hotel-reservation`, and renamed the previous `rpc_retry_storm`

### gc_capacity_degradation
Patch the `GOGC` env variables in containers to make garbage collection more frequent, then trigger the metastable state by load spike

### capacity_decrease_rpc_retry_storm
Modify the rpc setting to make requests prone to timeouot and retry, then trigger a retry storm by inject cpu containment with chaosmesh

### load_spike_rpc_retry_storm
Renamed the previous `rpc_retry_storm` since we have two similar problems with different triggers now.

### Misc
- The `capacity_decrease_rpc_retry_storm` needs chaosmesh injector, so I submit the `chaos_injector.py` in this branch as well, the injector is identical to that in `transient-issue` branch, so should there not be conflicts.
- The chaos problems along with the symptom injector are deleted following @HacksonClark 's opinion.


jq-huang:
Hi @HacksonClark, I tried to reproduce the bug you encountered, and I succeed in reproducing it, and I found out that the root cause of the bug is that somehow the currently specific version of aiopslab (`[aiopslab-applications @ d9fb7e0](https://github.com/xlab-uiuc/aiopslab-applications/tree/d9fb7e0c10434679c02143dcc5ffc149c3f114a0)`) that is using in SREArena doesn't contain Blueprint-Hotel-Reservation 👀

HacksonClark:
@777lefty Okay, you could update the submodules.

HacksonClark:
@777lefty I'll update the submodules and then update it here so I can test.

jq-huang:
> @777lefty I'll update the submodules and then update it here so I can test.

Sry I just saw this, it seems that you just updated it?

HacksonClark:
@777lefty Yep, I got it! I'm reviewing right now.

PR #264, opened 2025-09-23
Latent sector error

HacksonClark:
Adding latent sector error as a problem, works using debugfs and dm-dust to mark bad blocks.


Saadmrp1038:
Moved this to new PR.

PR #310, opened 2025-10-16
Fixed conflicting port issue

xinbowu2:
This pull request fixes an issue where both the API and MCP services were configured to use the same port, causing conflicts and preventing the application from starting correctly.


tianyin:
LGTM

Issue #278, opened 2025-10-01
Localization should use UID


HacksonClark:
@yimingsu01 and I were discussing in the lab how to change localization. As it currently exists, we just use string comparison (or a list of strings if there are multiple faulty resources). Given that many problems affect multiple different kinds of resources (like ConfigMaps) which are not captured well since we just focus mostly on pods.

However, kubernetes does assign a UID to EVERY resource. It makes sense to ask the agent for this (or a list). We will have to write some utility functions to find the mapping, but we can be a lot more specific with this. We may even make an agent tool for this as well.

An example:
```
kubectl get pod kube-flannel-ds-bq5jm -n kube-flannel -o jsonpath='{.metadata.uid}'
```
Output:
```
c82798a3-84a0-477b-a6c5-e90a8fe92bf5
```

tianyin:
I briefly discussed with @yimingsu01 in the lab today. 

It seems that a LLM-as-a-judge approach can be a general approach. 

OTOH the approach mentioned in this issue seems good?

HacksonClark:
@tianyin I don't think LLM as a judge is necessary for localization, but if we were to bring back RCA it would be as the old AIOpsLab categories are not mutually exclusive as we observed during the Stratus rebuttal.

tianyin:
sure. So @yimingsu01 says we should discuss -- it seems that you guys already have an answer.

HacksonClark:
@tianyin I agree.

tianyin:
good!

yimingsu01:
We should make an agent tool for this so that the agent can easily get the UID. 

Actually, why can't we also ask the agent to output different types of resources as the localization task answer? I.e., fundamentally, what do we gain from asking for UIDs instead of pod names/annotations?

We seem to be not satisfied about two things:
1. the parsing process of the agent's localization answer is extremely brittle
2. the agent might output a correct answer but the current localization oracle does not grade it correctly (false negatives).

Asking for UID resolves 1, but not necessarily resolves 2 IMO.

Problem 2 boils down to a rubric problem: how do we specify the correct answer for localization?

yimingsu01:
From the aiopslab paper:
> The localization (Level 2) task asks the agents to specify the exact location of the fault, e.g., a service or pod name in Kubernetes.

Regardless of it's UID or "faulty service" we ask, the task semantics itself is the same. So if it's mostly a parsing problem, we can specify a better interface for that, e.g., asking the agent to submit a json like:
```json
[...]

PR #313, opened 2025-10-18
UID for Localization

lilygn:
I created an agent tool that retrieves the agent’s guessed UID. I then implemented a version of the LocalizationOracle that compares the agent’s guessed UID against pods belonging to the faulty service. If faulty service is empty, the ground truth UIDs would be the UIDs of the pods in the namespace. Additionally, the oracle counts the result as correct if the agent gets the UID of the service the faulty pod belongs to. 

@HacksonClark Is this a good start?

### DIFF
--- a/sregym/conductor/oracles/localization_oracle.py
+++ b/sregym/conductor/oracles/localization_oracle.py
@@ -121,12 +121,15 @@ class LocalizationOracle(Oracle):
         if isinstance(solution, str):
             # Check if it's a comma-separated list
             # strip char before [
-            solution = solution.split("[")[1]
-            # strip char after ]
-            solution = solution.split("]")[0]
-            if "," in solution:
-                # split by comma, strip space and quote
-                solution = [s.strip().strip("\"'") for s in solution.split(",")]
+            if "[" in solution and "]" in solution:
+                solution = solution.split("[")[1]
+                # strip char after ]
+                solution = solution.split("]")[0]
+                if "," in solution:
+                    # split by comma, strip space and quote
+                    solution = [s.strip().strip("\"'") for s in solution.split(",")]
+                else:
+                    solution = [solution.strip().strip("\"'")]
             else:
                 solution = [solution.strip().strip("\"'")]
         elif isinstance(solution, list):


HacksonClark:
Closes #278

@lilygn @Jackcuii will be taking this over. Focus on your problem testing tool.

HacksonClark:
I converted it to a draft, please reopen once it's complete.

Jackcuii:
I read all the problems' implementation and refactor localization according to their RC.

Features
1. 79/86 problems' oracles are refactored except the 'missing xxx'
2. submit format is fixed
3. I have a safeguard check, after the fault is injected, the oracle result will be stored as a checkpoint. And after the agent submit, the oracle will be compared with the checkpoint to make sure the env is stable.
4. Layer design. Support OrOracle. e.g. when two config contradict, then each of them be reported could be a valid solution.
5. Extension place for more flexible oracle  (such as sub-resource) is reserved.
6. Root cause labels are added to the Eval Sheet.

Limitation
1. Some oracles can be more precise
2. Not implement the Missing XXX by now. (need special oracle, now just comment out)
3. Not implement false positive and false negative by now.

Jackcuii:
Closes #278

HacksonClark:
Why are there so many different localization oracles? I figured this would just be a fairly simple change.

Jackcuii:
These reserve room for further more precise localization. Different resource has different relationship to tolerate imprecise localization. e.g. For the RC in deployment, it is okay to report of the pods' uids of it / replicaset's uid. We can support these logic in these files.

Jackcuii:
If we do not have more urgent goals, I can continue working on implement them.

HacksonClark:
> Different resource has different relationship to tolerate imprecise localization.
Please explain.

> For the RC in deployment, it is okay to report of the pods' uids of it / replicaset's uid.
Why wouldn't we just use the UID of the deployment?

Jackcuii:
> > Different resource has different relationship to tolerate imprecise localization.
> > Please explain.
> 
> > For the RC in deployment, it is okay to report of the pods' uids of it / replicaset's uid.
> > Why wouldn't we just use the UID of the deployment?

By now, the oracle is just the UID of the deployment, but the multiple files reserve room for other reasonable solutions (pods' uids of it / replicaset's uid) In many scenario, there could be multiple reasonable answers. E.g. If a problem modifies the configMap, giving the uid of the configMap or the deployment is reasonable. This is because the configMap is **BINDED** to deployment. So we can design a set of logic for each kind of k8s resource according to its relationship with others. Multiple files are used to reserve for that.

HacksonClark:
@Jackcuii I understand.

Jackcuii:
> The submit tool for localization doesn't seem to support a single string despite it saying so in the args:
> 
> ```
> INFO:     127.0.0.1:53007 - "POST /submit/messages/?session_id=2801c8fc00e147b4a53078ea6e3dbd1a HTTP/1.1" 202 Accepted
> INFO - clients.stratus.stratus_utils.get_logger - [localization_mcp] Running command: kubectl get pods frontend -n hotel-reservation -o jsonpath={.metadata.uid}
> ERROR - clients.stratus.stratus_utils.get_logger - [localization_mcp] Unsupported resource type: pods
> INFO:     127.0.0.1:53019 - "POST /submit/messages/?session_id=2801c8fc00e147b4a53078ea6e3dbd1a HTTP/1.1" 202 Accepted
> INFO - clients.stratus.stratus_utils.get_logger - [localization_mcp] Running command: kubectl get pod frontend-78986558db-lft9n -n hotel-reservation -o jsonpath={.metadata.uid}
> INFO - clients.stratus.stratus_utils.get_logger - [localization_mcp] Retrieved UID using Kubernetes client: 07dc0469-2d81-4ee2-8cc2-01c2d7811dff
> INFO:     127.0.0.1:53022 - "POST /submit/messages/?session_id=2801c8fc00e147b4a53078ea6e3dbd1a HTTP/1.1" 202 Accepted
> INFO - clients.stratus.stratus_utils.get_logger - [submit_mcp] submit mcp called
> INFO - all.sregym.conductor - Start Eval for Localization
> ERROR - all.sregym.conductor_api - Grading error: list index out of range
> ```
> 
> When I wrapped the UID in a list it worked: image

Oh let me fix it :saluting_face:

HacksonClark:
@Jackcuii List should be supported (since some faults will affect multiple things at the same time). But it should support just a single string as well.

Jackcuii:
For Stratus it could be fine, since I use `You should submit a python-formatted list of string(s)` in the prompt. 

But supporting this tolerance is reasonable, let me impl it.

HacksonClark:
Bugs are never fine.

Jackcuii:
I mean with this prompt `You should submit a python-formatted list of string(s)`, if it uses only a string without `[ ]`, then it does not correctly follow the prompt instruction.

HacksonClark:
@Jackcuii We're working on the benchmark not the agent, decouple them mentally. I feel we keep running into these issues where things are assumed because of what Stratus does. It does not matter what Stratus does, it matters that our tools work like they say we do.

PR #323, opened 2025-10-27
Clean up logs

Jackcuii:
I clean up most of the dirty logs in the backbone function. 

The debug level logs is stored in log file.

And the readable clean logs are in the stdout.

Many logs in other functions are not cleaned, too many, I will clean up them when I come across them.

The log by now is enough for users of the frameworks.

But I am not sure whether the debug logs is useful for us folks to develop, so please feel free to tell me your suggestions to make the debugging logs more clear.

### DIFF
--- a/agents.yaml
+++ b/agents.yaml
@@ -1,0 +1,6 @@
+agents:
+- name: stratus
+  kickoff_command: python -m clients.stratus.stratus_agent.driver.driver --server
+    http://localhost:8000
+  kickoff_workdir: .
+  kickoff_env: null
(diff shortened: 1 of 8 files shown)


Jackcuii:
> There is no magic for logs.
> 
> @Jackcuii the common practice is:
> 
>     * Allow users to configure the verbosity of the logs
> 
>     * Set the info as the default verbosity of logs
> 
>     * All the logs should be in the log file or stdout if configured.

Got it~ I believe I basically follow this practice. The above DEBUG-level logs are dumped to the logfile, and the above INFO-level logs are printed to the stdout by default. And the INFO level logs are clean enough to be human-readable

And `logging` is used in most of the logs, except the messages (to utilize the `pretty_print` infrastructure.)

Issue #339, opened 2025-11-03
`namespace_memory_limit` submission does not work (11/7)


HacksonClark:
https://github.com/xlab-uiuc/SREGym/issues/333#issuecomment-3479176445

None of the oracles work for this problem:
```
INFO - all.sregym.conductor - ✅ Deployment complete. Ready for submission. Current stage is: noop
SREGym> submit("Yes")
INFO - all.sregym.conductor - Start Eval for Noop
INFO - all.sregym.oracle - == Detection Evaluation (expected: No) ==
INFO - all.sregym.oracle - ❌ Detection: Yes
== Fault Injection ==
INFO - all.infra.kubectl - ✅ Deleted ReplicaSet 'search-5c64c8f5bc' in namespace 'hotel-reservation'
[❌] Grading error: name 'dynamic' is not defined
SREGym> submit("Yes")
INFO - all.sregym.conductor - Start Eval for Noop
INFO - all.sregym.oracle - == Detection Evaluation (expected: No) ==
INFO - all.sregym.oracle - ❌ Detection: Yes
== Fault Injection ==
INFO - all.infra.kubectl - ✅ Deleted ReplicaSet 'search-5c64c8f5bc' in namespace 'hotel-reservation'
[❌] Grading error: name 'dynamic' is not defined
SREGym> submit("Yes")
INFO - all.sregym.conductor - Start Eval for Noop
INFO - all.sregym.oracle - == Detection Evaluation (expected: No) ==
INFO - all.sregym.oracle - ❌ Detection: Yes
== Fault Injection ==
INFO - all.infra.kubectl - ✅ Deleted ReplicaSet 'search-5c64c8f5bc' in namespace 'hotel-reservation'
[❌] Grading error: name 'dynamic' is not defined
SREGym> 
```

Jackcuii:
Fixed in  #347

PR #346, opened 2025-11-04
Fix #345 & old transient issue leftovers

### LINKED ISSUE #345: Fix loggers & Fix Dashboard (11/7)
Thanks Pial @Saadmrp1038 good catch!

Now the agent logger is messed up again.
And the dashboard does not work.

I found the RC is that in our new AgentLauncher, the agent is running in a different proc. 
So the configurations of the loggers are not applied to the agent proc. PR on the way

Jackcuii:
Fix #345 
And lost property bug in `Conductor`

### DIFF
--- a/dashboard/dashboard_app.py
+++ b/dashboard/dashboard_app.py
@@ -1,11 +1,11 @@
+import atexit
 import json
 import os
 import re
+import signal
 import sys
 import time
 from datetime import datetime
-import atexit
-import signal
 from pathlib import Path
 
 import dash
@@ -21,6 +21,7 @@ from collections import deque
 
 import yaml
 from flask import request
+
 from sregym.service.kubectl import KubeCtl
 
 DASHBOARD_URL = "http://127.0.0.1:11451"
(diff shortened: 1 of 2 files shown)


HacksonClark:
Let's fix #339 in a separate PR.

Jackcuii:
> The dashboard is not running when I ran `python3 main.py` after switching to this branch.

It should not runs automatically. It is an optional function and if we run it in as a sub proc, when the main proc exit , it will also be killed, then the browser will keep generate alerts. So we expect the user to start the dashboard by themselves if they want to get visualized dialog. And the agent user brings may not implement the logger report to the trace collector (or even can't implement, for example it could be a binary), so this could be an optional feature.

HacksonClark:
@Jackcuii I thought we updated it to run automatically? I don't really see a reason why not. The user can then choose if they want to use it or not.

yimingsu01:
> And the agent user brings may not implement the logger report to the trace collector (or even can't implement, for example it could be a binary), so this could be an optional feature.

I don't understand this part.

Regardless. Like Jackson said, we should make it into a feature and expose an interface (e.g., env var) for the user to turn it on. We can't expect users to reach into the code to turn this feature on, that's just too much.

HacksonClark:
@Jackcuii mentioned to me that the dashboard is very tightly coupled with Stratus -- so it wouldn't work with other agents out of the box. But I think this is completely alright as it's mostly to give the user a better demo experience. We can work on the dashboard more at a later time.

yimingsu01:
@Jackcuii what are the lines I need to change to enable the dashboard with Stratus, to test this PR?

Jackcuii:
> @Jackcuii what are the lines I need to change to enable the dashboard with Stratus, to test this PR?

Oh I just found the README is changed. You can just run `python3 dashboard/dashboard_app.py`

HacksonClark:
I was able to verify this fix with the dashboard. @Jackcuii can you add back to code to launch the dashboard automatically? I'll merge after that.

Jackcuii:
Okay~ By 17:00 :saluting_face:

Issue #317, opened 2025-10-22
Refactoring Noise Mechanism (11/19)


777lefty:
**This issue might be edited and updated.**

Followed our previous discussion on the noise, the random injection of chaos experiments, though seems natural as it simulated the nature of random of transient issues noise in real production environment, is not efficient in confusing agents. The main causes might be:
- The noise (chaos experiments) is not specifically targeted within the service that agent is currently inspecting: the previous version of injection provided injection scope control, maybe not fine-grained enough (?)
- The injection timing doesn't fit well into the process of interaction between agent and cluster, if injection are done when agents are not interact with the cluster, then it's invisible to the agent and then it's meaningless:  this is completely out of consideration before since we were to simulate randomness, but it seems that injecting at random time is inefficient

Based on the discussion and thoughts above, our new noise might need:
- Spatial injection control: Figure out (maybe for each problem) which are the part agents care about (inspect) most when solving the problem, and then specify what kind of noise (pod kill, cpu containment, etc.) could affect them most, and only inject this specific noise into the corresponding specific part (service or pod)
- Temporal injection control: We might need some hooks into the agent side to tell each time when agent calls a specific tool (like kubectl) to inspect a specific part of cluster, then we would inject the specific noise only when the cluster is being observed

Forms of noise:
- Self-healing transient issues in the cluster: Simulate some jitters of the system
- Fake logs/Log flooding: Although @HacksonClark mentioned before that we could use real transient issues only, I still think this might be effective and more direct
- Ghost metrics: Simulate a 'ghost service' that is already offline or never existed, but its metrics are maliciously (or erroneously) pushed to Prometheus
- Zombie resources: In real production cluster, there might be some long-lived, failed but not cleaned-up K8s resources in the cluster (like failed but not cleaned pods), we could simulate them via `CronJob` to create failed pods
- Expected (Periodical) Change & CI/CD Chatter, like:
  - High-frequency updates: Simulate an active development team that performs a rolling update on a service every certain time interval
  - Planned node maintenance: Simulate an SRE performing node maintenance, such as upgrading the Kubelet or applying a kernel patch, in this case, the process of: 1. tag a node as noschedule; 2. evict all the pods on it; 3. recover the node might be periodical executed

A gdoc to list potential forms of noises: https://docs.google.com/document/d/1ynCCka4NMPxIb7z7LkactOCpOyIVnCUw4nvbaSkK4Hs/edit?usp=sharing

Implementations of noise or useful tools:
[...]

PR #350, opened 2025-11-04
I-337

### LINKED ISSUE #337: stale_core_dns problems not cleaning up properly (11/7)
@lilygn found this while working on #333 

To reproduce, you just need to run one of the `stale_core_dns_config` problems. The first problem will execute normally and allow you to submit, but when you go to run the next problem you will see an error.

Here is an example of the bug when running the `stale_coredns_config_social_network` problem after the `stale_coredns_config_astronomy_shop` problem:

```
stream closed EOF for social-network/media-frontend-59bb995b98-xgf4b (alpine-container)
media-frontend 2025/11/03 18:00:29 [emerg] 1#1: host not found in resolver "kube-dns.kube-system.svc.cluster.local" in /usr/local/openresty/nginx/conf/nginx.conf:36
alpine-container Cloning into '/DeathStarBench'...
media-frontend nginx: [emerg] host not found in resolver "kube-dns.kube-system.svc.cluster.local" in /usr/local/openresty/nginx/conf/nginx.conf:36
```

I believe that it would impact nearly any problem that you run after it though, since DNS will no longer be working.

Saadmrp1038:
Fixes #337 

## Root Cause

The error occurs in `Conductor.submit()` method when the submission stage transitions to "done". At line 235, the code attempts to call:

```python
if self.transient_config["switch"]:
      self.transient_issue_generator.stop_continuous_injection()
```

However, `transient_issue_generator` is never initialized because the transient issues feature is disabled. So we get the following error:

```bash
ERROR - all.sregym.cli - [CLI] submit() raised exception: 'Conductor' object has no attribute 'transient_issue_generator'
  File "/home/pial/Academics/UIUC/Projects/SREGym/cli.py", line 117, in interactive_loop
    resp = await self.conductor.submit(wrapped)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pial/Academics/UIUC/Projects/SREGym/sregym/conductor/conductor.py", line 236, in submit
    self.transient_issue_generator.stop_continuous_injection()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Conductor' object has no attribute 'transient_issue_generator'
```

## Solution
Solution is very simple. Set `transient_config["switch"]` default to `False` to match the disabled feature state.

---

Recovery was failing for all the faults, but the issue manifested here because it modified something that does not go away on app cleanup.

### DIFF
--- a/sregym/conductor/conductor.py
+++ b/sregym/conductor/conductor.py
@@ -46,7 +46,7 @@ class Conductor:
         self.local_logger = logging.getLogger("all.sregym.conductor")
 
         self.transient_config = {
-            "switch": True,
+            "switch": False,
             "min_duration": 40,
             "max_duration": 60,
             "fault_types": [FaultType.FAIL_SLOW, FaultType.FAIL_STOP],


Saadmrp1038:
I think we need better error handling. This error not being captured anywhere is an issue.

HacksonClark:
@Saadmrp1038 I see. The transient issue generator (our noise maker) should be disabled at this point until we refactor #317

PR #353, opened 2025-11-04
Fix Transient issue in trainticket_f17_nested_sql_select_clause_error

### LINKED ISSUE #342: Transient issue in `trainticket_f17_nested_sql_select_clause_error` (11/7)
This will need some further investigation as it is not observed consistently.

> When I ran the Train Ticket Fault Injector Problems, on one run, I got, at this problem:           **trainticket_f17_nested_sql_select_clause_error**                       
> 
> `TypeError: KubeCtl.wait_for_job_completion() got an unexpected keyword argument 'name'
> `
> And the benchmark crashed. When I ran it again on the same node, I did not get the error and the benchmark continued on. 

 _Originally posted by @lilygn in [#333](https://github.com/xlab-uiuc/SREGym/issues/333#issuecomment-3482596955)_

Jackcuii:
Fix #342

### DIFF
--- a/sregym/service/apps/train_ticket.py
+++ b/sregym/service/apps/train_ticket.py
@@ -39,7 +39,7 @@ class TrainTicket(Application):
             self.kubectl.create_namespace_if_not_exist(self.namespace)
 
         Helm.install(**self.helm_configs)
-        self.kubectl.wait_for_job_completion(job_name="train-ticket-deploy", namespace="train-ticket")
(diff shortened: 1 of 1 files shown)


HacksonClark:
I made some changes to clean things up since we don't need 3 different functions that all do the same thing. I am having some trouble deploying this problem though so it's hard for me to fully test it. For some reason train-ticket isn't deploying (I tried on both kind and CloudLab) and it seems to fail differently on each. However, I think it's unrelated to this issue.

Jackcuii:
Gotcha :smiley:

PR #357, opened 2025-11-05
sequentially run all agents

777lefty:
modified the driver to support running all agents sequentially in agents.yaml

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -25,6 +25,7 @@ from dashboard.dashboard_app import SREGymDashboardServer
 from dashboard.proxy import LogProxy
 from logger import init_logger
 from sregym.conductor.conductor import Conductor
+from sregym.conductor.constants import StartProblemResult
 from sregym.service.shell import Shell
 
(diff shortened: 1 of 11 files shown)


HacksonClark:
@777lefty could you resolve the conflicts and write a brief note about how you tested it? (i.e. what agents.yaml did you use?) I remember from our discussion we would just test Stratus a couple of times.

jq-huang:
@HacksonClark Just fixed the conflict. To test it, just run `python main.py` with `agents.yaml` lije what below shows:
```
agents:
- name: stratus1
  kickoff_command: python -m clients.stratus.stratus_agent.driver.driver --server
    http://localhost:8000
  kickoff_workdir: .
  kickoff_env: null
- name: stratus2
  kickoff_command: python -m clients.stratus.stratus_agent.driver.driver --server
    http://localhost:8000
  kickoff_workdir: .
  kickoff_env: null
```
Namely we run stratus with different names for two times.

PR #363, opened 2025-11-07
Update repository references

Saadmrp1038:
Simple chore task. Update repository references from `xlab-uiuc/SREGym` to `SREGym/SREGym`. Noticed it when working on the docs.


HacksonClark:
This would have caused some issues later...

PR #383, opened 2025-11-19
376 auto submit as an agent 1121

lilygn:
Created an autosubmit agent. The autosubmit agent submits yes every minute.


yimingsu01:
@HacksonClark Tested on two consecutive problems and it submits well. This is mainly a debugging tool and can be tuned anyway wanted. Looks great to me.

PR #385, opened 2025-11-20
I264: Latent Sector Error

Saadmrp1038:
Closes #264 

## Problem

Targeting random blocks for the LSE error was not causing any visible symptoms. The previous approach injected bad blocks in sectors that MongoDB never accessed during normal operation, and the WiredTiger cache kept data in memory, avoiding disk reads where bad blocks were located.

## Current Implementation

This PR implements a targeted approach that causes consistent pod crashes. Instead of injecting bad blocks at random sectors, we now use `filefrag` to identify the actual sectors where MongoDB's critical data files reside (WiredTiger.wt metadata file and journal files). Bad blocks are injected at these specific locations, ensuring MongoDB will encounter them.

Additionally, we drop page caches after injection to force disk reads, ensuring the bad blocks are hit immediately. The fault manifests as a pod crash when it attempts to read corrupted blocks, resulting in a `CrashLoopBackOff` state that is easily detectable.

```bash
{"t":{"$date":"2025-11-19T23:27:04.655+00:00"},"s":"E",  "c":"STORAGE",  "id":22435,   "ctx":"initandlisten","msg":"WiredTiger error","attr":{"error":5,"message":"[1763594824:655663][1:0x7868db13dac0], file:WiredTiger.wt, connection: __posix_file_read, 431: /data/db/WiredTiger.wt: handle-read: pread: failed to read 4096 bytes at offset 0: Input/output error"}}
{"t":{"$date":"2025-11-19T23:27:04.671+00:00"},"s":"E",  "c":"STORAGE",  "id":22435,   "ctx":"initandlisten","msg":"WiredTiger error","attr":{"error":5,"message":"[1763594824:671604][1:0x7868db13dac0], file:WiredTiger.wt, connection: __posix_file_read, 431: /data/db/WiredTiger.wt: handle-read: pread: failed to read 4096 bytes at offset 0: Input/output error"}}
```

## Results

The fault now consistently manifests as `mongodb-geo` pod crashes after fault injection (takes 15-20s to take effect), with the pod entering `CrashLoopBackOff` state.

## Notes

This PR also updates the `setup_cluster.yml` script to install the Ubuntu generic 6.8 kernel and matching headers so that we can support `dm-dust` and `dm-flakey` modules. You will need to run this script on a fresh cluster to get ready.

### DIFF
--- a/sregym/conductor/problems/latent_sector_error.py
+++ b/sregym/conductor/problems/latent_sector_error.py
@@ -23,6 +23,7 @@ class LatentSectorErrorStrategy(StrEnum):
 
     TEST = "test"
     EVERY_1000 = "every_1000"  # Also test strategy
+    TARGETED = "targeted"
 
 
 class LatentSectorError(Problem):
@@ -35,7 +36,7 @@ class LatentSectorError(Problem):
         self,
         target_deploy: str = DEFAULT_TARGET_DEPLOY,
         namespace: str = DEFAULT_NAMESPACE,
-        strategy: LatentSectorErrorStrategy = LatentSectorErrorStrategy.EVERY_1000,
+        strategy: LatentSectorErrorStrategy = LatentSectorErrorStrategy.TARGETED,
     ):
         self.app = HotelReservation()
         self.kubectl = KubeCtl()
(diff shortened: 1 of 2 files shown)


HacksonClark:
This is *badass*.

tianyin:
I'm also impressed. @Saadmrp1038 great work!

PR #386, opened 2025-11-20
Hotfix for two bugs

777lefty:
1. In conductor, currently we are using
```
if hasattr(self.problem, "localization_oracle"):
                self.problem.localization_oracle.load_localization_checkpoint()
```
But in the base class of problem, problems are initialized with `localization = None`, so even if the specific problem class doesn't provide localization_oracle, it will enter the branch, causing error. This is fixed with
```
if hasattr(self.problem, "localization_oracle") and self.problem.localization_oracle is not None:
                self.problem.localization_oracle.load_localization_checkpoint()
```

2. In workload_oracle for stratus, `self.kubectl.wait_for_job_completion(job_name=job_name, namespace="default")`, which is not async function, is called with `await`

### DIFF
--- a/sregym/conductor/conductor.py
+++ b/sregym/conductor/conductor.py
@@ -181,7 +181,7 @@ class Conductor:
 
             self.logger.info(f"[ENV] Injected fault")
 
-            if hasattr(self.problem, "localization_oracle"):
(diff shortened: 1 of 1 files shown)


HacksonClark (approved):
LGTM

PR #387, opened 2025-11-22
I273: Silent Data Corruption

Saadmrp1038:
## Overview

The SDC implementation is complete. We leverage the `random_read_corrupt` and `random_write_corrupt` options of `dm-flakey` to corrupt data written to or read from disk with a configurable probability.

## Implementation Details

The implementation uses `dm-flakey`'s corruption features:
- `random_read_corrupt`: Corrupts data during read operations with specified probability
- `random_write_corrupt`: Corrupts data during write operations with specified probability

This allows us to inject silent data corruption into any data that is written to or read from the affected disk.

## Testing Results

I have run some tests to check the behavior of the current implementation:

### 1. Disk Corruption

The SDC is functional. I wrote a custom script to write and read a random file on the affected disk. When the fault is active, the data is corrupted properly.

**Test Output:**
```bash
========================================
=== Testing node2.exp-722006.aiopslab-pg0.wisc.cloudlab.us ===
========================================
--- dm-flakey configuration ---
0 10485760 flakey 7:0 0 0 1 4 random_read_corrupt 1000000000 random_write_corrupt 1000000000

--- Corruption test (1MB random data) ---
Source MD5: c38a10e1843fe63ea6a23cbdebd0d49a
Read MD5:   b8425b1e7682a1160802ebaf8c01c69d

Result: ✗ CORRUPTED (MD5 mismatch)
✗ Corruption detected or test failed on node2.exp-722006.aiopslab-pg0.wisc.cloudlab.us
```

**Result:** Corruption is successfully injected and detected via MD5 mismatch.

### 2. MongoDB Pods Crash

When forcing MongoDB to read data from disk (by dropping the cache), MongoDB pods crash due to reading corrupted data. The crashes manifest as:
- **WiredTiger storage engine corruption**: MongoDB detects corrupted WiredTiger data files (Exit code 14)
- **Filesystem corruption**: MongoDB encounters corrupted filesystem metadata (Exit code 1)

**Observed Behavior:**
- With 100% corruption probability and cache drop: MongoDB pods crash immediately
- With lower corruption probability and without cache drop: Effects are harder to detect as pods don't crash instantly. Also, API calls from the workload hit the cache and return valid data.

---

## Possible Paths

Based on these behaviors we can go about it in a couple of ways:
1. We can force the pods to crash (high probability of corruption, cache clear to force disk reads) in which case the effect of the fault will be immediate and pods will crash.
2. We can keep it as it is in which case there might not be any immediate detectable effects (unless of course someone directly try to read/write on the disk).

### DIFF
--- a/sregym/conductor/problems/silent_data_corruption.py
+++ b/sregym/conductor/problems/silent_data_corruption.py
@@ -226,9 +226,23 @@ class SilentDataCorruption(Problem):
             )
             print(f"[SDC] ✅ dm-flakey device restored to normal operation")
         
-        # Clean up and redeploy the app to get rid of any corrupted data
+        # Clean up and redeploy the app
         self.app.cleanup()
-        self.dm_flakey_manager.setup_openebs_dm_flakey_infrastructure()
+        
+        try:
+            cleanup_pods = self.kubectl.exec_command(
+                "kubectl get pods -n openebs --no-headers | grep 'cleanup-pvc-' | awk '{print $1}'"
+            ).strip()
+            if cleanup_pods:
+                pod_list = [p for p in cleanup_pods.splitlines() if p.strip()]
+                for pod in pod_list:
+                    # Delete failed cleanup pods
+                    self.kubectl.exec_command(f"kubectl delete pod -n openebs {pod} --ignore-not-found")
+                print(f"[SDC] Cleaned up {len(pod_list)} OpenEBS cleanup pod(s)")
+        except Exception as e:
+            print(f"[SDC] ⚠️  Warning: Failed to clean up OpenEBS cleanup pods: {e}")
+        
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
@HacksonClark need some suggestions here.

HacksonClark:
@Saadmrp1038 sure, what's up?

Saadmrp1038:
@HacksonClark There isn't really a clean way to get consistent, detectable symptoms here. We can either keep it realistic (in which case there's no guarantee if or when the agent would detect any error), or we can force the pods to crash by forcing corrupted data read (cache drop).

HacksonClark:
@Saadmrp1038 let's force it for now.

Saadmrp1038:
Got it. 100% corruption rate and cache drop does the trick.

PR #390, opened 2025-11-25
Using codeql to get rid of trivial bugs

lilygn:
As of right now, we have code ql queries for checking if problem attributes are null or missing.
 These problem attributes include:

- self.app
- self.nullspace
- self.faulty_service
- mitigation and detection oracles

In addition, we have queries for checking if every fault is recovered and if the inject() and recover() functions are marked with the  `@mark_fault_injected` decorators.
Checks if there is a workload or port forward process that is started but not stopped.
Checks if namespace is created and not deleted.
Checks if application is deployed but not deleted/cleaned up.
Detects applications that do not load helm configs. 

Planning to add more features. Please let me know if I should change or add anything thing. We ultimately plan to add it to the GitHub Action. 

[draft_PR.csv](https://github.com/user-attachments/files/23738693/draft_PR.csv)

### DIFF
--- a/tests/sre-ql/queries/application_attributes/application_checks.ql
+++ b/tests/sre-ql/queries/application_attributes/application_checks.ql
@@ -11,9 +11,6 @@ class ApplicationSubclass extends Class {
   ApplicationSubclass() {
     // Direct inheritance from Application
     this.getABase().(Name).getId() = "Application"
-    or
-    // Transitive inheritance (inherits through other classes)
-    this.getABaseClass+().getName() = "Application"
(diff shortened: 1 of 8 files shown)


yimingsu01:
@lilygn Amazing progress!!

yimingsu01:
@lilygn next step is incorporate all these things into a github action

yimingsu01:
@lilygn great job! This is a good start in building a CI pipeline for SREGym!

yimingsu01 (approved):

PR #391, opened 2025-11-27
Add GitHub Actions workflow for SRE-QL queries

lilygn:
CodeQL to be ran upon pushes/pr requests

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -80,5 +80,6 @@ jobs:
           echo "Found $ISSUE_COUNT issues"
           if [ "$ISSUE_COUNT" -gt 0 ]; then
             echo "::warning::Found $ISSUE_COUNT SRE reliability issues"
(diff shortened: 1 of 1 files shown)


yimingsu01:
@lilygn It looks like the check is failing because it's not being set up correctly. Please fix the errors first.

lilygn:
It fails because my CodeQL is not merged in main.

lilygn:
Failing because Github action does not accept warnings. Warnings cause it to fail. Fixing warnings now.

lilygn:
It is supposed to fail when one of the queries reports a bug. This works.

yimingsu01:
@lilygn Impressive job as always. Thank you for kicking start our CI process.

yimingsu01:
@lilygn Let's monitor the first few PR/pushes in the next days to see if this workflow works well or not.

PR #396, opened 2025-12-01
Slight modification for System Intelligence

Jackcuii:
We need to make some modifications for the System Intelligence framework.

I will push some modifications here.

I will mark it ready once all the modifications are done.

Modifications:
- modify main.py to allow triggering the running through direct call of `main`
- add agent selection to main
- refactor the model selection / separate the llm_backend
- fix bugs in misconfig_app
- fix duplicated logger

I think I can just start squashing commits from next PR, since this one is actually pushed from the subtree of the System Intelligence repo. So force push could cause something strange to happen.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,3 +38,6 @@ LLM_QUERY_INIT_RETRY_DELAY=1
 # WATSONX_API_BASE="https://us-south.ml.cloud.ibm.com"
 # WX_PROJECT_ID="HaveIllinisLotsOfIt"
 # WATSONX_API_KEY="HaveCornsLotsOfIt"
+
+# AZURE_API_KEY="HaveBluesLotsOfIt"
+# AZURE_API_BASE="HaveBasesLotsOfIt"


Jackcuii:
- Clean up the llm backend.  Delete lots of unused vars
- Use pre-defined configurations to make it totally clear. 
- Separate the LLM Backend from the agent. 
- Expose the LLM selection, not use `.env` to select, `.env` only provide keys
- Add a form to describe the LLM configurations.

HacksonClark:
@Jackcuii Is this PR ready to review? It says open.

Jackcuii:
> @Jackcuii Is this PR ready to review? It says open.

@HacksonClark  Yes! I believe this version is the most neat and clear solution. I added some pre-defined configs, and the users can choose among them in the cli or add some. (also included a form to describe parameters). The only used env for model selection is `MODEL_ID` and it is not needed to be set by the users. This avoid the mix of different configurations.

Configurations:
https://github.com/SREGym/SREGym/pull/396/files#diff-5e12f6853ef6b0b80226d149ad482519f72b33aa3634db17d77855068942c133

I also extract the backend from the agent, so that if the users also want to reuse this backend, the CLI selection is also valid for their agent. And the LLM-as-a-judge also use this backend, so it is logically proper make it individual.

HacksonClark:
@Jackcuii Okay, how should I go about testing this? I know we will need to use the new system intelligence framework, but I haven't used that before.

Jackcuii:
> @Jackcuii Okay, how should I go about testing this? I know we will need to use the new system intelligence framework, but I haven't used that before.

I think Xuan, Cheng and Bo have tested it on the system-intelligence side usability :smiley: They found everything goes well. We just need make sure with this adaptions we can still run it well independently. :cowboy_hat_face: so I think we can just `python3 main.py` or with some model params

HacksonClark:
@Jackcuii Is there a PR on their side or something I can review? I want to make sure the implementation is clean there as well.

Jackcuii:
> @Jackcuii Is there a PR on their side or something I can review? I want to make sure the implementation is clean there as well.

Yep! Here it is: https://github.com/sys-intelligence/system-intelligence-benchmark/pull/30
Our PR is actually an upstream subtree PR from the https://github.com/sys-intelligence/system-intelligence-benchmark/pull/30

You can also try this: https://github.com/SREGym/system-intelligence-benchmark main branch
They are the same

Jackcuii:
Resolves #396

PR #397, opened 2025-12-04
Add warning for reboot

Jackcuii:
I think we need to warn the users that the setup will reboot the machines (it is a critical behavior).


HacksonClark:
I don't think this is really that relevant or worth the space on the README.

PR #398, opened 2025-12-09
I388: Implement all khaos problems

### LINKED ISSUE #388: Implement all khaos problems
We currently only have `read_error` from khaos implemented, but there are many more problems in khaos. Let's implement all of them in SREGym!

Saadmrp1038:
Closes #388 
1. Adds a single Khaos fault wrapper that covers all Khaos syscall faults.  
2. Remove the dedicated `read_error` problem file. 
3. Updates the hardware fault injector to pass optional parameters through to Khaos when injecting.


HacksonClark (approved):
This looks great! I ended up disabling most of the problems because I think we should change how khaos works quite a bit. But that will be a separate issue.

PR #427, opened 2026-01-02
fix: move super init before initializing root cause

JothamWong:
The current codebase incorrectly calls `super().__init__(...)` after setting the root cause. This is incorrect because the call to the super constructor resets the root cause back to None. As a result, the LLM-as-a-Judge will incorrectly evaluate the agent's diagnosis as a false positive.

This PR simply moves all calls to the super constructor to before the root cause is set.


HacksonClark (approved):
Great catch!

PR #475, opened 2026-01-16
Implement proxy to hide disallowed namespaces

### LINKED ISSUE #450: Add RBAC policy for agent execution
From some of our preliminary results and trajectories, we can see that the agent does a good job of spotting chaos engineering tools such as ChaosMesh and our `khaos` tool. We should add an RBAC policy for the executing agent so that way these chaos engineering tools are hidden from the agent.

HacksonClark:
Closes #450 however, it doesn't use RBAC.

We implemented a lightweight HTTP proxy that sits between agents and the Kubernetes API server. The proxy intercepts API responses and filters out chaos-mesh and khaos namespaces, preventing agents from discovering that faults are being injected via chaos engineering tools, which we couldn't do with just RBAC since RBAC would reveal the names, and chaos-mesh is a well known tool that we've seen the agents recognize.

### DIFF
--- a/main.py
+++ b/main.py
@@ -61,6 +61,11 @@ def driver_loop(
             console.log(f"Starting agent now: {agent_to_run}")
             conductor.register_agent(agent_to_run)
 
+            # Start K8s API proxy to hide chaos engineering namespaces from the agent
+            console.log("🔒 Starting Kubernetes API proxy to hide chaos namespaces...")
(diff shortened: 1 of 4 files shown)


HacksonClark:
Tested successfully with `read_error`, the `khaos` namespace is effectively hidden from the agent.

Issue #436, opened 2026-01-05
Add new flagd faults


HacksonClark:
Astronomy-shop has some new feature flag faults: https://opentelemetry.io/docs/demo/feature-flags/

We should add the following:
- emailMemoryLeak
- llmInaccurateResponse
- llmRateLimitError

PR #481, opened 2026-01-17
 Extend Astronomy-shop with new `flagd` flags

bastoica:
Addressing #436 

### PR Description

Adds new flags to `Astronomy-shop`.

### Changes
* Adds `set_flag` and `get_flag_status` method along with helpers
* Extends `AstronomyShop` with private constant fields for `flagd` metadata (TODO: this is hardcoded, needs to be discoverable)

### Testing
* Tested via `cli.py`


HacksonClark:
Is this a draft? I saw that you requested my review. Is it in progress or ready to review?

bastoica:
Sorry, @HacksonClark !! I reverted to "draft mode" since I discovered it actually breaks... I'll ping you here when is ready for review, most likely within the next 24-36h. Thanks

HacksonClark:
@bastoica sounds great!

HacksonClark:
After taking a peek at your implementation, I would advise you to look at the other astronomy shop problems as well as the `inject_otel` file, as that's what we use to trigger the feature flags.

bastoica:
Ah... makes sense. Let me take a look now before I push more commits :-)

bastoica:
@HacksonClark Followed your suggestion and implemented the problems using the `OtelFaultInjector` and an llm-as-a-judge oracle to be consistent with the other `AstronomyShop` problems.

I just realized that I implemented a `set_flag` method here: https://github.com/SREGym/SREGym/blob/c996f6047da9c25258f5484a6b3487b1feb740d8/sregym/service/apps/astronomy_shop.py#L83
but we want to actually flip those flags using the fault injection.

Should I remove that method or do you think we may think of a scenario where these flags can be set via metadata as well?

bastoica:
Sorry, I don't know why this was in draft mode.... Changed it.

PR #498, opened 2026-01-22
Cleanup stale coredns

### LINKED ISSUE #496: stale_coredns fault persisting after problem
I believe I already have a fix for this.

Basically, the fault from stale_coredns persisted after the problem was complete. So, future social-network pods were not able to be deployed. We should ensure this is cleaned before running a problem.

HacksonClark:
Closes #496


Saadmrp1038:
@HacksonClark, also add the `service_dns_resolution` fault recovery, just in case, since that modifies `kube-system` too.

PR #509, opened 2026-01-27
fix: Remove faulty_service and mitigation_oracle CodeQL checks

### LINKED ISSUE #508: Fix CodeQL Script
The CodeQL code that checks the problem format isn't entirely correct.

Here is a log:
```
cat results.csv | cut -d, -f4,5
"NO self.faulty_service defined","/sregym/conductor/problems/khaos_faults.py"
"NO self.mitigation_oracle defined","/sregym/conductor/problems/email_memory_leak.py"
"NO self.mitigation_oracle defined","/sregym/conductor/problems/khaos_faults.py"
"NO self.mitigation_oracle defined","/sregym/conductor/problems/llm_inaccurate_response.py"
"NO self.mitigation_oracle defined","/sregym/conductor/problems/llm_rate_limit_error.py"
```

However, not every problem will have a mitigation oracle. Also, `faulty_service` is no longer required as we don't have a localization task anymore.

HacksonClark:
Fixes #508

Removed two CodeQL queries that were producing false positives:

- `faultyservice.ql`: faulty_service is no longer required since localization task was removed
- `mitigation_oracle.ql`: mitigation_oracle is optional and not all problems need one


yimingsu01:
I don't think we should delete `mitigation_oracle.ql` entirely, it's a good sanity check to me. @lilygn I remember when we are working on this issue, you were able to exclude some problems that don't have mitigation phase. Could you check and add these new problems to that exclude list?

yimingsu01 (approved):

PR #515, opened 2026-01-28
Type-check SREGym

yimingsu01:
A lot of trivial bugs (e.g., missing imports or class attributes) can be avoided if we type check our implementation. This PR slowly adds any type annotations needed to the repo to make ruff and pyright happy.


yimingsu01:
Not planned.

PR #519, opened 2026-01-29
Concurrent fault problems

yimingsu01:
This PR implements problems that are compositions of existing problems.

This is not an exhaustive list of problems. This 5 problems are for preliminary explorations of a potential fault composer idea.

### DIFF
--- a/sregym/conductor/problems/registry.py
+++ b/sregym/conductor/problems/registry.py
@@ -33,6 +33,7 @@ from sregym.conductor.problems.misconfig_app import MisconfigAppHotelRes
 from sregym.conductor.problems.missing_configmap import MissingConfigMap
 from sregym.conductor.problems.missing_env_variable import MissingEnvVariable
 from sregym.conductor.problems.missing_service import MissingService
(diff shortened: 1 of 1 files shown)


sregym/conductor/problems/registry.py
```diff
@@ -33,7 +33,6 @@
 from sregym.conductor.problems.missing_configmap import MissingConfigMap
 from sregym.conductor.problems.missing_env_variable import MissingEnvVariable
 from sregym.conductor.problems.missing_service import MissingService
-from sregym.conductor.problems.multiple_failures import MultipleIndependentFailures
```
HacksonClark: Why is this removed? I had to add it back in to try and test the problems.

HacksonClark (changes requested):
Seems like some things are still missing.

yimingsu01:
@HacksonClark I commented out for now because 
1. we haven't figured out a principled way to compose individual faults, 
2. it's hard to make any valuable conclusions on the agent when we don't understand how each of these problems will manifest on the cluster.

yimingsu01:
Basically, I don't want those problems to be in the next e2e run because we haven't understood them enough, so I commented them out for now

PR #527, opened 2026-01-31
Create alert rules for Prometheus

yimingsu01:
This PR adds pod-level status alerts in Prometheus. We enable alertmanager and kube-state-metrics in Prometheus' `values.yaml` to monitor the pod states .After deploying the problems and port-forward Prometheus, you can see the alerts in `http://localhost:9090/alerts`.

Currently, we only monitor pod status. After this PR is merged, we should look into:
1. Add alert oracle to stratus
2. add app-specific alerts. (and maybe instrument the application to be queried by prometheus).


yimingsu01:
Will mark as ready for review once I added all error pod status to the alert. This serves as the first alert of Gym and we can add more later.

PR #533, opened 2026-02-05
Add LLM judge oracle to kubelet_crash problem

### LINKED ISSUE #531: `kubelet_crash` is not a "problem"
We don't have any oracles in `kubelet_crash`, so the problem will have its application deployed, fault injection phase skipped, and immediately torn down, as we skip the problem if there's no oracles attached.

This is being handled together with #523 by me.

yimingsu01:
This adds an LLMAsAJudgeOracle to the kubelet_crash problem, which will allow the problem to be tested with agents. Previously, kubelet_crash had no oracles attached, causing it to skip fault injection and immediately tear down.

The oracle uses the existing root_cause description to evaluate whether agent diagnoses correctly identify the kubelet daemon crash issue.

Fixes #531

### DIFF
--- a/sregym/conductor/problems/kubelet_crash.py
+++ b/sregym/conductor/problems/kubelet_crash.py
@@ -1,5 +1,6 @@
 import time
 
+from sregym.conductor.oracles.llm_as_a_judge.llm_as_a_judge_oracle import LLMAsAJudgeOracle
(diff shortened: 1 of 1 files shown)


HacksonClark (approved):
Looks great!

PR #535, opened 2026-02-06
Instruments API server with request auditing

yimingsu01:
Part of efforts in making problems idempotent.

Quoting myself:

> We mean two things when we want to make problems idempotent:
> 1. restore any deleted k8s objects.
> 2. remove any agent-created objects.

1 is easily achievable with `velero`, this PR is the first step towards 2: we need to understand what's been added before being able to remove them.

This PR adds a basic audit policy that traces requests to the API server, which will save all API requests to the API server in the log `/var/log/kubernetes/audit.log`. We will leverage this to record what has the agent created and remove them later.


yimingsu01:
Cluster cleanup done.

PR #548, opened 2026-02-09
Made MCP server port consistent to default port in code

yimingsu01:
`stratus` fails silently if the agent tool's port not the port that MCP server is listening on.


Saadmrp1038:
I also faced this today. Wasted so much time 🤦

PR #565, opened 2026-02-13
fix: Return HTTP response immediately without waiting for fault recovery

### LINKED ISSUE #244: Submission API does not return HTTP response until other logic finishes
@HacksonClark 

Consider the noop phase. If fault injection for a problem takes a long time, the submission HTTP call will not receive a response until the fault injection finishes. In cases where we are running multiple problems at the same time (i.e., concurrent faults), fault injection can take a while.

Same case for fault recovery, which can take minutes to finish depending how many problems we are running at the same time.

I don't think the HTTP response should wait for fault injection/recovery to finish.

HacksonClark:
This change prevents the submission API from blocking while performing slow operations like fault injection/recovery and app teardown. These operations can take minutes when running multiple problems concurrently.

## Changes

- Add async cleanup method that performs fault recovery, undeploy, and cluster reconciliation in the background
- Update `_finish_problem` to set state to "tearing_down" immediately and schedule async cleanup
- Modify `submit()` to capture results snapshot and return immediately
- Prevent noise restart during teardown state
- Add "tearing_down" to possible stages in API documentation

Fixes #244

---


Saadmrp1038:
Found an error while working.
```
2026-02-14 05:14:46,935 - clients.stratus.tools.submit_tool - INFO - HTTP submission failed: {'status': 'ok', 'text': 'Submission received'}
```

The submit tool (SREGym/clients/stratus/tools/submit_tool.py) expects status to be 200.
```
if result["status"] != "200":
        logger.info(f"HTTP submission failed: {result}")
```

yimingsu01:
@HacksonClark I'm testing this with a MS problem and a normal problem

HacksonClark:
Closes #575

yimingsu01:
@HacksonClark can confirm that the previous broken MS problem now can correctly finish (`capacity_decrease_rpc_retry_storm`)

yimingsu01:
Closes #585

yimingsu01:
Testing three problems tonight to merge this.
- [x] capacity MS problem
- [x] one tidb problem
- [x] one normal old problem for regression

PR #589, opened 2026-02-23
Add demo agent to execute kubectl_cmds.txt sequentially

SathsaraniJA:
This PR adds a lightweight CLI-based demo agent that reads a kubectl_cmds.txt file and executes kubectl commands sequentially on the active cluster context.

The goal is to provide a deterministic demo mode that does not rely on an LLM during live demos.

### DIFF
--- a/demo_agent.py
+++ b/demo_agent.py
@@ -1,0 +1,65 @@
+import argparse
+import subprocess
+import sys
+
+
+def read_commands(file_path):
+    with open(file_path, "r") as f:
+        lines = f.readlines()
+
+    commands = []
+    for line in lines:
+        line = line.strip()
+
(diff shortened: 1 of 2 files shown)


SathsaraniJA:
Hi! Please let me know if any changes are needed. Happy to improve it.

yimingsu01:
@SathsaraniJA thanks for your PR. 

Apologies if I wasn't clear in my original comment. I would like an agent that is semantic equivalent to the `stratus` agent in the repo, but instead of using an LLM to output `kubectl` commands, we deterministically output commands by a predefined sequence. So that we can also have an agent trace at the end of a benchmark run.

HacksonClark:
Closing

Issue #588, opened 2026-02-23
Add a demo agent


yimingsu01:
We have a lot of demo opportunities. Relying on an LLM in the actual demo might be risky. Let's create a demo agent that reads a `kubectl_cmds.txt` (lines of `kubectl` commands) and execute them on the cluster turn-by-turn like a real agent.

I would like an agent that is semantic equivalent to the stratus agent in the repo, but instead of using an LLM to output kubectl commands, we deterministically output commands by a predefined sequence. So that we can also have an agent trace at the end of a benchmark run.

SathsaraniJA:
Hi, I would like to work on this issue. Can you assign it to me?

SathsaraniJA:
Thanks! I’ll start working on it and open a PR soon. I’ll also join the Slack if I have questions

yimingsu01:
@SathsaraniJA @Jsze30 Please see my comment above.

Saadmrp1038:
Resolved in #634 (merged). Closing this issue.

PR #590, opened 2026-02-23
Added demo_agent for scripted kubectl demo #588

Jsze30:
This PR adds a demo agent script that runs pre-defined kubectl commands in kubectl_cmds.txt

- Executes pre-scripted kubectl commands for reliable demos
- Works with any agent type and model
- Supports configurable pause between steps
- Continues on error or stops on first failure
- Generates summary of command execution results
- Run using python demo_agent.py

### DIFF
--- a/demo_agent.py
+++ b/demo_agent.py
@@ -1,0 +1,239 @@
+"""
+Demo Agent that executes pre-scripted kubectl commands from a file.
+
+This agent reads commands line-by-line and executes them sequentially.
+Works with any agent type and model.
+"""
+
+import argparse
+import logging
+import subprocess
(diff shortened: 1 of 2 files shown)


yimingsu01:
Hi @Jsze30 , thanks for your PR.

We want a real agent similar to the `stratus` agent we have in the repo. Please see my comment here: #588

yimingsu01:
I will close this PR for now, as it does not meet the requirements described in my comment above.

yimingsu01:
@Jsze30 I also recommend you to look at other issues in the repo, as this one is taken already, see the original issue.

PR #592, opened 2026-02-24
Added tidb operator wrong container image problem

yimingsu01:
Added a simple problem, misconfiguring tidb operator container image


Saadmrp1038:
Works great!
Made a small change of replacing the `subprocess` calls with `kubectl.exec_command`

PR #603, opened 2026-02-25
fix: exit benchmark problem when Stratus hits retry ceiling

### LINKED ISSUE #593: Stratus Infinite Retry
When running the problems, I was alarmed that the problem was taking over 1 hour to complete. Then I realized that it was stuck in an infinite loop. The agent would run out of attempts, and the benchmark would force the agent to submit:
`[94mINFO - all.stratus.mitigation - Should we force the agent submit? Yes![0m
`
The agent would submit but then it would enter another loop of 
"You are in the thinking stage" and continue to mitigate the problem. This is not problem dependent and seems to be an agent issue.

HacksonClark:
When an agent exhausts its step limit (max_step), should_submit_router triggers force submit. If the LLM fails to call the submit tool, submitted stays False while num_steps remains at max_step. On the next while-loop iteration, the router fires again immediately, and this repeats infinitely.

Adds a num_steps >= max_step guard in the arun loops of BaseAgent, DiagnosisAgent, and MitigationAgent to break and fail the benchmark problem instead of looping forever.

Closes #593

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,6 +1,16 @@
 name: Run SRE-QL Queries
 on:
   pull_request:
+    paths-ignore:
+      - '*.md'
+      - 'docs/**'
+      - 'LICENSE'
+      - '.github/workflows/claude.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   codeql-custom:
     runs-on: ubuntu-latest
@@ -9,32 +19,32 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
-    
+
     - name: Download CodeQL CLI v2.23.5
       run: |
         curl -L https://github.com/github/codeql-cli-binaries/releases/download/v2.23.5/codeql-linux64.zip -o codeql.zip
         unzip codeql.zip
         mv codeql /opt/codeql
         echo "/opt/codeql" >> $GITHUB_PATH
-    
+
     - name: Verify CodeQL Installation
       run: |
         codeql --version
-    
+
     - name: Install CodeQL Packs
       run: |
         cd tests/sre-ql
         codeql pack install
         cd ../..
-    
+
     - name: Create CodeQL DB
       run: |
         codeql database create sreql-db \
           --language=python \
           --source-root=. \
           --overwrite
-    
-    - name: Run Queries 
+
+    - name: Run Queries
       run: |
         QUERIES=$(find tests/sre-ql/queries -name "*.ql" ! -name "problem_lifecycle.ql")
         echo "Running queries:"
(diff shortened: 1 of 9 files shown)


HacksonClark:
@lilygn Could you take a look at this implementation?

lilygn:
The code seems to fix the issue. However, I do not know how to test this since this issue comes up randomly. @HacksonClark How do you recommend that I verify the fix if I cannot reproduce it reliably?

HacksonClark:
@lilygn The amount of attempts stratus gets is configurable, so just set a lot step limit so that code path gets ran quickly. We would mostly see this happen with tiny (so not good models) that could never solve the issue.

So just set the retry to something small like 5 instead of (I think 30?) is the value we're currently using.

lilygn:
@HacksonClark  I set max steps to 1 and max retry attempts to 1. However, the agent still reached loop 9.

HacksonClark:
@lilygn what happened at loop 9?

lilygn:
The root cause may be correlated to the f_submit_tool.  I noticed that this retry loop happens when the agent is forced to submit but the "f_submit_tool" does not exist. And then the agent is back to being able to call whichever tool it wants to continue to mitigate the issue. In this case, the agent tried submitting but `Tool f_submit_tool does not exist!`. After that, it was reset to be back in the thinking stage and kept retrying.

lilygn:
@HacksonClark The agent chose to submit without being forced to at loop 9. It believed that it mitigated the problem and submitted

HacksonClark:
@lilygn seems like the bug is still there. Can you work on this?

lilygn:
Yes, I can work on it after my exam (Friday) @HacksonClark

Saadmrp1038:
@lilygn You'll need to use the `--force-build` flag for the new agent changes to take effect. Otherwise, the old docker image with the old code will be run.
Maybe we should add an automatic way to rebuild the image whenever relevant files are updated :)

lilygn:
@Saadmrp1038 Yeah, Yiming showed me :) I have been running the benchmark with force build.

Saadmrp1038:
Nice. I sometimes forget myself :). Need to think about whether something can be done here for better user experience.

lilygn:
Claude's fix did not update the benchmark stage, it only broke from the for loop. Therefore, the fix was to manually submit to advance the benchmark stage. Otherwise, the benchmark stage would stay at diagnosis rather than advancing to the expected mitigation stage, making the mitigation agent not run. Note: this fix does not allow the agent to retry submission. This fix is viable for now, given @yimingsu01 and I plan to rewrite the agent.

lilygn:
@yimingsu01 could you review?

HacksonClark:
@lilygn I have an upcoming PR that introduces a time limit on agent execution. So most of these issues shoul dgo away soon.

Also, if a PR is ready for review just mark it as "ready for review"

PR #609, opened 2026-02-26
Simplify genictl

Saadmrp1038:
Stripped out the K8s cluster bootstrapping, SSH remote execution, and hardware scraping logic from genictl. Provisioning now just creates slice + sliver on a specified CloudLab site.

I don't really use the CloudLab UI and still use this tool to provision CloudLab clusters. Since there's no other use for it except that (I'm pretty sure no one else uses this), removing the other stuff implemented a long time ago to only keep the cluster management commands. It doesn't touch any of our core benchmark code so there should not be any issues.

---
Also, can anyone confirm if `scripts/terraform` is functional. Otherwise, we should remove/refactor it.


yimingsu01:
I don't think we use any terraform stuff, so those are safe for removal

yimingsu01 (approved):

PR #613, opened 2026-02-26
Cluster cleanup heuristic

### LINKED ISSUE #460: Cluster cleanup in between problems
One important thing brought up during the discussion with Resolve today is was how the environment is cleaned between problems. While our benchmark does clean up the faults that were injected after an agent submission by running `recover_fault`, we don't handle any issues that may have been introduced by the agent itself. So, it's not a truly fresh state.

Earlier we tried to snapshot and restore etcd but we found it too slow. However, I feel there may be a better solution to get back to a "clean" state from before a problem is ran.

HacksonClark:
Not a perfect cleanup, but I think it may be good enough.
Closes #460


HacksonClark:
I'm going to try and test this by running my remaining eval problems

PR #616, opened 2026-02-27
Added CI

### LINKED ISSUE #617: Pod status oracle sometimes flaky
I ran into this when working on https://github.com/SREGym/SREGym/pull/616 (I will resolve this issue in that PR).

Basically, on certain faults where there are pod errors, the pod may be running for a certain period of time before crashing again. If you submit in this time period, you would be marked correct even though the failure is still present. We should monitor for a time period instead of checking just once.

HacksonClark:
Thank you @srmanda-cs for your help on this!

Closes #617

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,6 +1,16 @@
 name: Run SRE-QL Queries
 on:
   pull_request:
+    paths-ignore:
+      - '*.md'
+      - 'docs/**'
+      - 'LICENSE'
+      - '.github/workflows/claude.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   codeql-custom:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


ashmanda-cs:
> Thank you @srmanda-cs for your help on this!
> 
> Closes #617

Always happy to help!

sregym/conductor/oracles/mitigation.py
```diff
@@ -1,17 +1,55 @@
...
+# Time to wait for deployments to settle after agent submission, so we
+# evaluate a stable state rather than a transient rolling-update window.
+_ROLLOUT_SETTLE_SECONDS = 60
+_ROLLOUT_POLL_INTERVAL = 5
+
 
 class MitigationOracle(Oracle):
     importance = 1.0
 
+    def _wait_for_rollouts(self, kubectl, namespace):
```
yimingsu01: Should these checks be in the vanilla mitigation oracle definition? It seems like this logic is universal to all mitigation oracles, so it maybe should be placed in conductor, before calling `problem.mitigation_oracle.evaluate()`?

HacksonClark: @yimingsu01 Are we then abstracting the mitigation oracle to the conductor? I don't really agree with this because if you have a bad rollout you may block the benchmark from proceeding. And even with a timeout, we would want that error to be caught by the oracle so we could mark the problem as incorrect.

HacksonClark: @yimingsu01 ahh, I understand your point. My fear is that we may no longer be evaluating how the agent does temporarily (i.e. waiting for a fix to be applied before submitting). In which case maybe I should delete that section of code.

PR #632, opened 2026-03-09
Add otel + jackson's alerts

HacksonClark:
From implementing alerts for my problem set, it became clear we needed otel collector + span metrics.

Also, found an issue with ingress-misroute.

### DIFF
--- a/mcp_server/k8s/deployment.yaml
+++ b/mcp_server/k8s/deployment.yaml
@@ -20,6 +20,10 @@ spec:
       serviceAccountName: mcp-server
       nodeSelector:
         node-role.kubernetes.io/control-plane: ""
+      tolerations:
+        - key: "node-role.kubernetes.io/control-plane"
+          operator: "Exists"
+          effect: "NoSchedule"
       containers:
         - name: mcp-server
           image: sregym:latest


HacksonClark:
@Saadmrp1038 @yimingsu01 please review just for a sanity check. I haven't had any issues but it's always nice to have another set of eyes. No need to be super thorough, but please let me know if I did anything stupid.

yimingsu01 (approved):
I was testing these alerts last night too and they work great! They cover almost all problems, besides a few problems that only manifest on user requests.

HacksonClark:
> I was testing these alerts last night too and they work great! They cover almost all problems, besides a few problems that only manifest on user requests.

I believe I have removed all the client facing problems. If there are more let me know so we can get rid of them.

Saadmrp1038:
@HacksonClark it seems the smoke test is failing.

```bash
FAILED tests/integration/smoke_test.py::test_smoke_misconfig_app_hotel_res - Exception: [red]Timeout: Not all pods in namespace 'observe' reached the Ready state within 600 seconds.
```

Could be inotify issue.

HacksonClark:
> @HacksonClark it seems the integration test is failing.
> 
> 
> 
> ```bash
> 
> FAILED tests/integration/smoke_test.py::test_smoke_misconfig_app_hotel_res - Exception: [red]Timeout: Not all pods in namespace 'observe' reached the Ready state within 600 seconds.
> 
> ```
> 
> 
> 
> Could be inotify issue.

Yep. That's why I haven't merged. Going to investigate it tomorrow morning.

PR #634, opened 2026-03-09
Demo (deterministic) agent

yimingsu01:
# Synopsis
This PR adds the demo agent. The demo agent reads from the `kubectl_cmds.txt` in the same folder and run the kubectl commands line by line. Currently, this demo agent interacts with the kubectl MCP server to run kubectl on the cluster, and it can be extended to interact with other tools if needed.

# Usage
To run the demo agent, first write out the kubectl commands you would like to run, in `kubectl_cmds.txt`.

Then, run a problem you would like to test on, for example:
```bash
uv run main.py --model bedrock-claude-sonnet-4.5 --problem scale_pod_zero_social_net --agent demo --force-build
```
`--force-build` is needed for the container to contain the latest `kubectl_cmds.txt`.

Then, in another terminal session, find the container id of the agent container.
```bash
docker ps
```

Then, to step into each line of kubectl command in the text file, run:
```bash
docker exec <container_id> touch /tmp/next
```

To skip a command, run:
```bash
docker exec <container_id> touch /tmp/skip
```

To quit the agent, run:
```bash
docker exec <container_id> touch /tmp/quit
```

If you are sure that you only have one container, you can fast-track by:
```bash
docker exec $(docker ps -q) touch /tmp/next
```

# Available commands
You can put any kubectl commands in the `kubectl_cmds.txt` file.

To submit to the benchmark, write `submit()` commands in the same file. The demo agent will call the submission endpoint in the benchmark, such as `submit("the root cause is in...")`.

Generated and modified with Gemini.

### DIFF
--- a/mcp_server/Dockerfile
+++ b/mcp_server/Dockerfile
@@ -5,7 +5,8 @@ WORKDIR /app
 # Install system dependencies (curl for healthcheck, kubectl for commands, uv for Python deps)
 RUN apt-get update && apt-get install -y \
     curl \
-    && curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" \
+    && ARCH=$(dpkg --print-architecture) \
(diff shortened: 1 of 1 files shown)


HacksonClark (approved):
Looks reasonable -- approved

PR #635, opened 2026-03-09
Added unscheduleable variant of incorrect port assignment

yimingsu01:
Added a keyword argument to incorrect port assignment.
If set to `true`, then we first deploy a scheduling fault to prevent the pod from being scheduled, then deploy the actual port mis-assignment fault.

The agent must resolve the scheduling problem first, before resolving the port assignment problem.


HacksonClark (approved):
It seems as it has been thoroughly tested enough by Sasa's students -- I'll approve.

PR #637, opened 2026-03-10
Dynamic Application Deployment and Fault Injection 

lilygn:
I have created an interactive terminal UI for dynamic application deployment and fault injection.
I describe the functionality below,

### Application deployment mode:
- The applications in APP_REGISTRY are displayed to the user. When you scroll, you have an application selected and can click:
```
enter:  to deploy the application. 
p: to switch to fault injection mode
```

### Fault injection mode:

- The problems from the registry are displayed to the user. You can scroll through the problems, with the options of:
```
enter: run problem
r : recover problem
i: opens a textbox where you have the option of typing a problem_id to run the problem
```
 
The **esc** key is used to navigate back.
**TAB** can be used to switch between application deployment mode and fault injection mode.

### DIFF
--- a/sregym/conductor/problems/capacity_decrease_rpc_retry_storm.py
+++ b/sregym/conductor/problems/capacity_decrease_rpc_retry_storm.py
@@ -1,8 +1,8 @@
 from kubernetes import client
 
+from sregym.conductor.oracles.alert_oracle import AlertOracle
 from sregym.conductor.oracles.detection import DetectionOracle
 from sregym.conductor.oracles.llm_as_a_judge.llm_as_a_judge_oracle import LLMAsAJudgeOracle
-from sregym.conductor.oracles.rpc_retry_storm_mitigation import RPCRetryStormMitigationOracle
 from sregym.conductor.problems.base import Problem
 from sregym.generators.fault.inject_virtual import VirtualizationFaultInjector
 from sregym.generators.workload.blueprint_hotel_work import BHotelWrk, BHotelWrkWorkloadManager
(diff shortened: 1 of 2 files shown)


yimingsu01:
@Saadmrp1038: @lilygn made a interactive/debugging tool where we can deploy apps and problems dynamically, without overhead in the benchmark setup phase. This is meant for manual debugging of problems.

Could you:

1. cd into `visualizer/interactive_deployment`
2. run `go run .`
3. and give the tool a test?

I just tested it and I also used it during our alert audit phase and I think it works well. Just need another pair of eyes before we merge into main.

Saadmrp1038:
@lilygn Works great! One thing I ran into there's no feedback in the TUI when fault injection or recovery finishes. I end up having to check pod status manually to know if it's done, and that also doesn't work for problems that don't change pod status. Some kind of completion indicator in the UI would be nice.

yimingsu01:
> and that also doesn't work for problems that don't change pod status.

@Saadmrp1038 I remember I ran `incorrect_port_assignment` and the checkout service did get misconfigured. When you say "doesn't work", do you mean that the fault is not injected at all?

yimingsu01:
I think the fault injection logic is done by directly calling `inject_fault` in the corresponding problem definition

Saadmrp1038:
@yimingsu01 Sorry, my bad. I phrased it poorly. All the functionalities (deploy/inject/recover) are working correctly. The issue is that when injecting or recovering a fault, there's no indication when the injection/recovery process has actually completed. So, for faults that don’t change any pod status, I have no way to tell when the operation has finished.

yimingsu01:
@Saadmrp1038 Gotcha. No worries. I asked Claude to update it. Now at the bottom of the screen, when the action is done, it should have a `[DONE]` prefix to the message. Could you please test again?

Saadmrp1038:
Tested again. Looks great now!

PR #651, opened 2026-03-13
Replace LLMJudge with RCAJudge for Deterministic RCA Evaluation

EvanTian233:
## Summary
Replaces the freeform `LLMJudge` with a structured RCA evaluator (`RCAJudge`) that scores agent root cause analysis across 3 dimensions using a single LLM call with 9 Yes/No questions. Provides deterministic, interpretable scoring with per-dimension breakdowns.

### Core Implementation
- **`judge.py`**: Added `RCAJudge` class with dynamic YAML-driven evaluation logic. Single LLM call at temperature=0 evaluates all questions. Retry logic handles invalid JSON/missing questions. Returns `JudgmentReport` with composite score (0.0-1.0) and per-dimension breakdowns. Drop-in compatible with `LLMJudge.judge()`.

- **`rca_checklists.yaml`**: Version 2.0 configuration defining 3 evaluation dimensions (Fault Localization, Fault Characterization, Scope Precision) with 3 questions each (9 total). Dimensions derived from analysis of 70 ground-truth `root_cause` strings. Threshold: 0.70, equal weights (0.33/0.33/0.34).

- **`models.py`**: Dataclasses for structured evaluation results: `QuestionResult` (per-question answer/evidence/confidence), `DimensionResult` (per-dimension score), `JudgmentReport` (full report with legacy-compatible verdict/reasoning).

### Integration
- **`llm_as_a_judge_oracle.py`**: Updated to use `RCAJudge.judge_detailed()`. Accuracy now uses composite score scaled to 0-100 (was binary 0/100). Results include dimension breakdown and composite score.

### Testing
- **`test_rca_judge.py`**: Comprehensive test suite with 4 scenarios (all-yes, all-no, drop-in compatibility, empty solution). Prioritizes OpenAI backend (`gpt-4.1-mini`) if `OPENAI_API_KEY` is set, otherwise uses mocks. Includes sys.path fix for `llm_backend` imports. Prints detailed reports for all tests.

### Optional Utilities
- **`extract_root_causes.py`**: AST-based parser to extract ground-truth `root_cause` strings from problem files → `root_causes.csv`. Used for dimension design validation. Not required for runtime.
- **`root_causes.csv`**: 70 ground-truth entries (file, class, root_cause). Reference data for testing/validation only.

## Testing
Set `OPENAI_API_KEY` environment variable and run:
```bash
python sregym/conductor/oracles/llm_as_a_judge/extract_root_causes.py
python sregym/conductor/oracles/llm_as_a_judge/test_rca_judge.py
```
Expected outputs:
image

## Migration Notes
- `LLMJudge` remains in `judge.py` for backward compatibility but is no longer used.
- Oracle accuracy is now granular (0-100) instead of binary (0 or 100).
- Results dict now includes `composite_score` and `dimensions` fields.

### DIFF
--- a/sregym/conductor/problems/loadgenerator_flood_homepage.py
+++ b/sregym/conductor/problems/loadgenerator_flood_homepage.py
@@ -23,9 +23,9 @@ class LoadGeneratorFloodHomepage(Problem):
             component=self.faulty_service,
             namespace=self.namespace,
             description=(
-                "The load generator is sending a sustained burst of homepage requests against the frontend "
-                "deployment, saturating frontend capacity. This leads to queueing, high latency, and timeout "
-                "spikes during normal user flows. Users observe intermittent homepage errors and degraded "
+                f"The `{self.faulty_service}` deployment is experiencing a sustained traffic surge on the "
+                "homepage endpoint, saturating frontend capacity. This leads to queueing, high latency, and "
+                "timeout spikes during normal user flows. Users observe intermittent homepage errors and degraded "
                 "responsiveness across storefront interactions."
             ),
         )


tianyin:
This is a cool PR @EvanTian233 ! 

I was debating with students what should be the right way to evaluate RCA results and today I found this PR online. I agree that a fine-grained questionnaire should be able to improve the grading/judging results by guiding LLMs to think more and think on the right questions.

I personally think modeling RCA itself is a rather interesting and deep _research_ problem. The last great work I saw is Yongle's work on inflection point hypothesis, but it may not generalize to distributed system failures with many correlated events.  
 
One effort to have a more precise model is the NTAM metric (see Section C.6.3 of [the ITBench paper](https://arxiv.org/pdf/2502.05352)). The basic idea is to grade RCA based on how "close" the identified root causes to the real failing service. In practice, NTAM requires tedious annotations of the service dependency graph (which can be very brittle in practice) and can perhaps only work at the service level (@HacksonClark once argued that the distance function was hard to define; I didn't understand why but I guess it refers to root causes defined at lower level). Interestingly, what I heard is that ITBench itself chose to use LLM-as-a-judge instead of NTAM.

EvanTian233:
> This is a cool PR @EvanTian233 !
> 
> I was debating with students what should be the right way to evaluate RCA results and today I found this PR online. I agree that a fine-grained questionnaire should be able to improve the grading/judging results by guiding LLMs to think more and think on the right questions.
> 
> I personally think modeling RCA itself is a rather interesting and deep _research_ problem. The last great work I saw is Yongle's work on inflection point hypothesis, but it may not generalize to distributed system failures with many correlated events.
> 
> One effort to have a more precise model is the NTAM metric (see Section C.6.3 of [the ITBench paper](https://arxiv.org/pdf/2502.05352)). The basic idea is to grade RCA based on how "close" the identified root causes to the real failing service. In practice, NTAM requires tedious annotations of the service dependency graph (which can be very brittle in practice) and can perhaps only work at the service level (@HacksonClark once argued that the distance function was hard to define; I didn't understand why but I guess it refers to root causes defined at lower level). Interestingly, what I heard is that ITBench itself chose to use LLM-as-a-judge instead of NTAM.

Thank you Prof. Xu @tianyin for the encouragement and sharing your thoughts! :)

I totally agree that modeling RCA is a fascinating _research_ problem, and I just read the NTAM metric section from ITBench. Here's my chain-of-thought behind my design:
-> The problem formulation: SREGym is designed for AI Agent, and the RCAJudge's goal is to evaluate these Agents' performance. No matter how fancy the agentic design is, the expect output of RCA is a string, we use the one-liner `root_cause` definition for each problem as the ground truth, and we use an LLM to evaluate. (As per discussed with @HacksonClark) This makes RCA evaluation in the SREGym context a LLM-based natural language generation (NLG) evaluation problem. (Is it? Please argue if you have different opinion.)
-> From section 5 of this survey [1] we know that Human–LLM collaborative NLG evaluation seeks to leverage the strengths of both, and there are some recent developments in checklist-based techniques [2] [3]. In their research they've proven this LLM-as-a-Judge score is close to human judgement.
-> Now the differences when comparing RCA to traditional NLG is: the ground truth is a known entity (e.g service name), the 'analysis' contains reasoning process, the output in real-world should be consumed by SREs. To cover all these considerations, I've implemented this SURE-score [4] (Section 4.2) as an initial version of checklist-based Human-LLM judge in my prior work. 
-> Which motivates this PR to come up with a checklist-based Human-LLM judge that fits better for the SREGym RCA evaluation problem formulation.

If you like this direction we are going, we could collaborate further on the human aspect of this judge and and improve the dimensions/checklists. Or introduce some more deterministic dimensions (on the direction of NTAM) to make the composite score more robust. How does that sound?

[1] Gao, Mingqi, et al. "Llm-based nlg evaluation: Current status and challenges." Computational Linguistics 51.2 (2025): 661-687. (https://aclanthology.org/2025.cl-2.9.pdf)
[2] Lee, Yukyung, et al. "Checkeval: A reliable llm-as-a-judge framework for evaluating text generation using checklists." Proceedings of the 2025 Conference on Empirical Methods in Natural Language Processing. 2025. (https://aclanthology.org/2025.emnlp-main.796/)
[3] Chu, Seong Yeub, Jong Woo Kim, and Mun Yong Yi. "Think together and work better: Combining humans' and LLMs' think-aloud outcomes for effective text evaluation." Proceedings of the 2025 CHI Conference on Human Factors in Computing Systems. 2025. (https://arxiv.org/abs/2409.07355)
[4] Tian, Yifang, et al. "GALA: Can Graph-Augmented Large Language Model Agentic Workflows Elevate Root Cause Analysis?." arXiv preprint arXiv:2508.12472 (2025). (https://arxiv.org/pdf/2508.12472)

tianyin:
WoW thanks for the discussion and your work is deeper than I thought. 

I will read your paper. // I don't know much about NLG and checklist-based approach. 

> SURE-score 

Very cool name

>  human aspect 

Happy to chat more.

I'd love to see some results (e.g., a few problems) on how RCAJudge improves the grading over LLMJudge.

HacksonClark:
I agree with Tianyin, would like to see some runs.

Also, should we update the root cause labels to have some criteria specific to the rubric? Could that help our performance?

sregym/conductor/oracles/llm_as_a_judge/test_rca_judge.py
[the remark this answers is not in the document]
EvanTian233: Yes! It is just a testing file. Sounds good

EvanTian233:
Thanks @tianyin and @HacksonClark  for the discussion! Let's do some evaluation first then let's discuss the results.

@HacksonClark Would you have a few agent outputs already or could you help with generating some? So that I don't have to run the full pipeline this would make things easier, also in this way we could have a consistent evaluation set for improving the RCAJudge. 

About root cause labels - We are on the same page. I think more tailored and elaborated root cause label should help. We can discuss once we see the initial comparison.

HacksonClark:
@EvanTian233 Yes, I have some trajectories from recent runs you can examine. I shared a folder of a bunch of trajectories from our latest runs.

EvanTian233:
Hi @HacksonClark, I have addressed all your comments related to code quality above. Please check.

yimingsu01:
@EvanTian233 This is a great PR! I have two questions:
1. It seems that the judge rubric highly relies on the ground-truth labels we have in the codebase right now. As you can see in some problems, the ground-truth label is quite high-level. Can the judge itself access our fault injection logics and compare the the agent's answer to that? Would this be a better solution?
2. Can we also add some basic components like in-context learning and CoT in the judge's prompt? It would both be helpful to us for debugging purposes and helpful to the judge making better judgment.

yimingsu01:
Moreover, it would be much more sound and complete if we can ground the judge on the agent's traces. I.e., the judge should check whether the agent's answer is correctly supported by the agent's tool calls in the traces.

EvanTian233:
> @EvanTian233 This is a great PR! I have two questions:
> 
> 1. It seems that the judge rubric highly relies on the ground-truth labels we have in the codebase right now. As you can see in some problems, the ground-truth label is quite high-level. Can the judge itself access our fault injection logics and compare the the agent's answer to that? Would this be a better solution?
> 2. Can we also add some basic components like in-context learning and CoT in the judge's prompt? It would both be helpful to us for debugging purposes and helpful to the judge making better judgment.

Hi @yimingsu01 I like these two ideas! For 1 - After discussing with Tianyin and Jackson, we agreed that improving the ground-truth label is the next step. What you said is a great idea to enrich the context for ground truth. I will give it a try to prepare a more structured ground truth label based on the injection logic. For 2 - Sounds good! That makes sense too.

EvanTian233:
> Moreover, it would be much more sound and complete if we can ground the judge on the agent's traces. I.e., the judge should check whether the agent's answer is correctly supported by the agent's tool calls in the traces.

I 100% agree. In my previous research, I had a logical coherence dimension that somewhat does that. There are rubrics that's specific for evaluating agentic workflow's reasoning trajectory. Not just the answer - but how you get there should be logically coherent.  It would consume more tokens tho since the agent trace could be large and vary across different agents. Do we want to beef it up?

HacksonClark:
@EvanTian233 

Let's beef it up as much as possible!

yimingsu01:
Let me run a sanity-checking run before merging to make sure it doesn't break trunk.

yimingsu01:
@EvanTian233 Please see my uploaded results from two problems:

[0323_1603_stratus_ALL_results.csv](https://github.com/user-attachments/files/26195555/0323_1603_stratus_ALL_results.csv)

Where can I see the reasoning for the judge's verdict? In this case, it seems like the trajectory doesn't support the judge's verdict for the first problem.

EvanTian233:
@yimingsu01 I just pushed a commit. The reasoning of the judge is really the per checklist question answer, evidence and confidence. With the new change you should see the judge reasoning both in the reasoning field and a separate column named `checklist` for easier parsing. Please check and let me know how it goes.

sregym/conductor/oracles/llm_as_a_judge/judge.py
```diff
@@ -375,25 +494,88 @@ def judge_detailed(self, solution: str, expectation: str):
...
+        lines.append("### Natural-Language Root Cause")
+        lines.append(expectation if expectation else "(No fault - system is operating normally)")
+        lines.append("")
+
+        # ---- Section 2: Agent diagnosis ----
+        lines.append("## Agent Diagnosis")
+        lines.append(solution)
+        lines.append("")
+
+        # ---- Section 3: Agent trace (for D4) ----
```
EvanTian233: @yimingsu01 My guess is the agent trace is not loaded properly. It requires a string input but the actual format maybe a JSON or th? Would be great if you can help me print the log or show me an example of the agent trace then I can fix it accordingly.

sregym/conductor/oracles/llm_as_a_judge/DIAGNOSIS_JUDGE_V3_CHANGELOG.md
HacksonClark: Make sure to delete this file before merging. I don't mind these summaries but let's keep them in the PR conversation instead of in Markdown files.

sregym/conductor/oracles/llm_as_a_judge/judge.py
yimingsu01: This is a good question, and as of now I'm not sure how can the judge access the traces from the agent. At the end of a benchmark run, the traces are saved in `.jsonl` files, under `<repo_root>/results` directory with the latest `main` branch.

yimingsu01: One easy fix is that, we immediately save the agent traces as soon as they finish, then the LLM judge can access the `.jsonl` file.

EvanTian233: @HacksonClark @yimingsu01 this is done. Please check.

sregym/conductor/oracles/llm_as_a_judge/judge.py
```diff
@@ -8,12 +25,60 @@
...
 
+from llm_backend.get_llm_backend import LiteLLMBackend
 from llm_backend.init_backend import get_llm_backend_for_judge
 
+# ---------------------------------------------------------------------------
+# Backend helpers
+# ---------------------------------------------------------------------------
+
+
+def _create_judge_backend(
```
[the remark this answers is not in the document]
yimingsu01: @EvanTian233 Got it, thanks!

EvanTian233:
@HacksonClark @yimingsu01 Please try again? I've tuned everything to put more emphasis on the injection logic and completely removed the categories. Also the LLM backend has been updated to match its original usage.

EvanTian233:
@HacksonClark @yimingsu01 as per our discussion, I'm removing the need for all the weird mapping and deleting the fault spec file. I'm enriching the root_cause as much as possible.

I've updated 2 problems for pilot testing: `ad_service_high_cpu` and `ad_service_manual_gc`, could you please give it a try with these two cases and let me know your thoughts? I tried to avoid inferring anything but just keep it injection fact-based. If this pattern is good I'll go ahead and generalize it to all the problems.

HacksonClark:
@EvanTian233 That's quite good! You can be even more specific too and describe the full symptom.

EvanTian233:
> @EvanTian233 That's quite good! You can be even more specific too and describe the full symptom.

Sounds good! I just enriched those two test cases in my latest commit. I'll proceed with updated all the problems following the same pattern. Finger-crossed.

EvanTian233:
> > @EvanTian233 That's quite good! You can be even more specific too and describe the full symptom.
> 
> Sounds good! I just enriched those two test cases in my latest commit. I'll proceed with updated all the problems following the same pattern. Finger-crossed.

@HacksonClark @yimingsu01 Done. All following the same structure and elaborated symptom. Please check the latest code and let me know how it goes.

yimingsu01:
@EvanTian233 They all look reasonable. We can probably only be so accurate with natural language. I agree that including at least the faulty components is a net-positive step.

yimingsu01:
@HacksonClark I think this PR looks good to me and we can carefully start eval'ing now.

HacksonClark:
This is looking good overall, however I have some points when it comes to writing the `root_cause` labels in the problem classes. They should describe the *symptom* of the failure rather than the mechanism.

As an example:
```
description=(
                "dm-flakey corruption is injected on the underlying storage path used by this MongoDB workload, "
                "so reads and/or writes can be silently corrupted without immediate hard I/O errors, leading to latent "
                "data integrity issues. Symptoms include inconsistent query results, mismatched records, or "
                "intermittent application anomalies that do not map cleanly to obvious storage failures."
            ),
```

I think that's too specific, we shouldn't talk about the mechanism (dm-flakey) but the failure we're simulating (i.e. describing a latent sector error).

I think we will track this in a separate issue.

HacksonClark:
I will merge this later, may make some updates from going through the labels.

EvanTian233:
> I will merge this later, may make some updates from going through the labels.

Ok perfect! Sounds great!

HacksonClark (approved):
Okay, I'm good with the labels now.

PR #659, opened 2026-03-17
Taints all node so the pods actually stuck at Pending

yimingsu01:
Fixes problem `taint_no_toleration`, make sure the target pods stuck at pending by tainting all nodes. Previously it can still be scheduled on the control-plane node so no alerts fire.

After this fix alerts fire correctly for pending pods.


HacksonClark:
Does this undo my previous fix by chance?

yimingsu01:
This PR: https://github.com/SREGym/SREGym/pull/198/changes seems to only inject into worker nodes, There doesn't seem to be affinity requirements for the `user-service` pod to be scheduled only on the worker nodes, which is why the fault is not manifesting in my case.

yimingsu01:
The changes I propose prevent that from happening completely by tainting all nodes

PR #667, opened 2026-03-23
Remove dependence on inventory file and paramiko

### LINKED ISSUE #627: Refactor `inject_remote_os` to not use inventory
This was originally implemented using the `inventory.yml` file (which is very stupid) instead of just getting the nodes with `kubectl get nodes`. Inventory is way too brittle because it doesn't accommodate any other setup other than CloudLab.

HacksonClark:
Closes #627


Saadmrp1038:
Seems like these changes are already in main via the PR #668

PR #668, opened 2026-03-23
Wait for alerts before starting agent

### LINKED ISSUE #658: wait for alerts to fire
We have audited all alerts, but it does take some time for alerts to fire after injecting the fault. We should play around with the timing of the alerts and should only start the agent after the alert is firing.

HacksonClark:
Closes #658

### DIFF
--- a/sregym/conductor/conductor.py
+++ b/sregym/conductor/conductor.py
@@ -9,6 +9,7 @@ from pathlib import Path
 import yaml
 
 from sregym.conductor.constants import StartProblemResult
+from sregym.conductor.oracles.alert_oracle import AlertOracle
 from sregym.conductor.oracles.detection import DetectionOracle
 from sregym.conductor.oracles.diagnosis_oracle import DiagnosisOracle
 from sregym.conductor.problems.registry import ProblemRegistry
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
@HacksonClark what if alerts were already firing from the app deployment or fault injection process itself and not the actual fault. For example, when running `missing_service_astrnomoy_shop` fault during the fault injection process some alerts fires. They still persist for a while after fault injection is done.

```bash
  [astronomy-shop] ServiceEndpointDown [firing]
    Detail:    Service endpoint ad.astronomy-shop:8080 in astronomy-shop is unreachable.

  [astronomy-shop] KubePodNotReady [firing]
    Pod:       grafana-bdbfb995f-5knj7
    Detail:    Pod grafana-bdbfb995f-5knj7 in namespace astronomy-shop has been not ready for more than 1 minute.
```

Like here the first one is the alert associated with the fault (we removed the `ad` service). The second one goes away after 30-40s.

yimingsu01:
@Saadmrp1038 The change then might be to (1) wait until first alert fires, then (2) wait 2m (the longest alert duration we have) for other transient alerts to quiesce, then kick start the agent

Saadmrp1038:
@yimingsu01 sounds about right.

PR #669, opened 2026-03-23
Stop alert trigger for noises

### LINKED ISSUE #655: Make sure noises don't cause alerts that mess with the alert oracle.
We will need to do some testing, I think the noises are quick enough that it shouldn't cause the alert to fire (this is why alerts are on a timer in the first place).

### LINKED ISSUE #665: `helm upgrade` fails with "unknown flag: --server-side" on Helm 3
When running `auth_miss_mongodb` got the following error.

```bash
WARNING - all.infra.helm - Helm upgrade stderr output:
Error: unknown flag: --server-side
ERROR - all.infra.helm - Error during helm upgrade:
ERROR - __main__ - Driver thread crashed
```

`helm.py` uses `--server-side=true` flag in the `upgrade` method. This flag doesn't exist in `helm 3` and is a `helm 4` feature for Server-Side Apply (ref: https://alexandre-vazquez.com/helm-40/).  I was in helm version `3.19` which is why I encountered this error. Worked fine after upgrading to `4.1.3`. I believe `--force-conflicts` is also a `helm 4` specific flag.

These flags were introduced in the PR "Cluster cleanup heuristic #613". Since `helm 4` was released very recently (November 12th, 2025) we should either make the code compatible with `helm 3` or mention in our readme that `helm 4` is required.

Saadmrp1038:
Closes #655
Closes #665 (Just a small readme update)

Tested all the noises. Only `pod-failure` triggers alerts in the current setup, depending on which pod it targets. For example, `search` in `hotel-reservation`, `otel-collector` in `astronomy-shop`, and `compose-post-service` in `social-network` trigger alerts. Reduced the duration for this noise to 45s so it won't trigger alerts.

From all my testing I never saw the other noises trigger alerts. Although since noise targeting is random, I'm not sure if there's a unique pod for some app that could trigger alerts for a noise other than `pod-failure`. If we don't want to take that chance we'd have to reduce the duration for all noises.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ SREGym is inspired by our prior work on AIOpsLab and ITBench. It is architecture
 ### Requirements
 - Python >= 3.12
 - [Docker](https://docs.docker.com/get-docker/)
-- [Helm](https://helm.sh/)
+- [Helm](https://helm.sh/) >= 4.0
 - [brew](https://docs.brew.sh/Homebrew-and-Python)
 - [kubectl](https://kubernetes.io/docs/tasks/tools/)
 - [uv](https://github.com/astral-sh/uv)


Saadmrp1038:
@HacksonClark if you had something different in mind lemme know.

HacksonClark:
@Saadmrp1038 Is this ready for review? It's still a draft PR.

Saadmrp1038:
@HacksonClark yeah ig. I kept it in draft in case you had something totally different in mind.

HacksonClark:
@Saadmrp1038 Gotcha. If you want me to take a look just mark ready for review. Worst case we close it and move on. I don't check things that are drafts so I know what to review, so just always mark ready for review if you want me to take a look.

PR #696, opened 2026-04-01
Add sanity check for model string using LiteLLM

### LINKED ISSUE #695: Add sanity check for model string
Since we removed `configs.yml`, we should add a sanity check to check that the model string is actually valid. Otherwise, even with an incorrect model string, the app deploys successfully and only throws errors after fault injection. The same issue occurs when the user forgets to set the LLM API key in the environment. The error only appears after app deployment and fault injection (I run into this quite often 🙃).
This check can be done using LiteLLM’s `get_valid_models()` (https://docs.litellm.ai/docs/set_keys) helper function. This helper reads the environment and returns a list of supported LLMs for those env variables

HacksonClark:
Adds `validate_model_config()` to `main.py` that calls `litellm.utils.get_valid_models()` before any Kubernetes resources are deployed. If a model string is not in the valid models list, the function attempts to identify the provider and reports whether the issue is an unrecognised model string or a missing API key, then exits early with a clear error message.

Closes #695

### DIFF
--- a/clients/resolve/driver.py
+++ b/clients/resolve/driver.py
@@ -1,13 +1,12 @@
 """Resolve AI agent driver for SREGym.
 
-Sets up the Resolve satellite and ktunnel, fires a webhook alert to
-trigger Resolve AI, then waits for it to complete via MCP tools.
+Sets up the Resolve satellite, fires a webhook alert to trigger
+Resolve AI, then waits for it to complete via MCP tools.
 """
 
 import json
 import logging
 import os
-import shutil
 import signal
 import sys
 import time
@@ -107,26 +106,19 @@ def main():
     logger.info("Resolve AI agent driver starting")
 
     # The agent launcher sets KUBECONFIG to a filtering proxy, but the
-    # Resolve driver manages cluster infrastructure (ktunnel, helm, kubectl
-    # patches) that must talk to the real K8s API server.
+    # Resolve driver manages cluster infrastructure (helm, kubectl patches)
+    # that must talk to the real K8s API server.
     os.environ.pop("KUBECONFIG", None)
 
-    # Check that ktunnel is installed
-    if not shutil.which("ktunnel"):
-        logger.error("ktunnel is not installed. Install it from https://github.com/omrikiei/ktunnel")
-        sys.exit(1)
-
-    # Register cleanup handler so ktunnel + satellite are torn down
+    # Register cleanup handler so satellite is torn down
     # when main.py terminates this process
     signal.signal(signal.SIGTERM, _shutdown_handler)
     signal.signal(signal.SIGINT, _shutdown_handler)
 
-    # Get app info from conductor BEFORE starting ktunnel, because ktunnel
-    # binds to local port 8000 and intercepts connections to the conductor.
     app_info = get_app_info()
     logger.info(f"App info: {app_info}")
 
-    # Set up ktunnel and install Resolve satellite
+    # Set up submit proxy and install Resolve satellite
     resolve_setup.start()
 
     # Fire the alert
--- a/clients/resolve/setup.py
+++ b/clients/resolve/setup.py
@@ -1,12 +1,13 @@
 """Resolve AI infrastructure setup and teardown.
 
-Manages the ktunnel reverse tunnel, Resolve satellite Helm chart,
-and MCP server submit proxy configuration.
+Manages the Resolve satellite Helm chart and MCP server submit proxy
+configuration. The conductor API is reached directly via the host IP
+(no ktunnel tunnel required).
 """
 
 import logging
+import socket
 import subprocess
-import time
 
 from sregym.service.helm import Helm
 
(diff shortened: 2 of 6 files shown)


Saadmrp1038:
@HacksonClark can you please check if this works with bedrock model strings?

HacksonClark:
Sure thing!

HacksonClark:
@Saadmrp1038 Had to handle bedrock separately since it uses credentials instead of a key but it seems to be working.

Saadmrp1038:
I was thinking about doing a simple API call. Spin up a temporary container and make a simple "say ok" call. This would catch issues with all drivers (wrong model, wrong environment, etc.). The current one would only work for `stratus`. The one downside is that for `claudecode/codex` it takes a bit of time for the check to finish (as they need to be installed in the container first). We can skip this check for `claudecode/codex` if the time is too much. But for `stratus` it seems more robust (as we are making an actual call to validate the model string and env).

```bash
$ time uv run main.py --problem auth_miss_mongodb --agent stratus --model openai/gpt-5.5-mini
INFO - __main__ - 🔧 Config — agent: stratus, agent_model: openai/gpt-5.5-mini, judge_model: openai/gpt-5.5-mini
INFO - __main__ - 🔍 Running pre-flight check for 'stratus'...
preflight failed: litellm.NotFoundError: OpenAIException - The model `gpt-5.5-mini` does not exist or you do not have access to it.
ERROR - __main__ - ❌ Pre-flight check failed for 'stratus'

real    0m13.303s
user    0m6.329s
sys     0m1.359s

$ time uv run main.py --problem auth_miss_mongodb --agent claudecode --model openai/gpt-5.5-mini
INFO - __main__ - 🔧 Config — agent: claudecode, agent_model: openai/gpt-5.5-mini, judge_model: openai/gpt-5.5-mini
INFO - __main__ - 🔍 Running pre-flight check for 'claudecode'...
There is an issue with the selected model (gpt-5.5-mini). It may not exist or you may not have access to it. Run --model to pick a different model.
ERROR - __main__ - ❌ Pre-flight check failed for 'claudecode'

real    0m29.008s
user    0m4.037s
sys     0m0.374s

$ time uv run main.py --problem auth_miss_mongodb --agent codex --model openai/gpt-5.5-mini
INFO - __main__ - 🔧 Config — agent: codex, agent_model: openai/gpt-5.5-mini, judge_model: openai/gpt-5.5-mini
INFO - __main__ - 🔍 Running pre-flight check for 'codex'...
Reading additional input from stdin...
OpenAI Codex v0.118.0 (research preview)
--------
workdir: /logs
model: gpt-5.5-mini
provider: openai
approval: never
sandbox: danger-full-access
reasoning effort: none
reasoning summaries: none
session id: 019d495c-c85d-78a2-ae44-6eaa08965d67
--------
user
say ok
warning: Codex could not find system bubblewrap on PATH. Please install bubblewrap with your package manager. Codex will use the vendored bubblewrap in the meantime.
ERROR: Reconnecting... 2/5
ERROR: Reconnecting... 3/5
ERROR: Reconnecting... 4/5
ERROR: Reconnecting... 5/5
ERROR: {
  "error": {
    "message": "The requested model 'gpt-5.5-mini' does not exist.",
    "type": "invalid_request_error",
    "param": "model",
    "code": "model_not_found"
  }
}
ERROR: {
  "error": {
    "message": "The requested model 'gpt-5.5-mini' does not exist.",
    "type": "invalid_request_error",
    "param": "model",
    "code": "model_not_found"
  }
}
ERROR - __main__ - ❌ Pre-flight check failed for 'codex'

real    1m6.375s
user    0m4.068s
sys     0m0.366s
```

HacksonClark:
@Saadmrp1038 for pre-flight check could we just do a litellm check to cover any model? If LiteLLM fails then we throw an error.

Saadmrp1038:
That could work actually. I just need to cross check if the model string for litellm and claudecode/codex are compatible or not.

HacksonClark:
@Saadmrp1038 Not sure about codex, but for claude code it will yell at you that you don't have an API key set or an OAUTH token set. We could just validate that. I bet codex has something similar.

Saadmrp1038:
LiteLLM can be used to check all model strings. But the caveat is if you are using claudecode agent via the oauth token the litellm check will fail as it cannot use that. Same for codex oauth token.

HacksonClark:
@Saadmrp1038 Yep, makes sense. We will have to check claude code and codex separately.

HacksonClark:
@Saadmrp1038 I have a fresh cluster I was going to use for codex (and claude code) so I can test this PR there. in 1-2 hours if that would be helpful.

Saadmrp1038:
@HacksonClark yes, that would be very nice!
Also, it takes way less time for the pre-flight check when I run it inside a CloudLab node. My net was just shit.

*time taken when run inside CloudLab node*
```bash
$ time uv run main.py --problem service_port_conflict_hotel_reservation --agent stratus --model openai/gpt-5.5-mini

preflight failed: litellm.NotFoundError: OpenAIException - The model `gpt-5.5-mini` does not exist or you do not have access to it.
ERROR - __main__ - ❌ Pre-flight check failed for 'stratus'

real    0m16.084s

$ time uv run main.py --problem service_port_conflict_hotel_reservation --agent claudecode --model openai/gpt-5.5-mini

There is an issue with the selected model (gpt-5.5-mini). It may not exist or you may not have access to it. Run --model to pick a different model.
ERROR - __main__ - ❌ Pre-flight check failed for 'claudecode'

real    0m13.665s

$ time uv run main.py --problem service_port_conflict_hotel_reservation --agent codex --model openai/gpt-5.5-mini

ERROR: {
  "error": {
    "message": "The requested model 'gpt-5.5-mini' does not exist.",
    "type": "invalid_request_error",
    "param": "model",
    "code": "model_not_found"
  }
}
ERROR - __main__ - ❌ Pre-flight check failed for 'codex'

real    0m22.635s
```

HacksonClark:
@Saadmrp1038 This worked for me. Feel free to approve on you end if it's working for you. I'm ready to merge it when checks clear.

PR #700, opened 2026-04-01
Fix serviceaccount permissions issue in kubectl MCP server

### LINKED ISSUE #698: Service account blocked mitigation actions
There is a bug related to a service account in the MCP tools for `kubectl`. It should have full permissions so the agent can try whatever action it wants to take:
```
Root Cause
The frontend deployment was modified with an incompatible configuration that combines:

    Local-hostpath storage (requires all pods on same node)
    Pod anti-affinity rules (requires pods on different nodes)
    Replicas > 1

This is a configuration error - these three settings are mutually exclusive and create an impossible scheduling constraint.
Remediation
Diagnosis submitted to the benchmark system.
Recommended fix: Remove the pod anti-affinity rules from the frontend deployment to allow both pods to run on node5 where the local PV is bound.
Note: Remediation could not be executed due to RBAC permission limitations - the kubectl service account lacks permissions to modify deployments in the astronomy-shop namespace.
```
That's from an example run.

HacksonClark:
Closes #698 

Thanks Resolve for helping find this bug :)

### DIFF
--- a/clients/resolve/resolve.md
+++ b/clients/resolve/resolve.md
@@ -1,31 +1,84 @@
-Follow these steps to diagnose and fix the incident:
+## Time Constraint
 
-1. Investigate the root cause. Get all the pods and deployments in the affected namespace to understand what services are running and which are unhealthy. Use kubectl, prometheus, loki, and jaeger to gather information. Go as deep as you can into what is causing the issue.
+You have a **strict 30-minute time limit** for the entire workflow. If you do not submit within this window, you receive zero credit. Budget your time:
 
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
> Thanks Resolve for helping find this bug :)

Does the resolve agent generate any trajectory/report? I am interested to see how they work.

HacksonClark:
@Saadmrp1038 You can see it on their product internally, sadly folks who are not on our resolve account (which is just @yimingsu01 and I for now) cannot see it.

PR #705, opened 2026-04-02
Break out of mitigation retry loop when the agent decides to submit

yimingsu01:
@Saadmrp1038 could you take a look if this fixes it? Theoretically it should, and I couldn't reproduce the behavior you saw.

### DIFF
--- a/clients/stratus/stratus_agent/driver/driver.py
+++ b/clients/stratus/stratus_agent/driver/driver.py
@@ -209,6 +209,10 @@ def validate_oracles(oracles: list[BaseOracle]) -> list[bool | list[OracleResult
     return [True, results]
 
 
+def mitigation_submission_requested(last_state) -> bool:
+    return bool(getattr(last_state, "values", {}).get("submitted", False))
+
+
 def get_app_info():
     ltc = LanggraphToolConfig()
     url = ltc.benchmark_app_info_url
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
@yimingsu01 it does but I have concerns that this is not addressing the root cause. Shouldn't the mitigation agent only have access to the `f_submit_tool` and not the `submit_tool` in validate mode? The `stratus` code already supports the scenario that when the oracle results pass it manually calls submit. In current setup it seems that the `f_submit_tool` is not wired to the mitigation agent. It was changed in a [commit](https://github.com/SREGym/SREGym/pull/603/changes/39651be619c738aa0063b895fa0194a0503e3a0e#diff-c16fcbc61ef853cd50cec35b09cb1b7a042e7e7fd0abc486df96e4cd5eca226f) 4 weeks age.

*clients/stratus/stratus_agent/driver/driver.py (L640-L656)*
```bash
 # getting oracle result
            try:
                oracle_results = validate_oracles(oracles)
                oracle_results_lst.append(str(oracle_results))
                has_succeeded = oracle_results[0]
            except Exception as e:
                logger.error(f"Oracle validation failed with error: {e}", exc_info=True)
                oracle_results = [False, []]
                oracle_results_lst.append(f"Oracle error: {str(e)}")
                has_succeeded = False
            if has_succeeded:
                # agent succeeds, let's finish here.
                logger.info("agent succeeds! manually submitting for the agent")
                await manual_submit_tool("")
                logger.info("breaking the retry loop")
                break
```

Saadmrp1038:
> I feel we should do this regardless. It just makes sense to break in this case.

I agree

PR #708, opened 2026-04-02
Store diagnosis agent's submission in result CSV

### LINKED ISSUE #707: Store diagnosis agent's submission in result csv
Since we’ve moved to the new diagnosis eval without seeing the diagnosis agent’s original submission, it’s quite hard to check whether the judge’s judgment is correct (or at least whether I, as a human, would agree with it).
I think we should save the diagnosis agent’s submission results to the results CSV for easier comparison.

HacksonClark:
Store the agent's submitted diagnosis text in conductor results so it is automatically written as `Diagnosis.submission` in the results CSV for any agent. The submission is the direct input passed to the LLM judge, making it easy to audit judge decisions.

Closes #707

### DIFF
--- a/sregym/conductor/conductor.py
+++ b/sregym/conductor/conductor.py
@@ -293,6 +293,7 @@ class Conductor:
 
         self.logger.info("Start Eval for Diagnosis", extra={"sol": solution})
         r = problem.diagnosis_oracle.evaluate(solution)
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
Tested with stratus and claudecode. Works great!

PR #713, opened 2026-04-02
update ground truth labels

### LINKED ISSUE #709: List of lacking ground truths
List of ground truths that I believe are currently lacking and could result in wrong judge evaluations:

| Problem | Issue with ground truth label |
|---------|---------------|
| `service_port_conflict` | Does not mention `prometheus-node-exporter` DaemonSet (port 9100 occupant) with which the `{service}` (e.g., `frontend`) has port conflict with|
| `stale_coredns_config`| component `coredns` is correct but it feels like namespace should be `kube-system` not app namespace as the fault is in coredns config|
|`target_port` | Missing actual port numbers (`targetPort: 9999`) |
|`wrong_operator_image` | `deployment/tidb-operator-controller-manager` is from `tidb-operator` namespace not `tidb-cluster`|

Saadmrp1038:
Closes #709


Saadmrp1038:
@HacksonClark yeah definitely. But is a good way to stumble into a bunch of bugs :)

PR #721, opened 2026-04-06
Fix silent data corruption stopping progress by affecting k8s api server

HacksonClark:
`silent_data_corruption` had an issue where the k8s api server was getting affected causing the benchmark to not be able to make progress. From testing it seems to be working so far, will merge when complete.


HacksonClark:
This worked -- resolve and stratus ran end to end with these fixes! Going to merge when smoke test clears.

PR #725, opened 2026-04-09
Fix `flagd` fault propagation delay by restarting consuming services

### LINKED ISSUE #724: `flagd` faults take time to propagate in astronomy shop
Essentially, the change in flagd is not actually picked up right away by the corresponding services (which is why we restarted some of them IIRC).

From a debugging session with an AI that seems correct:
```
No, it only restarts flagd, not the ad service. The injection flow is:

Update flagd-config ConfigMap — set adFailure.defaultVariant to "on"
kubectl rollout restart deployment flagd — so flagd picks up the new config
That’s it — no restart of the ad deployment

The assumption is that the ad service’s OpenFeature SDK is connected to flagd via gRPC EventStream and will pick up the flag change in real-time. But in practice, there’s a race:

flagd restarts → the ad service’s EventStream connection drops
The OpenFeature SDK falls back to the code-level default (false = no failure) while reconnecting
Only after the SDK re-establishes the EventStream and syncs the flag state does adFailure=true take effect
That reconnection can take minutes, explaining the ~6–30 minute delay before errors appear
```

This seems to be the root cause.

Thanks @tianyi-tz for finding this and reaching out!

Saadmrp1038:
Closes #724

Restart the affected consuming service alongside flagd after ConfigMap updates so the OpenFeature SDK reconnects with the correct flag value immediately.

Flag-to-deployment mapping based on https://opentelemetry.io/docs/demo/feature-flags/

### DIFF
--- a/sregym/generators/fault/inject_otel.py
+++ b/sregym/generators/fault/inject_otel.py
@@ -4,12 +4,50 @@ import subprocess
 from sregym.generators.fault.base import FaultInjector
 from sregym.service.kubectl import KubeCtl
 
+# Maps feature flags to the deployment(s) that consume them.
+# After restarting flagd, these services must also be restarted
+# so their OpenFeature SDK reconnects and picks up the new flag value
+# immediately instead of falling back to code-level defaults during
+# the gRPC EventStream reconnection window.
+# Source: https://opentelemetry.io/docs/demo/feature-flags/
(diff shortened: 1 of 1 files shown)


HacksonClark:
@tianyi-tz could you try this out?

HacksonClark (approved):
User confirmed it's effectiveness.

PR #727, opened 2026-04-10
Cleanup orphaned dirs from deleted PVs

### LINKED ISSUE #726: Cleanup orphaned directories from persistent volumes
PVs are being cleaned up, but not the underlying directories. When you run the benchmark long enough you'll run out of storage since there's effectively a leak.

HacksonClark:
Closes #726 

Going to test by doing one of the end to end runs and will go from there.


HacksonClark:
Contains some additional fixes related to `khaos`.

HacksonClark:
Validated through evals

PR #729, opened 2026-04-15
Rename train-ticket flags

Saadmrp1038:
The old flag names like `fault-22-sql-column-name-mismatch-error` leak the root cause directly. If the agent discovers the flag names, it can easily find the root cause and turn off the flag to mitigate the fault, bypassing the actual diagnosis work. This PR renames all 22 flags in the train-ticket `flagd` config to `tt-feat-01` through `tt-feat-22` and updates all references.

Related PR: https://github.com/xlab-uiuc/train-ticket/pull/12

Need to update the submodule version in `SREGym` after merging the train-ticket PR.

### DIFF
--- a/sregym/conductor/problems/train_ticket_f22.py
+++ b/sregym/conductor/problems/train_ticket_f22.py
@@ -15,7 +15,7 @@ class TrainTicketF22(Problem):
     def __init__(self):
         self.app_name = "train-ticket"
         self.faulty_service = "ts-contacts-service"
-        self.fault_name = "tt-feat-22"
+        self.fault_name = "fault-22-sql-column-name-mismatch-error"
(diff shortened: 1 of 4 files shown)


Saadmrp1038:
Another issue found. The agent can see the code of `sregym/resources/trainticket/locustfile.py` by using the following command:
```bash
kubectl get configmap locustfile-config -n train-ticket -o yaml
```

This leaks fault information. We should remove fault specific information from this file.

Saadmrp1038:
Since only 2 of the 22 flags are actual faults, I decided to turn on some decoy flags (they do nothing) during fault injection.

HacksonClark (approved):
Great work!

PR #730, opened 2026-04-15
Fix `astronomy-shop` feature flag being set to to invalid variant

### LINKED ISSUE #728: `payment_service_failure` feature flag defaulting to wrong value
payment_service_failure
[4:36 PM]defaultVariant: "on" is injected, but on is not a valid value, so the flagd SDK just defaults to 0, which didn't trigger any failure
[4:38 PM]You can maybe verify with your claude if the following statement is correct
Trace through the actual flagd source code on GitHub (open-feature/flagd):

- validateDefaultVariants() in json.go rejects the entire config when any flag has an invalid defaultVariant
- SetState() returns the error without calling store.Update()
- On fresh flagd start (after the injector restarts it), the flag store is empty
- getNumberValue("paymentFailure", 0) → FLAG_NOT_FOUND → SDK returns 0 → payment succeeds

Saadmrp1038:
Closes #728

## Problem

`inject_fault()` in `OtelFaultInjector` hardcodes `defaultVariant: "on"` for all flags except `imageSlowLoad`. However, several flags in the upstream OpenTelemetry demo [demo.flagd.json](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/flagd/demo.flagd.json) don't have `"on"` as a valid variant (though all of them do have `"off"` as a universal recovery variant).

- `paymentFailure` uses `"100%"`, `"90%"`, `"75%"`, `"50%"`, `"25%"`, `"10%"`, `"off"`
- `imageSlowLoad` uses `"10sec"`, `"5sec"`, `"off"` (was already special-cased)
- `emailMemoryLeak` uses `"1x"`, `"10x"`, `"100x"`, `"1000x"`, `"10000x"`, `"off"`

## Fix

Introduced a `_FLAG_INJECT_VARIANT` mapping for flags that don't use the standard `"on"` pattern. `inject_fault()` now looks up the correct variant from this map, falling back to `"on"` for standard boolean flags.

### DIFF
--- a/sregym/generators/fault/inject_otel.py
+++ b/sregym/generators/fault/inject_otel.py
@@ -28,6 +28,15 @@ _FLAG_TO_DEPLOYMENTS: dict[str, list[str]] = {
     "llmRateLimitError": ["llm"],
 }
 
+# Flags whose "enabled" variant is NOT the standard "on".
+# These flags use percentage/multiplier variants instead of boolean on/off.
+# Source: https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/flagd/demo.flagd.json
+_FLAG_INJECT_VARIANT: dict[str, str] = {
+    "imageSlowLoad": "10sec",
+    "paymentFailure": "100%",
+    "emailMemoryLeak": "1000x",
+}
+
 
 class OtelFaultInjector(FaultInjector):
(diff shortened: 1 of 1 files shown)


HacksonClark (approved):
Great work!

Issue #534, opened 2026-02-06
Throw error if you try and run an agent not in the registry


HacksonClark:
If you start the benchmark with an agent that does not exist, the benchmark *should* exit with an error. Instead, it just prints but the API server keeps running:

```
INFO:     Started server process [52938]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
[23:37:58] ⚠️ Agent 'debug' not found in registry. Available   main.py:63
           agents: dict_keys(['stratus', 'autosubmit',                  
           'codex', 'claudecode'])                                      
INFO:     Started server process [52938]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:9954 (Press CTRL+C to quit)
```

Saadmrp1038:
@HacksonClark I believe this issue is already resolved.

```bash
pial@pial-Redmi-Book-Pro-15-2022:~/Projects/SREGym$ uv run main.py --problem service_port_conflict_hotel_reservation --agent stratusss --model claude-sonnet-4-6 2>&1 | ansifilter | tee main.log
INFO - __main__ - 🔧 Config — agent: stratusss, agent_model: claude-sonnet-4-6, judge_model: claude-sonnet-4-6
╭─────────────────────────────────────── SREGym API Server ────────────────────────────────────────╮
│  ____  ____  _____ ____                                                                          │
│ / ___||  _ \| ____/ ___|_   _ _ __ ___                                                           │
│ \___ \| |_) |  _|| |  _| | | | '_ ` _ \                                                          │
│  ___) |  _ <| |__| |_| | |_| | | | | | |                                                         │
│ |____/|_| \_\_____\____|\__, |_| |_| |_|                                                         │
│                         |___/                                                                    │
│                                                                                                  │
╰────────────────────────────────────── http://0.0.0.0:8000 ───────────────────────────────────────╯
Available Endpoints                                                                                 

 • POST /submit: { "solution": "<your-solution>" } → grades the current stage                       
 • GET /status: returns { "stage": "setup" | "diagnosis" | "mitigation" | "tearing_down" | "done" } 
[...]

PR #737, opened 2026-04-20
fix: shutdown API server when agent not found and remove broken test

JiayuuWang:
## Summary

Two bug fixes:

1. **Bug fix (issue #534)**: When running with an agent that doesn't exist in the registry, the benchmark now properly shuts down the API server before exiting, instead of leaving it hanging.

2. **Test cleanup**: Removed `tests/file_editing/test_file_editing_tool.py` which imports from `clients.test_k8s_agent.k8s_agent` — a module that doesn't exist in the codebase. This test was broken and couldn't run.

## Changes

- `main.py`: Added `request_shutdown()` call before `sys.exit(1)` when agent is not found, ensuring the API server terminates properly.
- `tests/file_editing/test_file_editing_tool.py`: Deleted — the imported module `clients.test_k8s_agent.k8s_agent` does not exist.

## Test plan

- [ ] Run `python main.py --agent nonexistent` and verify the process exits cleanly (no lingering API server process)


HacksonClark:
This was closed in a separate PR.

PR #743, opened 2026-04-27
Manifest fault in `operator_non_existent_storage` correctly

yimingsu01:
This PR made sure faults in `operator_non_existent_storage` manifest logically on the cluster.


yimingsu01:
Hi @tianyi-tz , could you check if this PR fixes the issue? Thank you!

PR #757, opened 2026-05-18
Add service wrong pod selection benchmark for Hotel Reservation

jackj6441:
## The Real-World Failure Story

This problem is inspired by a Kubernetes Service selector misconfiguration where a Service selector is too broad or labels are accidentally shared across unrelated workloads.

In this kind of failure, the Service still has endpoints, but the endpoint list is polluted with pods that do not actually serve the expected traffic. This is harder to diagnose than a simple empty-endpoint failure because the pods may all be Running, the Service may look healthy at first glance, and the endpoint list is not empty.
Reference:
https://medium.com/emburse/intermittent-connection-refused-errors-after-new-service-deployment-in-kubernetes-22dd5feae1ad
However, some requests can still be routed to unintended pods, which may reject connections, time out, or return invalid responses.

## How We Simulate the Failure on SREGym

This PR adds a new SREGym benchmark problem:

`service_wrong_pod_selection_hotel_reservation`

The problem uses the Hotel Reservation application and targets the `frontend` Service.

Under normal behavior, the `frontend` Service selects frontend pods using:

```yaml
io.kompose.service: frontend
```

The injected fault changes the Service selection behavior by:

1. Adding `sregym.io/frontend-route=true` to the intended `frontend` deployment.
2. Adding `sregym.io/frontend-route=true` to the unintended `search` deployment.
3. Replacing the `frontend` Service selector with:

```yaml
sregym.io/frontend-route: "true"
```

After the fault is injected, the `frontend` Service selects both the real `frontend` pod and the unintended `search` pod.

```text
frontend Service
├── frontend pod  correct endpoint
└── search pod    wrong endpoint
```

The `frontend` Service routes traffic to targetPort `5000`, while the `search` container listens on port `8082`. Therefore, traffic routed to the `search` pod through the `frontend` Service can fail.

## The Problem Runtime Behavior

After fault injection:

- The Hotel Reservation application deploys successfully.
- Application pods can remain Running and Ready.
- The `frontend` Service still has endpoints.
- The `frontend` EndpointSlice is non-empty but contains both `frontend` and `search` pods.
- Requests through `frontend:5000` may fail because some traffic can be routed to the `search` pod, which does not serve frontend traffic on port `5000`.
- The failure is caused by Service endpoint pollution, not an empty endpoint set or pod crash.

The deterministic failure signal is EndpointSlice pollution. Curl failures are useful supporting evidence, but the mitigation oracle does not depend on the exact curl failure rate.

Useful debugging commands:

```bash
kubectl get svc frontend -n hotel-reservation -o yaml

kubectl get endpointslices -n hotel-reservation \
  -l kubernetes.io/service-name=frontend \
  -o yaml

kubectl get pods -n hotel-reservation --show-labels

kubectl describe svc frontend -n hotel-reservation
```

Expected broken-state evidence:

```text
frontend Service selector:
{"sregym.io/frontend-route":"true"}

frontend EndpointSlice contains:
pod=frontend-...
pod=search-...
```

## The Agent Behavior

A successful agent should diagnose that the `frontend` Service is selecting unintended pods because its selector was broadened to:

```yaml
sregym.io/frontend-route: "true"
```

The agent should identify that both the intended `frontend` pod and the unintended `search` pod match this selector.

The root cause is the `frontend` Service selector misconfiguration. It is not a pod crash, not an empty endpoint set, and not a failure inside the `search` service itself.

Expected diagnosis:

```text
The frontend Service selector is too broad. It selects both frontend pods and the search pod, causing the frontend Service endpoints to be polluted. The search pod does not serve frontend traffic on targetPort 5000, so traffic through the frontend Service can fail.
```

Expected mitigation is to restore the `frontend` Service selector to its original value:

```yaml
io.kompose.service: frontend
```

The mitigation should result in the `frontend` EndpointSlice containing only frontend pods.

The mitigation oracle checks that:

- The `frontend` Service selector is exactly restored to `{"io.kompose.service": "frontend"}`.
- Frontend endpoints are non-empty.
- All selected endpoint pods are actual frontend pods.
- No unintended `search` endpoint remains.

In manual validation, the diagnosis submission passed the LLM-as-a-judge evaluation with 100/100, and the mitigation oracle passed after restoring the frontend Service selector and removing the polluted endpoint.

## Validation

Local compile check passed:

```bash
PYTHONPATH=. uv run python -m py_compile \
  sregym/conductor/problems/service_wrong_pod_selection_hotel_reservation.py \
  sregym/conductor/oracles/wrong_pod_selection_mitigation.py \
  sregym/conductor/problems/registry.py
```

Registry check passed:

```text
True
```

End-to-end SREGym validation passed:

- The problem started successfully.
- Fault injection succeeded.
- The `frontend` Service selector changed to `{"sregym.io/frontend-route":"true"}`.
- The `frontend` EndpointSlice contained both `frontend` and `search` pods.
- Curl to `frontend:5000` showed connection failures as supporting runtime evidence.
- Diagnosis passed with 100/100.
- Mitigation restored the `frontend` Service selector.
- The `frontend` EndpointSlice contained only frontend pods.
- The mitigation oracle passed.

## Files Changed

```text
Problem List.md
sregym/conductor/problems/registry.py
sregym/conductor/tasklist.yml.example
sregym/conductor/problems/service_wrong_pod_selection_hotel_reservation.py
sregym/conductor/oracles/wrong_pod_selection_mitigation.py
```


tianyin:
@jackj6441 This is fast. I let @Saadmrp1038 to review it. 

> The Real-World Failure Story

Can you put the link of the real-world failures there as the evidence support that it is "real world"?

sregym/conductor/problems/service_wrong_pod_selection_hotel_reservation.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+
+    @mark_fault_injected
+    def recover_fault(self):
+        print("== Fault Recovery ==")
+        self._replace_service_selector(self.expected_service_selector)
+
+        for deployment in [self.frontend_service, self.wrong_deployment]:
+            self._set_pod_template_route_label(deployment, None)
+            self._wait_for_rollout(deployment)
```
Saadmrp1038: We should use the pattern consistent with other faults and use `VirtualizationFaultInjector` here to inject and recover the faults. You can check `sregym/conductor/problems/wrong_service_selector.py` for reference.

sregym/conductor/problems/service_wrong_pod_selection_hotel_reservation.py
```diff
@@ -0,0 +1,81 @@
...
+        self.mitigation_oracle = WrongPodSelectionMitigationOracle(problem=self)
+
+    @mark_fault_injected
+    def inject_fault(self):
+        print("== Fault Injection ==")
+        for deployment in [self.frontend_service, self.wrong_deployment]:
+            self._set_pod_template_route_label(deployment, self.route_label_value)
+            self._wait_for_rollout(deployment)
+
+        self._replace_service_selector(self.faulty_service_selector)
```
Saadmrp1038: We should use the pattern consistent with other faults and use `VirtualizationFaultInjector` here to inject and recover the faults. You can check `sregym/conductor/problems/wrong_service_selector.py` for reference.

sregym/conductor/problems/service_wrong_pod_selection_hotel_reservation.py
```diff
@@ -0,0 +1,81 @@
...
+from sregym.utils.decorators import mark_fault_injected
+
+
+class ServiceWrongPodSelectionHotelReservation(Problem):
+    def __init__(self):
+        self.app = HotelReservation()
+        self.namespace = self.app.namespace
+        self.frontend_service = "frontend"
+        self.wrong_deployment = "search"
+        self.route_label_key = "sregym.io/frontend-route"
```
Saadmrp1038: nit: Let's not expose the `sregym` name here to agent. Consider using something more generic.

sregym/conductor/oracles/wrong_pod_selection_mitigation.py
```diff
@@ -0,0 +1,72 @@
...
+
+        kubectl = self.problem.kubectl
+        namespace = self.problem.namespace
+        service_name = self.problem.frontend_service
+        expected_selector = self.problem.expected_service_selector
+        expected_pod_label = self.problem.expected_endpoint_pod_label
+
+        service = kubectl.get_service(service_name, namespace)
+        selector = service.spec.selector or {}
+        if selector != expected_selector:
```
Saadmrp1038: This feels too strict. If an agent fixes the problem by removing the label from the `search` pods instead of restoring the original selector this check will fail.

Saadmrp1038 (changes requested):
Thanks for the effort! I left some inline comments below, please ping me when you've addressed them.

I also had a couple of queries:

1. This looks quite similar to `wrong_service_selector`. Can you explain what's different here?

2. If possible can you run this problem end-to-end and drop the agent trace here? For example to run on the stratus agent you can:
```
uv run python main.py --problem service_wrong_pod_selection_hotel_reservation --agent stratus
```

---
Please feel free to ping us if you have any questions or confusions!

jackj6441:
> @jackj6441 This is fast. I let @Saadmrp1038 to review it.
> 
> > The Real-World Failure Story
> 
> Can you put the link of the real-world failures there as the evidence support that it is "real world"?

Thanks! I added the real-world failure link as evidence in the PR description. https://medium.com/emburse/intermittent-connection-refused-errors-after-new-service-deployment-in-kubernetes-22dd5feae1ad

jackj6441:
> Thanks for the effort! I left some inline comments below, please ping me when you've addressed them.
> 
> I also had a couple of queries:
> 
> 1. This looks quite similar to `wrong_service_selector`. Can you explain what's different here?
> 2. If possible can you run this problem end-to-end and drop the agent trace here? For example to run on the stratus agent you can:
> 
> ```
> uv run python main.py --problem service_wrong_pod_selection_hotel_reservation --agent stratus
> ```
> 
> Please feel free to ping us if you have any questions or confusions!

1. I agree that they are similar, the key difference is the endpoint state:

- `wrong_service_selector`: the Service selector does not match the intended pods, so the Service has no ready endpoints.
- `service_wrong_pod_selection_hotel_reservation`: the Service still has ready endpoints, but the EndpointSlice is polluted with both the intended `frontend` pod and the unintended `search` pod.

So this problem models a broader-selector failure where the Service can look partially healthy because endpoints are non-empty, but some traffic is routed to the wrong pod.

2. I also ran this locally with Stratus:

The problem deployed successfully and reached the intended injected state. The trace confirmed that the `frontend` Service selector was changed to `service-route=frontend`, causing the Service to select both the intended `frontend` pod and the unintended `search` pod.

Stratus did not solve the intended fault in this run. It focused on unrelated startup noise around `rate` / `mongodb-rate`, patched the `rate` Deployment with an initContainer, and submitted that as mitigation. The benchmark oracle correctly rejected the attempt because the polluted `frontend` Service endpoints were not fixed. The run keep retrying in worng direction so I stopped the loop with Ctrl-C.
deployment.apps/rate patched

Saadmrp1038:
@jackj6441 the changes look great!

tianyin:
@Saadmrp1038 We should try to merge it and make it an official problem of the benchmark.

Certainly, please keep giving feedback to help @jackj6441 improve the PR till the quality you want.

Saadmrp1038 (approved):

Saadmrp1038:
The code looks good. I also ran the problem and everything seems to be working correctly. I am merging the PR.
Congrats on the first PR merged here @jackj6441 🎉

PR #762, opened 2026-05-20
Add init_container_dependency_hang problem

JamesWeng0428:
# Add `init_container_dependency_hang_*` problems

Three new SREGym problems, one per supported app:
`init_container_dependency_hang_hotel_reservation`,
`init_container_dependency_hang_social_network`,
`init_container_dependency_hang_astronomy_shop`.

A lot of production deployments use an init container to wait on a downstream service before the main container starts. The Kubernetes docs themselves suggest this pattern, usually as `until nslookup <dep>; do sleep 2; done`. It is everywhere.

The failure mode is plain. A refactor renames or removes the dependency service, but the dependent deployment's init container still points at the old DNS name. The init container loops forever. Pods stay in `Init:0/1`. The rollout never finishes. Nothing crashes, because the old ReplicaSet's pods keep serving traffic, so the incident is quiet until somebody notices that new replicas are not coming up.

Some references:

- Kubernetes init-container docs, which recommend the `nslookup` loop pattern: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
- kubernetes/website#12050, the documentation issue about busybox `nslookup` exit-code regression. Relevant for reproduction, see Section 2.
- CubeAPM, "5 common causes of pods stuck in init": https://cubeapm.com/blog/kubernetes-pod-stuck-in-init-container/

This is not the same as anything already in the registry:

- `rolling_update_misconfigured_*` also uses an init container, but as a way to wedge a synthetic deployment so the diagnosed root cause is the rolling-update strategy. Here the root cause is the init container itself.
- `rbac_misconfiguration` makes an init container fail on a permission denial. Here the container has every permission it needs. It just loops on a name that never resolves.

### Code

| File | Purpose |
|---|---|
| `sregym/generators/fault/inject_virtual.py` | New `inject_init_container_dependency_hang` / `recover_init_container_dependency_hang`. |
| `sregym/conductor/problems/init_container_dependency_hang.py` | New `InitContainerDependencyHang(Problem)`. |
| `sregym/conductor/oracles/init_container_hang_mitigation.py` | New `InitContainerHangMitigationOracle`. |
| `sregym/conductor/problems/registry.py` | Three registry entries. |
| `Problem List.md` | Three rows. |

### Injection

1. Get the deployment YAML, strip status / managedFields / runtime metadata, save the cleaned manifest to `/tmp/<service>_init_dep_hang_original.yaml` so recovery is a clean round-trip.
2. Append an init container:
   ```yaml
   - name: wait-for-legacy-config
     image: busybox:1.28
     command:
       - /bin/sh
       - -c
       - |
         echo 'waiting for legacy-config-service to become ready...';
         until nslookup legacy-config-service.<ns>.svc.cluster.local >/dev/null 2>&1; do
           echo 'still waiting on legacy-config-service';
           sleep 5
         done
   ```
3. `kubectl apply` plus `rollout restart` so the fault is visible on the next ReplicaSet, not just on a future deploy.

`busybox:1.28` is pinned on purpose. See kubernetes/website#12050: in `busybox:1.30+`, `nslookup` returns 0 even on NXDOMAIN, which would silently break the fault.

One thing worth flagging here. An earlier version of this loop used `grep '^Address'` as an extra guard, thinking it would catch the case where DNS returned an empty answer. That turned out to be wrong, and it made the fault not reproduce. busybox 1.28's `nslookup` prints the resolver's own self-id line (`Address 1: 10.96.0.10 kube-dns.kube-system.svc.cluster.local`) before reporting NXDOMAIN. The grep matched immediately, the init container exited 0, and the pod went straight to Running. I caught it by looking at the init container status on a "Running" frontend pod and seeing `startedAt == finishedAt`. The fix is to just rely on the exit code, which is what busybox 1.28 is pinned for in the first place.

### Recovery

`kubectl apply -f /tmp/<service>_init_dep_hang_original.yaml`, then `rollout restart` and `rollout status --timeout=120s`. Falls back to stripping the injected container by name from the live state if the snapshot file is missing.

### Why this is reproducible

- No PVC, no helm install, no openebs. Just `kubectl apply`.
- One image, `busybox:1.28`, which is small and gets cached after the first pull.
- CoreDNS NXDOMAINs every time. The fault is deterministic.
- Re-injection is idempotent (drops any prior copy of the same init container before appending).

### Oracles

- **Diagnosis:** `LLMAsAJudgeOracle` against a structured root cause built from `Problem.build_structured_root_cause(...)`.
- **Mitigation:** `CompoundedOracle` of two checks. The generic `MitigationOracle` ensures every pod is Running. The new `InitContainerHangMitigationOracle` checks that the target deployment no longer has an init container with the canonical hang signature. The compound oracle rejects reward-hack solutions where pods happen to be Running but the broken init container is still there (e.g. an agent that deletes the namespace and lets the framework recreate it).

Two valid mitigations both pass:

1. Remove the `wait-for-legacy-config` init container from the deployment spec.
2. Repoint its `nslookup` target at a service that actually exists.

After `start init_container_dependency_hang_hotel_reservation`:

```
$ kubectl get pods -n hotel-reservation
NAME                       READY   STATUS     AGE
frontend-7fc7ddc9db-6s97p  1/1     Running    5m         # old RS, still serving
frontend-7bb6b6cbf-psc2z   0/1     Init:0/1   5m         # new RS, wedged
...

$ kubectl logs -n hotel-reservation frontend-7bb6b6cbf-psc2z -c wait-for-legacy-config
waiting for legacy-config-service to become ready...
still waiting on legacy-config-service
still waiting on legacy-config-service
...
```

The deployment never reaches the desired updated replica count. The default `MitigationOracle` reports failure because at least one pod is not Running.

A correct mitigation:

```
$ kubectl patch deployment frontend -n hotel-reservation --type=json \
    --patch-file /tmp/init_patch.json
# where init_patch.json = [{"op":"remove","path":"/spec/template/spec/initContainers"}]
$ kubectl rollout status deployment/frontend -n hotel-reservation
deployment "frontend" successfully rolled out
```

Both oracles now pass.

## 4. Agent behavior

I validated this end-to-end with the human-agent CLI (`cli.py`) on a 4-node kind cluster (ARM, macOS), using `deepseek/deepseek-chat` as the judge.

```
Diagnosis: 100.0  (composite 1.00; D1 1.00, D2 1.00, D3 1.00)
  All 9 checklist questions: "Yes", High confidence.
TTL: 529s

Mitigation: 100.0  (pods_running ✓, init_container_fixed ✓)
TTM: 801s
```

The mitigation I applied was the JSON-patch removal of `spec.template.spec.initContainers`. The compound oracle accepted it. Pods reached Running and the `wait-for-legacy-config` container was gone from the deployment spec.

I have not run Stratus against this problem yet. I don't have Bedrock set up on this machine, and DeepSeek's tool-calling is weak enough that a Stratus run on it would not be a meaningful baseline. The judge oracle and the compound mitigation oracle are both validated by the human run, so a Stratus run should plug in cleanly when somebody (or I) gets to it. I'm happy to do it as a follow-up if maintainers want the data point before merging.

---

## Files touched

```
sregym/conductor/oracles/init_container_hang_mitigation.py   (NEW)
sregym/conductor/problems/init_container_dependency_hang.py  (NEW)
sregym/conductor/problems/registry.py                        (3 new entries + 1 import)
sregym/generators/fault/inject_virtual.py                    (new inject/recover pair)
Problem List.md                                              (3 new rows)
```

No changes to the `Problem` base class, no changes to existing problems, no new dependencies.


Saadmrp1038:
@JamesWeng0428 thanks for the PR! I will review it soon.

sregym/conductor/oracles/init_container_hang_mitigation.py
```diff
@@ -0,0 +1,85 @@
...
+            name = c.get("name", "")
+            cmd_tokens = c.get("command", []) or []
+            args_tokens = c.get("args", []) or []
+            joined = " ".join(str(t) for t in cmd_tokens + args_tokens)
+
+            looks_like_ours = name == self.INJECTED_CONTAINER_NAME
+            still_hangs = self.HANG_TARGET_SUBSTR in joined
+
+            if looks_like_ours and still_hangs:
+                offending.append(c)
```
Saadmrp1038: This logic feels too strict. If an agent creates the `legacy-config-service` the ns-lookup will resolve. Since the command string still contains `legacy-config-service` oracle will fail. But this is a valid mitigation from the agent's perspective since with this fix init container will complete and the stuck pod will become ready.

I tested with the below commands. And as expected the mitigation oracle failed.

*Mitigation commands*
```
Pial@node0:~$ kubectl get pods -n hotel-reservation -l io.kompose.service=frontend
NAME                        READY   STATUS     RESTARTS   AGE
frontend-5cbb9bdd8f-hrd62   0/1     Init:0/1   0          5m23s
frontend-7fc7ddc9db-rgtzb   1/1     Running    0          6m1s

Pial@node0:~$ kubectl create service clusterip legacy-config-service -n hotel-reservation --tcp=80:80
service/legacy-config-service created

Pial@node0:~$ kubectl get pods -n hotel-reservation -l io.kompose.service=frontend
NAME                        READY   STATUS        RESTARTS   AGE
frontend-5cbb9bdd8f-hrd62   1/1     Running       0          5m43s
frontend-7fc7ddc9db-rgtzb   1/1     Terminating   0          6m21s

Pial@node0:~$ kubectl get pods -n hotel-reservation -l io.kompose.service=frontend                         NAME                        READY   STATUS    RESTARTS   AGE
frontend-5cbb9bdd8f-hrd62   1/1     Running   0          6m4s
```

*Mitigation oracle*
```
{'status': 'ok', 'message': 'Submission received'}
SREGym> submit()
[05/20/26 19:56:44] INFO     all.sregym.conductor - Evaluating stage 'mitigation'
                    INFO     all.sregym.conductor - Start Eval for Mitigation
== Mitigation Evaluation ==
== Init-Container Dependency-Hang Mitigation Evaluation ==
❌ Deployment frontend still has init container(s) wedged on `legacy-config-service`: ['wait-for-legacy-config']
[05/20/26 19:56:45] INFO     all.sregym.conductor - [EVAL] Mitigation Failed
```

Saadmrp1038 (changes requested):
@JamesWeng0428 the implementation looks good! I just found one issue with the oracle logic. It is explained below. Please ping me when you have addressed it.

---

JamesWeng0428:
Hey @Saadmrp1038, thanks for catching this — you're totally right, the oracle was too strict. Just pushed a fix (`aa21b1a1`), here's what was going on:

**The bug:** I was checking the init container's *command string* for `legacy-config-service` to decide if it still hangs. But that string never changes — it says `legacy-config-service` whether the service exists or not. So in your repro, you created the service, `nslookup` resolved, the init container finished, the pod went Ready... and the oracle still failed just because the command text was unchanged. That's a healthy cluster getting marked as failed, which is wrong. (And honestly the problem docstring itself lists "point it at a resolvable dependency" as a valid fix, so my oracle was contradicting our own spec.)

**The fix:** Instead of grepping the manifest text, the oracle now looks at the *actual runtime state* of the pods — specifically the init container statuses on the live pods of the target deployment. The hang is a runtime thing, not a manifest thing, so that's what we should be checking. New logic:

- No pods still running the `wait-for-legacy-config` init container → it was removed, or the deployment rolled to a new ReplicaSet without it → **pass**
- Pods have it, but every copy is `terminated: Completed` → the `nslookup` loop resolved (agent created the service or repointed it, like in your test) → **pass**
- Any copy still `running` → genuinely wedged → **fail**

So all three legit mitigations now pass: remove the init container, fix the command, *or* create the missing service.

**Anti-reward-hack guard still there:** it still confirms the hang is gone *because the injected init container specifically completed or got removed* — not because someone nuked the namespace and let the framework rebuild it.

I ran your exact repro against the new logic and it passes now. One small thing I'd appreciate a second opinion on: I match a pod to its deployment via the ReplicaSet owner-ref (`<deployment>-<hash>`, strip the hash). Works fine for hotel-reservation / social-network / astronomy-shop, but if you'd rather I match on the deployment's own `spec.selector.matchLabels` instead, easy swap — lmk.

Pinging you to take another look 🙏

Saadmrp1038 (changes requested):
@JamesWeng0428 thanks for addressing the oracle issue. The current change looks correct. Although looking at the code. It seems somewhat functionally similar to the generic `MitigationOracle`. The generic `MitigationOracle` should be able to catch all the scenarios that the custom `InitContainerHangMitigationOracle` can catch.

1. Init container removed -> pods are Running -> `MitigationOracle` passes
2. Init container completed -> pods are Running -> `MitigationOracle` passes
3. Init container still hanging -> pod phase is Pending (stuck in Init:0/1) -> `MitigationOracle` fails on phase != "Running"

Also worth noting: the anti-reward-hack case you mentioned in the code comment (agent deletes the namespace). This custom oracle would actually pass there, since `list_pods` returns empty, `target_pods` is empty, and it hits the `if not target_pods` -> success branch. The generic `MitigationOracle` would catch that by failing on empty pod list.

Is there a scenario where the custom oracle rejects something the generic one would miss? If not, we could simplify and just use `MitigationOracle` directly instead of the `CompoundedOracle`.

---
As for your question, the owner-ref approach is fine. No need to change.

JamesWeng0428:
@Saadmrp1038 great question — you pushed me to actually think this through, and you're more right than I'd like to admit. Let me be honest about where this lands.

**Your three cases are all correct.** For the normal mitigations (init container removed / completed / still hanging), `MitigationOracle` catches everything the custom oracle does. No disagreement there.

**There is one scenario the generic oracle misses — the reward hack the custom oracle was created for.** `MitigationOracle` loops over *every pod in the namespace* and only checks `phase == "Running"`. It has no idea which deployment a pod belongs to. Picture the wedged state:

- `frontend-<old>` → `1/1 Running` (old ReplicaSet still serving traffic)
- `frontend-<new>` → `0/1 Init:0/1`, phase `Pending` (the wedged one)

Now an agent does `kubectl delete pod frontend-<new>`, or deletes the namespace and lets the framework rebuild a clean app. The hung pod is just *gone*. The deployment was never fixed — the broken init-container spec is still there — but at eval time every *remaining* pod is `Running`, so `MitigationOracle` passes. That's exactly the "pods Running for unrelated reasons" hack the compound oracle's docstring calls out.

**But here's the honest part:** my *current* custom oracle doesn't fully close that gap either. It inspects live pods that currently exist — if the agent deletes the hung pod, my `target_pods` is empty too and it also returns success. So as written, you're basically right: it's near-redundant.

**So I don't think the answer is "delete the custom oracle" — it's "make it actually earn its place."** The fix is to check the **deployment spec**, not just live pods: if `wait-for-legacy-config` is still in `spec.template.spec.initContainers`, require that the dependency is genuinely resolvable — i.e. the `legacy-config-service` Service now exists, or the current ReplicaSet's pods completed it. That closes the delete-the-pod / recreate-the-namespace hack, which `MitigationOracle` structurally cannot.

I'll push that tightening shortly so the custom oracle covers something `MitigationOracle` provably can't — then the `CompoundedOracle` is justified. If you'd still rather drop it and accept that the namespace-recreation hack goes uncaught for this problem, that's a reasonable call too — just want the tradeoff to be explicit. lmk which way you'd like to go.

Saadmrp1038:
@JamesWeng0428 about the two cases you mentioned.

1. An agent does `kubectl delete pod frontend-<new>`
This should not be an issue as the ReplicaSet will spawn a new pod almost instantly and that would get stuck.

```bash
Pial@node0:~$ kubectl get pods -n hotel-reservation -l io.kompose.service=frontend -w
NAME                        READY   STATUS     RESTARTS   AGE
frontend-7fc7ddc9db-59nlv   1/1     Running    0          30m
frontend-b7cc54994-mvflz    0/1     Init:0/1   0          22m
frontend-b7cc54994-mvflz    0/1     Terminating   0          22m <- deleted stuck new pod
frontend-b7cc54994-6sk8l    0/1     Pending       0          0s
frontend-b7cc54994-6sk8l    0/1     Pending       0          0s
frontend-b7cc54994-6sk8l    0/1     Init:0/1      0          0s <- another new pod created and got stuck
frontend-b7cc54994-6sk8l    0/1     Init:0/1      0          2s 
```

2. Deletes the namespace and lets the framework rebuild a clean app
The SREGym framework does not recreate the app between agent submission and oracle evaluation. So this particular scenario should be impossible.

---
Also, one general note: please try to keep the comments more concise. It's harder to read and understand too verbose comments.

JamesWeng0428:
> @JamesWeng0428 about the two cases you mentioned.
> 
> 1. An agent does `kubectl delete pod frontend-<new>`
>    This should not be an issue as the ReplicaSet will spawn a new pod almost instantly and that would get stuck.
> 
> ```shell
> Pial@node0:~$ kubectl get pods -n hotel-reservation -l io.kompose.service=frontend -w
> NAME                        READY   STATUS     RESTARTS   AGE
> frontend-7fc7ddc9db-59nlv   1/1     Running    0          30m
> frontend-b7cc54994-mvflz    0/1     Init:0/1   0          22m
> frontend-b7cc54994-mvflz    0/1     Terminating   0          22m <- deleted stuck new pod
> frontend-b7cc54994-6sk8l    0/1     Pending       0          0s
> frontend-b7cc54994-6sk8l    0/1     Pending       0          0s
> frontend-b7cc54994-6sk8l    0/1     Init:0/1      0          0s <- another new pod created and got stuck
> frontend-b7cc54994-6sk8l    0/1     Init:0/1      0          2s 
> ```
> 
> 2. Deletes the namespace and lets the framework rebuild a clean app
>    The SREGym framework does not recreate the app between agent submission and oracle evaluation. So this particular scenario should be impossible.
> 
> Also, one general note: please try to keep the comments more concise. It's harder to read and understand too verbose comments.

Hi, Try now. Sorry about the long reply. I was trying out the claude feature. I will be aware next time!

Saadmrp1038:
@JamesWeng0428 no worries! I will check out the changes soon.

Saadmrp1038 (approved):

Saadmrp1038:
@JamesWeng0428 Great work! I tested the problem end-to-end and everything is working nicely.
Congrats on the first PR merged here! 🎉

PR #768, opened 2026-05-22
Add node conntrack exhaustion hotel reservation problem

munimthahmid:
This PR adds a new SREGym problem that reproduces a production Kubernetes failure class where the visible symptoms look like application, DNS, or observability flakiness, but the real fault is node-level Linux connection-tracking exhaustion.

The new benchmark problem is:

```text
node_conntrack_exhaustion_hotel_reservation
```

This problem is inspired by Mark Betz's production incident report, ["Exhausting conntrack table space crippled our k8s cluster"](https://www.markbetz.net/2023/12/12/exhausting-conntrack-table-space-crippled-our-k8s-cluster/), also mirrored on [Medium](https://medium.com/@betz.mark/exhausting-conntrack-table-space-crippled-our-k8s-cluster-98564f6f34e0).

In that incident, the first symptoms looked like ordinary service/network trouble: slow RPC services, intermittent failures, and DNS-looking errors such as temporary name-resolution failures. The actual root cause was lower-level. A small number of Kubernetes nodes had accumulated enough active network flows to fill the Linux `nf_conntrack` table. Once that table was full, new connection attempts could be dropped even though pods, services, endpoints, and the control plane could still look mostly normal.

The interesting operational lesson is that this is not primarily a "pod crashed" failure. It is a node networking state-exhaustion failure. Existing connections may continue, many Kubernetes resources still look healthy, and the misleading symptoms can send responders toward CoreDNS, Jaeger, Consul, databases, or random application logs before they inspect the node's `nf_conntrack_count` and `nf_conntrack_max`.

This PR reproduces that failure mode in SREGym: a connection-heavy workload is concentrated on one worker node until that node's conntrack table is exhausted, causing fresh Hotel Reservation traffic from that node to time out while the application objects still appear mostly healthy.

The new problem lives in:

```text
sregym/conductor/problems/node_conntrack_exhaustion.py
```

Fault injection creates two support workloads inside the `hotel-reservation` namespace:

```text
edge-traffic-client  ->  rpc-gateway:9090
```

`rpc-gateway` is a small Python TCP listener. It accepts connections and holds them open.

`edge-traffic-client` is pinned to the victim worker node. Each client pod opens many TCP connections to `rpc-gateway` and keeps them open. This fills the victim node's Linux conntrack table, not a Kubernetes Service endpoint table.

The injection is calibrated against the victim node's actual conntrack capacity:

```text
/proc/sys/net/netfilter/nf_conntrack_count
/proc/sys/net/netfilter/nf_conntrack_max
```

The current implementation reads `nf_conntrack_max` at injection time and computes enough client replicas to target roughly 105% of the table capacity, capped by `max_client_replicas` so an unexpectedly large local cluster fails early instead of creating an unbounded workload.

On the local kind cluster used for validation, `nf_conntrack_max` was `262144`, so the injected workload is approximately:

```text
28 edge-traffic-client pods * 10,000 held TCP connections per pod
```

That scale is large enough to saturate the node conntrack table while still keeping the benchmark contained to the local kind cluster.

## 3. Runtime Behavior

After fault injection, the failure is intentionally non-obvious:

```text
kubectl get deploy -n hotel-reservation
kubectl get svc -n hotel-reservation
```

These can look mostly healthy. The failure is not that the frontend Service has no endpoints, not that CoreDNS is deleted, and not that Jaeger is broken.

The deterministic broken-state signal is on the victim node:

```text
nf_conntrack_count / nf_conntrack_max ~= 1.0
```

Useful debugging commands:

```bash
kubectl get pods -n hotel-reservation -o wide
kubectl describe deploy edge-traffic-client -n hotel-reservation
kubectl logs -n hotel-reservation -l app=edge-traffic-client --tail=20
kubectl describe deploy rpc-gateway -n hotel-reservation
```

```text
edge-traffic-client pods are concentrated on one worker node
edge-traffic-client logs show held connection counts increasing
the victim node's nf_conntrack_count approaches nf_conntrack_max
fresh frontend requests from the victim node intermittently time out or fail
```

The failure should be diagnosed as node-level conntrack exhaustion caused by `edge-traffic-client` opening excessive held TCP connections to `rpc-gateway`.

## 4. Expected Diagnosis

A correct diagnosis should identify:

```text
Root cause:
The victim worker node's Linux nf_conntrack table is exhausted.

Faulty workload pattern:
edge-traffic-client is pinned to the victim node and opens many held TCP connections to rpc-gateway.

Failure mechanism:
Once nf_conntrack is full, new connections from that node can time out or be dropped even though Kubernetes pods/services/endpoints mostly look healthy.
```

Common incorrect diagnoses include:

```text
CoreDNS failure
Jaeger/tracing failure
Consul bootstrap issue
MongoDB or memcached outage
frontend application bug
empty Service endpoints
```

Those are symptoms or distractions, not the root cause.

## 5. Expected Mitigation

A correct mitigation should reduce the connection pressure and verify that the victim node recovers.

Valid fixes include:

```text
delete edge-traffic-client
scale edge-traffic-client down enough to reduce conntrack pressure
optionally delete rpc-gateway as cleanup after the connection source is removed or reduced
```

The mitigation oracle is implemented in:

```text
sregym/conductor/oracles/conntrack_mitigation.py
```

It intentionally goes beyond the generic "all pods are Running" check. That generic check is insufficient for this failure because pods can be Running while the node conntrack table remains full.

The custom oracle verifies:

```text
generic MitigationOracle passes
victim node conntrack ratio is below the oracle threshold
edge-traffic-client is deleted or reduced to a small fraction of the injected scale
a frontend probe scheduled onto the victim node succeeds
```

This catches symptom-only fixes where the application objects look healthy but node networking is still saturated.

## 6. Validation

Local validation was run repeatedly on a 4-node kind cluster.

Clean scored results:

| Agent | Result | Diagnosis | Mitigation | Notes |
| --- | --- | ---: | ---: | --- |
| ClaudeCode | End-to-end pass | `89/100` | Pass | Correctly identified conntrack exhaustion; conntrack dropped to about `0.29%`. |
| ClaudeCode | End-to-end pass | `100/100` | Pass | Correctly identified the node conntrack exhaustion and mitigated successfully; conntrack dropped to about `0.27%`. |
| ClaudeCode | Correct diagnosis, rejected mitigation | `100/100` | Fail | Diagnosis was correct, but mitigation did not reduce conntrack pressure enough; oracle rejected it. |
| ClaudeCode | Correct diagnosis, agent timeout | `89/100` | Timeout | Diagnosis was correct, but the agent hit its timeout before mitigation completed. |
| Stratus GPT-5 | Recovered cluster with wrong diagnosis | `11/100` | Pass | Stratus removed the synthetic workloads and recovered the cluster, but diagnosed gateway/socket exhaustion rather than node conntrack exhaustion. |

Stratus difficulty signal:

Across seven scored Stratus GPT-5 attempts, diagnosis did not pass. One attempt still recovered the cluster by deleting the synthetic workloads, but the submitted diagnosis missed the node-level conntrack mechanism. Six scored attempts produced failed diagnoses and then timed out before a successful mitigation result. Common incorrect diagnoses included Jaeger/tracing misconfiguration, Consul startup noise, CoreDNS degradation, profile database connectivity, and generic service bootstrap problems.

This is consistent with the intended difficulty: the visible symptoms are misleading unless the agent connects workload placement and TCP connection fanout to host-level `nf_conntrack` state.

Several additional local Stratus runs were excluded because LiteLLM/OpenAI retry exhaustion occurred before the agent produced a meaningful scored result. These are disclosed for completeness, but they are not counted as problem failures because the agent did not receive a completed model response.

No new Python dependencies are added.

No agent RBAC expansion is required.

Fault recovery is idempotent: if an agent already deletes or scales down the synthetic support resources, `recover_fault()` tolerates the missing resources and waits for conntrack usage to drop.


Saadmrp1038:
Thanks for the PR! I will review it soon.

HacksonClark:
/validate-problem node_conntrack_exhaustion_hotel_reservation

Saadmrp1038 (commented):
@munimthahmid when I try to run this fault the code is crashing. Tried both on kind and real clusters. Crashed both times. Can you please check why this is happening?

*kind cluster*
```
[05/22/26 12:44:34] INFO     all.infra.workload - Job created: wrk2-job
                    INFO     all.sregym.conductor - [ENV] Start workload
                    INFO     all.sregym.conductor - App deployed.
== Fault Injection ==
Victim node: kind-worker | Gateway node: kind-worker2
  File "/users/Pial/Projects/SREGym/cli.py", line 168, in <module>
    asyncio.run(main())
  File "/users/Pial/.local/share/uv/python/cpython-3.12.13-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/users/Pial/.local/share/uv/python/cpython-3.12.13-linux-x86_64-gnu/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/Pial/.local/share/uv/python/cpython-3.12.13-linux-x86_64-gnu/lib/python3.12/asyncio/base_events.py", line 691, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/users/Pial/Projects/SREGym/cli.py", line 154, in main
    result = await conductor.start_problem()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/Pial/Projects/SREGym/sregym/conductor/conductor.py", line 440, in start_problem
    self._advance_to_next_stage(start_index=0)
  File "/users/Pial/Projects/SREGym/sregym/conductor/conductor.py", line 288, in _advance_to_next_stage
    self._inject_fault()
  File "/users/Pial/Projects/SREGym/sregym/conductor/conductor.py", line 220, in _inject_fault
    problem.inject_fault()
  File "/users/Pial/Projects/SREGym/sregym/utils/decorators.py", line 4, in wrapper
    result = method(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/Pial/Projects/SREGym/sregym/conductor/problems/node_conntrack_exhaustion.py", line 49, in inject_fault
    self._calibrate_client_replicas()
  File "/users/Pial/Projects/SREGym/sregym/conductor/problems/node_conntrack_exhaustion.py", line 87, in _calibrate_client_replicas
    _, maximum = read_node_conntrack_usage(self.kubectl, self.victim_node)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/Pial/Projects/SREGym/sregym/conductor/oracles/conntrack_mitigation.py", line 16, in read_node_conntrack_usage
    output = subprocess.run(
             ^^^^^^^^^^^^^^^
  File "/users/Pial/.local/share/uv/python/cpython-3.12.13-linux-x86_64-gnu/lib/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['docker', 'exec', 'kind-worker', 'bash', '-lc', 'cat /proc/sys/net/netfilter/nf_conntrack_count; cat /proc/sys/net/netfilter/nf_conntrack_max']' returned non-zero exit status 1.
```

*real cluster*
```
[05/22/26 23:28:27] INFO     all.infra.workload - Job created: wrk2-job                  
                    INFO     all.sregym.conductor - [ENV] Start workload                 
                    INFO     all.sregym.conductor - App deployed.                        
== Fault Injection ==
Victim node: node1.exp-242212.aiopslab-pg0.wisc.cloudlab.us | Gateway node: node2.exp-242212.aiopslab-pg0.wisc.cloudlab.us
  File "/home/pial/Projects/SREGym/cli.py", line 168, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/pial/Projects/SREGym/cli.py", line 154, in main
    result = await conductor.start_problem()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pial/Projects/SREGym/sregym/conductor/conductor.py", line 440, in start_problem
    self._advance_to_next_stage(start_index=0)
  File "/home/pial/Projects/SREGym/sregym/conductor/conductor.py", line 288, in _advance_to_next_stage
    self._inject_fault()
  File "/home/pial/Projects/SREGym/sregym/conductor/conductor.py", line 220, in _inject_fault
    problem.inject_fault()
  File "/home/pial/Projects/SREGym/sregym/utils/decorators.py", line 4, in wrapper
    result = method(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pial/Projects/SREGym/sregym/conductor/problems/node_conntrack_exhaustion.py", line 49, in inject_fault
    self._calibrate_client_replicas()
  File "/home/pial/Projects/SREGym/sregym/conductor/problems/node_conntrack_exhaustion.py", line 87, in _calibrate_client_replicas
    _, maximum = read_node_conntrack_usage(self.kubectl, self.victim_node)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pial/Projects/SREGym/sregym/conductor/oracles/conntrack_mitigation.py", line 32, in read_node_conntrack_usage
    raise RuntimeError(f"Could not parse conntrack usage from {node_name}: {output}")
RuntimeError: Could not parse conntrack usage from node1.exp-242212.aiopslab-pg0.wisc.cloudlab.us: error: unknown flag: --rm
See 'kubectl debug --help' for usage.
```

sregym/conductor/oracles/conntrack_mitigation.py
```diff
@@ -0,0 +1,114 @@
...
+            check=True,
+            text=True,
+            timeout=10,
+        ).stdout
+    else:
+        cmd = (
+            f"kubectl debug node/{node_name} --quiet --attach=true --rm --image=busybox -- "
+            f"chroot /host sh -c {shlex.quote(CONNTRACK_CMD)}"
+        )
+        output = kubectl.exec_command(cmd)
```
Saadmrp1038: I don't think `--rm` is a flag on `kubectl debug`. (ref: https://github.com/kubernetes/kubectl/issues/1681)

```
Pial@node0:~$ kubectl debug node/kind-worker -it --rm --image=busybox -- chroot /host sh -c 'echo hello'
error: unknown flag: --rm
See 'kubectl debug --help' for usage.

Pial@node0:~$ kubectl debug node/kind-worker -it --image=busybox -- chroot /host sh -c 'echo hello'
Creating debugging pod node-debugger-kind-worker-6gmbt with container debugger on node kind-worker.
hello
```

Saadmrp1038 (changes requested):
The kind crash happened because my user was not added to the docker group and I just used `sudo docker` for any docker related work which is not common. The current `docker exec` approch should be fine.

Please take a look at the other crash. I marked the possible reason in the comment below.

sregym/conductor/oracles/conntrack_mitigation.py
munimthahmid: @Saadmrp1038   Yes, that was the issue. I replaced the kubectl debug --rm path with a short-lived hostNetwork pod pinned to the target node, and clean it up in a finally block. I also kept the fast docker exec path for kind, with fallback if docker is unavailable.

I re-ran the conntrack problem end-to-end on kind after this change, and it completed successfully.

Saadmrp1038: @munimthahmid this works now but ran into another issue. The code is crashing with the following error on a real cluster:

```bash
File "/home/pial/Projects/SREGym/sregym/conductor/problems/node_conntrack_exhaustion.py", line 91, in _calibrate_client_replicas
    raise RuntimeError(
RuntimeError: node_conntrack_exhaustion_hotel_reservation requires 138 client replicas for nf_conntrack_max=1310720, above cap 80
```

The CloudLab cluster I use has a much higher `nf_conntrack_max` value than kind. Around ~1.3M.

```bash
kubectl run conntrack-check --image=busybox --restart=Never --rm -it \
  --overrides='{"spec":{"hostNetwork":true,"nodeName":"node1.exp-242212.aiopslab-pg0.wisc.cloudlab.us"}}' \
  -- cat /proc/sys/net/netfilter/nf_conntrack_max
1310720
```

The simplest fix would be to bump the `max_client_replicas` value to something higher. But that would mean during fault injection ~130+ pods will be created, each holding 10k connections. That's quite a large number and could have unintended side effects.

I had a few suggestions that you can look into:

1. A cleaner fix might be to set the `nf_conntrack_max` value to something lower but still realistic during fault injection, like 262144 or 131072 (if this is common enough). During fault recovery we can reset the value to something normal like 262144 (if we set it below that during injection).

2. Currently we are spinning up multiple identical pods. This can actually make it very easy for the AI agent to detect something's fishy going on here. Can you explore ways where it might be possible to shrink down to ideally only one client pod that holds all the connections? Maybe the gateway can listen on multiple ports and the pod opens connections to each of those ports. The namespace would look much less suspicious without 20+ instances of the same pod.

munimthahmid: @Saadmrp1038  I’m thinking of making two changes:

1. During injection, temporarily set nf_conntrack_max on the victim node to a smaller bounded value, such as 262144, and restore that node’s original value during recovery.

2. Replace the many edge-traffic-client replicas with one client pod. The gateway would listen on multiple ports, and that single client pod would open and hold connections across those ports.

Using different destination ports allows one pod to create enough distinct TCP connections to fill the bounded conntrack table without creating dozens of identical pods.

Would implementing both changes in this way be reasonable?

sregym/conductor/oracles/conntrack_mitigation.py
```diff
@@ -0,0 +1,154 @@
...
+        probe_ok = self._frontend_probe_succeeds(node_name)
+
+        results.update(
+            {
+                "conntrack_ratio": ratio,
+                "client_reduced": client_reduced,
+                "frontend_probe": probe_ok,
+                "success": ratio < self.ratio_threshold and client_reduced and probe_ok,
+            }
+        )
```
Saadmrp1038: This feels too strict. The `_client_reduced` logic forces one specific fix (scale down `edge-traffic-client`) and rejects several valid mitigations. For example:

- Increasing `nf_conntrack_max` on the node
- Blocking client to gateway network calls

These might not be the most ideal fix but still valid mitigation.

sregym/conductor/oracles/conntrack_mitigation.py
```diff
@@ -0,0 +1,154 @@
...
+    if len(values) < 2:
+        raise RuntimeError(f"Could not parse conntrack usage from {node_name}: {output}")
+    return values[0], values[1]
+
+
+def _read_conntrack_with_host_network_pod(kubectl, node_name: str, namespace: str) -> tuple[int, int]:
+    core_v1 = getattr(kubectl, "core_v1_api", client.CoreV1Api())
+    pod_name = f"conntrack-read-{int(time.time() * 1000)}"
+    pod = {
+        "metadata": {"name": pod_name, "namespace": namespace, "labels": {"app": "conntrack-read"}},
```
Saadmrp1038: nit: helper pod names contain `conntrack-read` and runs during injection. An agent watching `kubectl get events -n hotel-reservation` could probably get hints about the fault by seeing event names. We should name it something generic.

Saadmrp1038 (changes requested):

sregym/conductor/oracles/conntrack_mitigation.py
Saadmrp1038: @munimthahmid Yup, sounds good!

sregym/conductor/oracles/conntrack_mitigation.py
Saadmrp1038: This check becomes a no-op if we move to a single client pod (1 <= 1 is always true), so we can drop it then.

munimthahmid:
@Saadmrp1038 I pushed the latest changes.

I addressed the review points by bounding nf_conntrack_max during injection and restoring the original value during recovery, replacing the many client replicas with a single multiprocessing client plus multi-port gateway, pinning both to the same victim node, keeping the kind docker exec path with a portable hostNetwork fallback, using generic helper pod names, and relaxing mitigation scoring so valid fixes are accepted based on conntrack recovery plus frontend health.
Please take another look when you get a chance.

sregym/conductor/problems/node_conntrack_exhaustion.py
```diff
@@ -0,0 +1,292 @@
...
+            while True:
+                ready, _, _ = select.select(listeners, [], [])
+                for listener in ready:
+                    c, _ = listener.accept(); held.append(c)
+            """
+        )
+        return {
+            "name": "gateway",
+            "image": "python:3.12-alpine",
+            "command": ["python", "-c", script],
```
Saadmrp1038: The default container soft ulimit is 1024 file descriptors on most clusters (in my case CloudLab Cluster). The client needs ~275,000 sockets. I got the following error:
```
OSError: [Errno 24] No file descriptors available
```

Passing the script via env var and raising ulimit before exec worked for me:
```
"command": ["sh", "-c", "ulimit -n 524288; exec python -c \"$SCRIPT\""],
"env": [
    ...,
    {"name": "SCRIPT", "value": script},
],
```

sregym/conductor/problems/node_conntrack_exhaustion.py
```diff
@@ -0,0 +1,292 @@
...
+                maximum = open("/host-proc/sys/net/netfilter/nf_conntrack_max").read().strip()
+                print(f"workers={len(workers)} target={total} nf_conntrack_count={count} nf_conntrack_max={maximum}", flush=True)
+                time.sleep(2)
+            raise RuntimeError("client connection worker stopped")
+            """
+        )
+        return {
+            "name": "client",
+            "image": "python:3.12-alpine",
+            "command": ["python", "-c", script],
```
Saadmrp1038: Same ulimit issue as the client. The gateway needs to `accept()` and hold ~275k sockets.

sregym/conductor/problems/node_conntrack_exhaustion.py
```diff
@@ -0,0 +1,292 @@
...
+            workers, remaining = [], total
+            while remaining:
+                goal = min(per_worker, remaining)
+                worker = mp.Process(target=hold_connections, args=(goal,))
+                worker.start(); workers.append(worker)
+                remaining -= goal
+            while all(worker.is_alive() for worker in workers):
+                count = open("/host-proc/sys/net/netfilter/nf_conntrack_count").read().strip()
+                maximum = open("/host-proc/sys/net/netfilter/nf_conntrack_max").read().strip()
+                print(f"workers={len(workers)} target={total} nf_conntrack_count={count} nf_conntrack_max={maximum}", flush=True)
```
Saadmrp1038: An agent checking logs will immediately see these print statements. The `nf_conntrack_count` and `nf_conntrack_max` in the logs will leak the fault mechanism to the agent. We should either remove this print or not say conntrack here.

Saadmrp1038 (changes requested):
@munimthahmid some minor issues found but other than those everything looks okay. I will merge the PR after you address them.

---
On a side note, I did some agent runs on this fault. Claude code can detect something fishy is going on but can't actually pinpoint the fault (I removed the print statement while testing). It notices abnormal behaviour like `kubectl exec` failing on the victim node and realizes there's something wrong with the clulster's networking. But it gets stuck in debuggin spiral and times out.

Saadmrp1038:
/validate-problem node_conntrack_exhaustion_hotel_reservation

munimthahmid:
@Saadmrp1038 I addressed the minor issues and pushed the update. Both the gateway and client now raise ulimit -n before execing Python,the scripts are passed through env vars, and I removed the conntrack-specific client log output plus the unused /host-proc mount.

On the side note, that matches what I saw as well. The fault was already subtle, and now that the explicit conntrack log hint is removed, it becomes even harder to diagnose. In most of my runs, the agent spent a lot of time in diagnosis and timed out during mitigation; after increasing the timeout by 10 minutes, it was able to mitigate, although that run still failed diagnosis with 67/100. 

Do you think we should tune anything to make the fault more diagnosable?

Saadmrp1038 (approved):

Saadmrp1038:
@munimthahmid Great work! 

---

> Do you think we should tune anything to make the fault more diagnosable?

Let's keep it as it is now. Since the fault mechanism itself is working correctly, we can think about what the agent would need to do later. This is a really hard problem for the agent, since the agent can't really exec/SSH into the victim node (as conntrack is full) and run diagnostics on the victim node directly.

tianyin:
> This is a really hard problem for the agent

hard is good :))

Great work @munimthahmid !! Thank you for working hard towards the end!

PR #776, opened 2026-05-24
Add ImagePullBackOff problem variants for Hotel Reservation

varuni7:
### Problem 1 — Wrong image tag (`image_pull_backoff_hotel_reservation`)

On December 17, 2023, **Logto** (an open-source authentication platform) suffered an 18-minute full outage. An automated GitHub Container Registry image retention workflow deleted production image layers that were untagged but still referenced by the live `prod` tag (a multi-arch manifest structure). When Kubernetes tried to schedule new pods, it could not pull the image, causing `ImagePullBackOff` across all instances.

> Reference: https://blog.logto.io/postmortem-docker-image-not-found

### Problem 2 — Invalid registry credentials (`image_pull_wrong_secret_hotel_reservation`)

In June 2022, **Airwallex** (a payments company) experienced nearly a full day of degradation after a GKE cluster auto-upgrade broke the `gke-metadata-server` endpoint. Nodes could no longer fetch Google Cloud credentials for pulling images from GCR. All image pulls failed with authentication errors, preventing pod restarts and new deployments.

> Reference: https://medium.com/airwallex-engineering/how-we-deal-with-a-google-kubernetes-engine-gke-metadata-server-outage-b627958ee7ad

---

## How the failures are simulated on SREGym

**Problem 1** — The `search` deployment's container image tag is replaced with `nonexistent-tag-xyzzy` (a tag that does not exist in the registry). Kubernetes attempts to pull it, fails, and the pod enters `ErrImagePull` → `ImagePullBackOff`.

**Problem 2** — The `profile` deployment's image is pointed at a fake private registry (`fake-private-registry.sregym.internal`) and an `imagePullSecret` with invalid base64-encoded credentials is injected. Kubernetes cannot authenticate to the unreachable registry, causing the same `ErrImagePull` → `ImagePullBackOff` failure but with a distinct root cause.

Both faults use `VirtualizationFaultInjector` and target Hotel Reservation microservices not already used by other problems (`search` and `profile`). They are distinct from the existing `incorrect_image` problem which targets AstronomyShop using `ApplicationFaultInjector`.

---

## Problem runtime behavior

1. All Hotel Reservation pods start healthy.
2. Fault is injected — the target pod immediately enters `ErrImagePull` then `ImagePullBackOff`.
3. All other pods remain `Running`; only the faulted service is affected.
4. Agent investigates with `kubectl get pods`, `kubectl describe pod`, reads the image pull error event.
5. **Diagnosis stage**: agent identifies the root cause (wrong tag / bad credentials).
6. **Mitigation stage**: agent patches the deployment to restore a valid image and removes the fake pull secret (Problem 2).
7. A custom `ImagePullBackoffMitigationOracle` verifies the fix is genuine — it rejects shortcuts such as deleting the deployment or scaling it to zero replicas.

---

## Agent behavior

Both problems were tested end-to-end using the SREGym CLI as the human agent with `gemini/gemini-2.5-pro` as judge:

| Problem | Diagnosis score | Mitigation | TTL | TTM |
|---|---|---|---|---|
| `image_pull_backoff_hotel_reservation` | 100/100 ✅ | success ✅ | 56s | 92s |
| `image_pull_wrong_secret_hotel_reservation` | 100/100 ✅ | success ✅ | 80s | 114s |

The oracle correctly rejected shortcuts (deleting the deployment, scaling to 0 replicas) and only passed when the image was genuinely fixed.

---

## Files changed

- `sregym/conductor/problems/image_pull_backoff.py` — new problem (wrong tag)
- `sregym/conductor/problems/image_pull_wrong_secret.py` — new problem (bad credentials)
- `sregym/conductor/oracles/image_pull_backoff_mitigation.py` — strict mitigation oracle
- `sregym/generators/fault/inject_virtual.py` — two new injector/recovery method pairs
- `sregym/conductor/problems/registry.py` — register both problems
- `sregym/observer/loki/loki-values.yaml` — disable chunksCache/resultsCache for local kind clusters (fixes OOM on machines with < 16GB RAM)

/validate-problem image_pull_backoff_hotel_reservation
/validate-problem image_pull_wrong_secret_hotel_reservation

### DIFF
--- a/sregym/conductor/oracles/image_pull_backoff_mitigation.py
+++ b/sregym/conductor/oracles/image_pull_backoff_mitigation.py
@@ -1,0 +1,92 @@
+import time
+
+from sregym.conductor.oracles.base import Oracle
+
+_ROLLOUT_SETTLE_SECONDS = 60
+_ROLLOUT_POLL_INTERVAL = 5
+_INVALID_TAG = "nonexistent-tag-xyzzy"
+
+
+class ImagePullBackoffMitigationOracle(Oracle):
+    """Stricter mitigation oracle for ImagePullBackoff problems.
+
+    Checks:
+    1. The faulty deployment still exists (not deleted as a shortcut).
+    2. The deployment has at least 1 desired replica (not scaled to 0).
+    3. The container image no longer uses the injected invalid tag.
+    4. All pods in the namespace are Running and ready.
+    """
+
+    importance = 1.0
+
+    def __init__(self, problem, faulty_service: str, invalid_tag: str = _INVALID_TAG):
+        super().__init__(problem)
+        self.faulty_service = faulty_service
+        self.invalid_tag = invalid_tag
+
+    def _wait_for_rollouts(self, kubectl, namespace):
+        deadline = time.monotonic() + _ROLLOUT_SETTLE_SECONDS
+        while time.monotonic() < deadline:
+            deployments = kubectl.list_deployments(namespace)
+            all_settled = all(
+                (dep.status.updated_replicas or 0) >= (dep.spec.replicas or 1)
+                and (dep.status.ready_replicas or 0) >= (dep.spec.replicas or 1)
+                and (dep.status.unavailable_replicas or 0) == 0
+                for dep in deployments.items
+            )
+            if all_settled:
+                return
+            time.sleep(_ROLLOUT_POLL_INTERVAL)
+        print("⚠️ Timed out waiting for rollouts to settle; evaluating current state")
+
+    def evaluate(self) -> dict:
+        print("== Mitigation Evaluation (ImagePullBackoff) ==")
+
+        kubectl = self.problem.kubectl
+        namespace = self.problem.namespace
+
+        # 1. Deployment must still exist
+        try:
+            deployment = kubectl.get_deployment(self.faulty_service, namespace)
+        except Exception:
+            print(f"❌ Deployment '{self.faulty_service}' not found — deleting it is not a valid fix")
+            return {"success": False}
+
+        # 2. Deployment must have >= 1 desired replica
+        desired = deployment.spec.replicas or 0
+        if desired < 1:
+            print(
+                f"❌ Deployment '{self.faulty_service}' scaled to {desired} replicas — scaling to 0 is not a valid fix"
+            )
+            return {"success": False}
+
+        # 3. Container image must not still use the invalid tag
+        containers = deployment.spec.template.spec.containers
+        for container in containers:
+            if self.invalid_tag in (container.image or ""):
+                print(f"❌ Container '{container.name}' still uses invalid image: {container.image}")
+                return {"success": False}
+        print(f"✅ Deployment '{self.faulty_service}' image tag is fixed")
+
+        # 4. All pods must be Running and ready
+        self._wait_for_rollouts(kubectl, namespace)
+        pod_list = kubectl.list_pods(namespace)
+
+        if not pod_list.items:
+            print("❌ No pods found in namespace")
+            return {"success": False}
+
+        for pod in pod_list.items:
+            if pod.status.phase != "Running":
+                print(f"❌ Pod {pod.metadata.name} is in phase: {pod.status.phase}")
+                return {"success": False}
+            for cs in pod.status.container_statuses or []:
+                if cs.state.waiting and cs.state.waiting.reason:
+                    print(f"❌ Container {cs.name} is waiting: {cs.state.waiting.reason}")
+                    return {"success": False}
+                if not cs.ready:
+                    print(f"❌ Container {cs.name} is not ready")
+                    return {"success": False}
+
+        print("✅ All pods are Running and ready")
+        return {"success": True}
--- a/sregym/conductor/problems/image_pull_wrong_secret.py
+++ b/sregym/conductor/problems/image_pull_wrong_secret.py
@@ -1,0 +1,60 @@
+from sregym.conductor.oracles.image_pull_backoff_mitigation import ImagePullBackoffMitigationOracle
+from sregym.conductor.oracles.llm_as_a_judge.llm_as_a_judge_oracle import LLMAsAJudgeOracle
+from sregym.conductor.problems.base import Problem
+from sregym.generators.fault.inject_virtual import VirtualizationFaultInjector
+from sregym.service.apps.hotel_reservation import HotelReservation
+from sregym.service.kubectl import KubeCtl
+from sregym.utils.decorators import mark_fault_injected
+
+_FAKE_REGISTRY = "fake-private-registry.sregym.internal"
+
+
+class ImagePullWrongSecret(Problem):
+    """Problem that injects incorrect registry credentials causing ImagePullBackOff.
+
+    This simulates a real-world scenario where a deployment is updated to pull
+    from a private container registry but the imagePullSecret contains invalid
+    credentials (or the registry is unreachable), causing the pod to fail image
+    pull with ErrImagePull / ImagePullBackOff.
+    """
+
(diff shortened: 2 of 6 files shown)


sregym/conductor/problems/image_pull_wrong_secret.py
```diff
@@ -0,0 +1,60 @@
+from sregym.conductor.oracles.image_pull_backoff_mitigation import ImagePullBackoffMitigationOracle
+from sregym.conductor.oracles.llm_as_a_judge.llm_as_a_judge_oracle import LLMAsAJudgeOracle
+from sregym.conductor.problems.base import Problem
+from sregym.generators.fault.inject_virtual import VirtualizationFaultInjector
+from sregym.service.apps.hotel_reservation import HotelReservation
+from sregym.service.kubectl import KubeCtl
+from sregym.utils.decorators import mark_fault_injected
+
+_FAKE_REGISTRY = "fake-private-registry.sregym.internal"
```
Saadmrp1038, line 9: This gives the agent hint that the scenario is fake. We should use generic names.

sregym/generators/fault/inject_virtual.py
```diff
@@ -1161,6 +1161,125 @@ def recover_liveness_probe_misconfiguration(self, microservices: list[str]):
...
+
+            print(f"Recovered from ImagePullBackOff fault for service: {service}")
+
+    def inject_wrong_pull_secret(self, microservices: list[str]):
+        """Simulate wrong registry credentials by pointing to a fake private registry with an invalid pull secret."""
+        import base64
+        import json
+
+        fake_registry = "fake-private-registry.sregym.internal"
+        secret_name = "fake-registry-credentials"
```
Saadmrp1038, line 1215: We should use generic names.

sregym/observer/loki/loki-values.yaml
```diff
@@ -90,6 +90,12 @@ monitoring:
 test:
   enabled: false
 
+# Disable cache components to reduce memory usage on local clusters
+chunksCache:
+  enabled: false
+resultsCache:
+  enabled: false
+
```
Saadmrp1038, line 98: I don't think this change is needed. I noticed this before. Both the `chunksCache` and `resutlsCache` requests around ~11GB of memory but they do not use that much. 16GB of ram is a reasonable requiement these days.

sregym/conductor/oracles/image_pull_backoff_mitigation.py
```diff
@@ -0,0 +1,92 @@
+import time
+
+from sregym.conductor.oracles.base import Oracle
+
+_ROLLOUT_SETTLE_SECONDS = 60
+_ROLLOUT_POLL_INTERVAL = 5
+_INVALID_TAG = "nonexistent-tag-xyzzy"
```
Saadmrp1038, line 7: We should use generic names.

Saadmrp1038 (changes requested):
@varuni7 Thanks for the PR! The code looks great. I left some comments (mostly nit). 
But I do have some important design related queries:
1. `image_pull_backoff_hotel_reservation` looks very similar to the existing `incorrect_image` problem from the agent's perspective. Both present as a single service in `ImagePullBackOff` due to a bad image reference, and the fix is to patch the deployment with a valid image. Except the application being different the full diagnosis and mitigation path looks exactly identical to me. Can you please explain if there's anything distinct here? Otherwise I think we can drop this fault.
2. For `image_pull_wrong_secret_hotel_reservation` I deployed the fault. When running `kubectl describe pod`. The actual error shows up as:
```
Failed to pull image "fake-private-registry.sregym.internal/hotelreservation:latest":
dial tcp: lookup fake-private-registry.sregym.internal on 127.0.0.53:53: no such host
```
Since the fake registry domain doesn't exist we get a DNS resolution failure. This is different from the Airwallex incident you cited. That was a broken GKE metadata server (node-level credential mechanism failure after a cluster upgrade). The error there was an authentication error.
Image
For this to be a unique fault, the registry needs to be reachable but reject credentials (return 401). As currently implemented, this problem also somewhat overlaps with `incorrect_image` and the diagnostic signals are similar: read error -> see docker image pull failed (dns error/image does not exist) -> check images of other services -> realize the crashing service's image does not match the rest of them (different registry/wrong name pattern) -> fix image name.

For example a simple correct mitigation command for current implemetation is:
```
 kubectl set image deployment/profile hotel-reserv-profile=yinfangchen/hotelreservation:latest -n hotel-reservation
```

I don't think we can currently faithfully simulate the real Airwallex fault easily.
The real incident was about:
 - GKE metadata server DaemonSet was outdated (version 20200626 vs 20220301)
 - After node upgrade to 1.21+, iptables rules forwarded requests to 169.254.169.252:988
 - But the old metadata server was still listening on 127.0.0.1:988
 - Nodes couldn't obtain GCR tokens resulting in all image pulls failed with auth errors
 - The fix given was to delete the gke-metadata-server DaemonSet so GKE auto-recreates it with the correct version

My overall suggestion would be to pivot to a new fault. 

---

varuni7:
Hi @Saadmrp1038 , thanks for the reviews! 

1. I was not aware of a similar imagepullbackoff existing , we can remove this 

2. yea, I cited  the Airwallex issue mostly because i wanted another way of imagepullbackoff and took inspiration from the fact that underneath infra can fail and also lead to this, it is exactly not the same , my main objective was to have a different way to reach  ImagePullBackOff essentially , reason I cited the incident was because it shows how ImagePullBackOff can surface from an entirely different layer , not from a bad tag or misconfigured deployment, but from infrastructure failing underneath the container runtime. intent was to introduce a distinct root-cause path to the same symptom, an authentication failure at the registry level. From a diagnostic standpoint, these are meaningfully different right,  a wrong tag points you to check the image ref, but a 401 auth error points you to check imagePullSecrets, registry credentials, and whether the secret is actually mounted correctly in the deployment. I was trying to widen the coverage of failure modes that all produce the same surface symptom. The reference I cited is a vage citation in that sense because I got the root idea from there,  because to implement it faithfully I would have to have a GKE cluster with some fake credential-serving registry , it is hard to implement. Irrespective of the incident I mentioned please let me know if this failure mode is a good fit

sregym/conductor/oracles/image_pull_backoff_mitigation.py
varuni7: that makes sense , I did not think from standpoint that this could give the agent hints!

sregym/observer/loki/loki-values.yaml
varuni7: I see! Okay yea I did this mostly because I could not afford this config on my local setup , I forgot to revert this

Saadmrp1038:
@varuni7 Thanks for the response!
1. Sounds good 👍
2. The concept is definitely good . A 401 auth error will be a meaningfully different fault. But the issue is the implementation. Currently, since `fake-private-registry.sregym.internal` doesn't resolve in DNS, the kubelet never reaches the authentication step. The error the agent sees is `no such host` (DNS failure), not `unauthorized` (auth failure). So, the agent just fixes the image reference. Also, even if we somehow get 401 auth error that means the ideal mitigation should be fixing the auth error so that the service can successfully auth and pull the image and not simply update the image reference to point to something else. This is would be quite complex to implement but would make it a genuinely unique fault. So, as it stands the auth error should do two things:
	- The error logs will say `unauthorized`
	- The ideal mitigation would be to fix in such a way that the auth succeeds and not point to some image reference that does not have auth

If you can pull off the auth error implementation that would be really awesome! But that might be quite complex and not be easily doable in our current setup. If you think you can pull it off feel free to try! Or feel free to explore new faults. I am fine with any unique real world fault.

---

Implementing a fault can be tricky. Take your time and feel free to reach out if you need help!

tianyin:
I'm closing this PR as I see @varuni7 has opened a new one #807 

Please feel free to re-open it if you still want to make progress on this.

PR #777, opened 2026-05-24
Add admission webhook TLS mismatch problem

mohamedharake:
Adds a new SREGym problem simulating an admission-webhook TLS trust mismatch in the Hotel Reservation app. The failure is a production-style Kubernetes admission-control issue: the webhook backend is reachable, but the `ValidatingWebhookConfiguration` contains a stale/wrong `caBundle`, so the kube-apiserver rejects pod CREATE requests with an `x509` verification error.

The PR is limited to the new problem plus registry wiring.

---

Admission webhooks sit directly in the Kubernetes API request path. If a `ValidatingWebhookConfiguration` has `failurePolicy: Fail`, then matching requests are rejected whenever the webhook call fails. This is risky because the affected application can be healthy, but Kubernetes may still be unable to create replacement pods.

This problem models a certificate-rotation / CA-drift failure: the webhook backend is reachable and serving HTTPS, but the `caBundle` stored in the `ValidatingWebhookConfiguration` is stale or wrong. The kube-apiserver therefore cannot verify the webhook server certificate and rejects pod CREATE admission requests with an `x509` error.

Several real-world reports anchor this simulation:

- **OPA cert-controller, October 2020** — [“CA and Server certificate potentially get updated before ValidatingWebhookConfiguration”](https://github.com/open-policy-agent/cert-controller/issues/13). The issue describes the risk this problem models: the CA is renewed, but the `ValidatingWebhookConfiguration` may still contain the old CA, so webhook calls can fail.

- **ingress-nginx, August 2020** — [“apply ingress rule error after install ingress-nginx: x509 certificate is not valid”](https://github.com/kubernetes/ingress-nginx/issues/5968). A validating admission webhook rejects resource creation because the serving certificate does not validate for the webhook service DNS name.

- **ingress-nginx, January 2022** — [“Kubernetes: ingress-nginx-controller-admission error, x509 certificate signed by unknown authority”](https://fabianlee.org/2022/01/29/nginx-ingress-nginx-controller-admission-error-x509-certificate-signed-by-unknown-authority/). The writeup shows an admission-controller `x509` failure after reinstalling ingress-nginx, with a workaround that patches the validating webhook `caBundle`.

- **cert-manager, September 2023** — [“Webhook inject-ca-from annotation causes downtime”](https://github.com/cert-manager/cert-manager/issues/6350). The report describes a rotation race where the API server accepts one certificate chain while the webhook server still serves another, causing webhook downtime.

The SREGym scenario keeps the same failure class but makes it reproducible: the backend exists and is reachable, while the trust relationship between the kube-apiserver and webhook is intentionally broken.

---

## 2. How the failure is simulated in SREGym

The new problem class lives at:

`sregym/conductor/problems/admission_webhook_tls_mismatch.py`

and is registered as:

`admission_webhook_tls_mismatch_hotel_reservation`

The fault injection does the following:

1. Creates a temporary `policy-system` namespace.
2. Generates TLS material:
   - one CA that signs the webhook server certificate,
   - a separate wrong CA used as the stale `caBundle`.
3. Deploys a small HTTPS webhook backend as `policy-system/pod-policy-webhook`.
4. Installs a `ValidatingWebhookConfiguration` named `pod-policy.validation.k8s.io`.
5. Scopes the webhook to the target app namespace using `namespaceSelector`.
6. Configures the webhook with:
   - `failurePolicy: Fail`
   - pod `CREATE` admission rules
   - the reachable webhook service
   - the intentionally wrong `caBundle`
7. Deletes one pod from the `recommendation` deployment so the ReplicaSet has to recreate it.

The recreate attempt then hits the webhook and fails during TLS verification.

---

## 3. Problem runtime behaviour

Verified locally through the SREGym CLI and Kubernetes events.

| Phase | Observation |
|---|---|
| App deploy | Hotel Reservation deploys normally before fault injection |
| Backend setup | `policy-system/pod-policy-webhook` is created as a reachable HTTPS service |
| Fault injection | `recommendation` pod is deleted |
| Symptom | `recommendation` ReplicaSet has `desired=1`, `current=0`, `ready=0` |
| Diagnostic signal | ReplicaSet events show `failed calling webhook` with an `x509` certificate error |
| Blast radius | Scoped to `hotel-reservation` through `namespaceSelector` |
| Recovery | Deleting the broken webhook allows the ReplicaSet to recreate the missing pod |

The key event observed locally:

```text
failed calling webhook "pod-policy.validation.k8s.io":
tls: failed to verify certificate:
x509: certificate signed by unknown authority
```

This is the main diagnostic signal an agent should follow. The pod is not failing because of its own image, command, memory, or service config; pod creation is being rejected by admission control.

---

## 4. Mitigation and validation

Valid mitigations:

- delete the broken `ValidatingWebhookConfiguration`,
- patch `failurePolicy` from `Fail` to `Ignore`,
- or repair the webhook trust chain by restoring the correct `caBundle`.

Local mitigation test used the first path:

```bash
kubectl delete validatingwebhookconfiguration pod-policy.validation.k8s.io
```

After that, the ReplicaSet recreated the missing `recommendation` pod and the SREGym mitigation oracle passed.

Local final result:

| Check | Result |
|---|---|
| Problem starts | Passed |
| Fault injection | Passed |
| Runtime symptom | `recommendation` under-replicated |
| Expected event | `x509: certificate signed by unknown authority` observed |
| Mitigation oracle | Passed |
| Diagnosis judge | Skipped locally because `JUDGE_MODEL_ID` was not configured |

---

## 5. Expected agent behaviour

I have not run Stratus end-to-end on this problem yet.

Expected successful agent path:

1. Notice `recommendation` is under-replicated.
2. Inspect the ReplicaSet / events instead of only checking the application pod logs.
3. Find that pod creation is failing at admission time.
4. Identify the cluster-scoped `ValidatingWebhookConfiguration`.
5. Distinguish a TLS trust mismatch from a missing webhook backend.
6. Mitigate by deleting the webhook, changing `failurePolicy`, or repairing `caBundle`.

The intended challenge is that the visible symptom is application-level availability loss, but the root cause is a cluster-scoped admission-control dependency.

---

## Notes for reviewers

- No new Python package dependency is added.
- The webhook backend is created dynamically as a small Python HTTPS server.
- The failure is scoped to the target app namespace using `namespaceSelector`.
- Recovery deletes both the webhook configuration and the temporary `policy-system` namespace.
- Currently only the Hotel Reservation variant is registered; sibling variants for `social_network` or `astronomy_shop` can be added if useful.

### DIFF
--- a/sregym/conductor/problems/admission_webhook_tls_mismatch.py
+++ b/sregym/conductor/problems/admission_webhook_tls_mismatch.py
@@ -1,0 +1,382 @@
+"""Problem: admission webhook TLS trust mismatch blocks pod admission in an app namespace.
+
+This models a production-style Kubernetes failure during webhook certificate
+rotation: a ValidatingWebhookConfiguration with ``failurePolicy: Fail`` points
+at a reachable HTTPS webhook backend, but its ``caBundle`` is stale/wrong.
+The kube-apiserver therefore cannot verify the webhook server's TLS certificate
+and rejects pod CREATE admission requests.
+
+In SREGym we scope the failure to a single application namespace via
+``namespaceSelector`` and then delete one pod of a single-replica deployment
+(e.g. ``recommendation`` in hotel-reservation). The ReplicaSet's recreate
+attempt hits the webhook and is rejected with a TLS/x509 admission error, so the
+deployment stays under-replicated even though its spec, image, service, and
+resources are healthy.
+
+Valid mitigations include deleting the webhook config, changing
+``failurePolicy`` to ``Ignore``, or repairing the webhook TLS trust chain by
+restoring a valid ``caBundle``.
+"""
+
+import base64
+import subprocess
+import tempfile
+import textwrap
+import time
+from pathlib import Path
+
+from kubernetes import client
+from kubernetes.client.exceptions import ApiException
+
+from sregym.conductor.oracles.deployment_readiness import DeploymentReadinessOracle
+from sregym.conductor.oracles.llm_as_a_judge.llm_as_a_judge_oracle import LLMAsAJudgeOracle
+from sregym.conductor.problems.base import Problem
+from sregym.service.apps.astronomy_shop import AstronomyShop
+from sregym.service.apps.hotel_reservation import HotelReservation
+from sregym.service.apps.social_network import SocialNetwork
+from sregym.service.kubectl import KubeCtl
+from sregym.utils.decorators import mark_fault_injected
+
+
+class AdmissionWebhookTLSMismatch(Problem):
+    """Inject a reachable admission webhook backend but configure the
+    ValidatingWebhookConfiguration with a stale/wrong caBundle."""
+
+    APPS = {
+        "hotel_reservation": HotelReservation,
+        "social_network": SocialNetwork,
+        "astronomy_shop": AstronomyShop,
+    }
+
+    WEBHOOK_NAME = "pod-policy.validation.k8s.io"
+    BACKEND_SVC_NAME = "pod-policy-webhook"
+    BACKEND_SVC_NAMESPACE = "policy-system"
+    BACKEND_DEPLOYMENT_NAME = "pod-policy-webhook"
+
+    def __init__(self, app_name: str = "hotel_reservation", faulty_service: str = "recommendation"):
+        if app_name not in self.APPS:
+            raise ValueError(f"Unsupported app name: {app_name}")
+
+        self.app_name = app_name
+        self.faulty_service = faulty_service
+        self.app = self.APPS[app_name]()
+        super().__init__(app=self.app, namespace=self.app.namespace)
+
+        self.namespace = self.app.namespace
+        self.kubectl = KubeCtl()
+        self.admission_api = client.AdmissionregistrationV1Api()
+        self.core_api = client.CoreV1Api()
+        self.wrong_ca_bundle = None
+
+        self.root_cause = self.build_structured_root_cause(
+            component=f"ValidatingWebhookConfiguration/{self.WEBHOOK_NAME}",
+            namespace=self.namespace,
+            description=(
+                f"A cluster-scoped ValidatingWebhookConfiguration named `{self.WEBHOOK_NAME}` has been installed "
+                f"with `failurePolicy: Fail` and a `namespaceSelector` scoped to the `{self.namespace}` namespace. "
+                f"The webhook intercepts pod CREATE operations and points to the reachable HTTPS service "
+                f"`{self.BACKEND_SVC_NAMESPACE}/{self.BACKEND_SVC_NAME}`, but the webhook configuration contains "
+                "a stale/wrong `caBundle`. As a result, the kube-apiserver cannot verify the webhook server's TLS "
+                "certificate and rejects pod creation with a `failed calling webhook` / `x509` certificate error. "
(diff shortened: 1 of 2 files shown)


mohamedharake:
Hi, I opened this PR for the evaluation task.

It adds a new SREGym problem that simulates an admission webhook TLS trust mismatch caused by a stale/wrong caBundle. I tested it locally: the fault produces a `failed calling webhook` / `x509: certificate signed by unknown authority` error, and mitigation succeeds by deleting the broken ValidatingWebhookConfiguration.

I’d appreciate any feedback.

Saadmrp1038:
@mohamedharake Thanks for the PR! Could you please write a more detailed/structured PR description? It would help me greatly in reviewing it. Some points you can include:

mohamedharake:
Thank you so much for the feedback. I updated the PR description with the real world failure story, simulation details, runtime behavior, mitigation/testing notes, and expected agent behavior. I’ll also try running Stratus now and add the results if I can get it working locally.

Saadmrp1038:
@mohamedharake Thank you! There are quite a few PRs I have to review. I'll try to get to this one as soon as I can.

sregym/conductor/problems/admission_webhook_tls_mismatch.py
```diff
@@ -0,0 +1,382 @@
...
+            ], stdout=subprocess.DEVNULL)
+
+            self._run(["openssl", "genrsa", "-out", str(wrong_ca_key), "2048"], stdout=subprocess.DEVNULL)
+            self._run([
+                "openssl", "req", "-x509", "-new", "-nodes",
+                "-key", str(wrong_ca_key),
+                "-sha256", "-days", "365",
+                "-subj", "/CN=sregym-stale-wrong-ca",
+                "-out", str(wrong_ca_crt),
+            ], stdout=subprocess.DEVNULL)
```
Saadmrp1038, line 124: nit: Let's use generic/realistic names here for `/CN`. I don't like to use names that can reveal internal information to the agent.
For example, when running with claudecode it read this data:
```
The root cause is clear now:\n\n1. The CA  bundle in the `ValidatingWebhookConfiguration` (`pod-policy.validation.k8s.io`) contains a CA with CN = `sregym-stale-wrong-ca`\n2. The server certificate in the `pod-policy-webhook-tls` secret was signed by a CA  with CN = `sregym-real-webhook-ca`
```

Saadmrp1038 (changes requested):
@mohamedharake This is a great PR! Very clean implementation. I tested the fault and everything works nicely.
There's just this one small nit. Ping me when it's done. I will merge the PR!

Saadmrp1038:
On a side note, I ran this problem with multiple agents. The mitigation by Claude Code was very interesting. Instead of taking a shortcut (deleting the webhook), it successfully repaired the trust chain by generating a new CA, signing a new server cert, and patching both the TLS secret and the webhook's caBundle to match.

mohamedharake:
Done, thanks a lot! I renamed the certificate common names to more generic/realistic values.
Also that Claude Code mitigation is really interesting. Thank you for testing it with multiple agents. I really appreciate the review and feedback.

Saadmrp1038 (approved):

Saadmrp1038:
@mohamedharake Awesome! Congrats on your first PR merged here 🎉
Keep up the good work!

tianyin:
This is great work!!

mohamedharake:
Thank you!!

PR #780, opened 2026-05-25
Revert kubelet crash injection method from kubectl debug to SSH

Saadmrp1038:
Reverts the kubelet crash fault injection method from `kubectl debug node` back to SSH (paramiko).

`kubectl debug node` introduced in two issues:

1. `--rm` flag doesn't exist, causing silent injection failure. Added error log to `kubectl.py` to detect these errors in the future.
2. Recovery is impossible. Can't schedule debug pods on nodes with dead kubelet so `recover_inject` won't work.

Since recovery is impossible in this way, we are reverting to previous ssh approch.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,6 +98,7 @@ dependencies = [
     "opentelemetry-semantic-conventions==0.52b1",
     "opentelemetry-util-http==0.52b1",
     "packaging>=24.0",
+    "paramiko>=5.0.0",
     "pandas==2.2.3",
     "pathspec==0.12.1",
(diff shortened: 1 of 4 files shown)


yimingsu01:
I actually forgot why we switched away from SSH. Why did we switch?

Saadmrp1038:
We didn't want to read `inventory.yml` and directly SSH into the nodes.

Saadmrp1038:
@HacksonClark @yimingsu01 can anyone approve this please if no more changes needed? :)

tianyin (approved):

PR #783, opened 2026-05-25
Add CPU-throttling brownout problem

TheDeadcoder:
# Add CPU-throttling brownout problem (first gray-failure fault + performance oracle)
## Context

Every fault in SREGym is fail-stop: a pod crashes, fails to schedule, or returns errors, and the default `MitigationOracle` treats success as every pod reaching `Running`. But many production incidents are gray failures, or brownouts, where the system stays up but runs degraded. This PR adds this brownout problem to SREGym, based on a well documented real-world failure: latency degradation from Linux CFS CPU-quota throttling under Kubernetes CPU limits ([[kubernetes/kubernetes#67577](https://github.com/kubernetes/kubernetes/issues/67577)](https://github.com/kubernetes/kubernetes/issues/67577)). When a container reaches its CPU limit it is throttled within each 100ms period, so latency rises while pods stay `Running` and average CPU usage looks low. Detecting this needs a new fault category and a performance-based oracle, because liveness checks are blind to a brownout. The problem reuses the existing [[Hotel Reservation application](https://github.com/SREGym/SREGym-applications)](https://github.com/SREGym/SREGym-applications).

## Code changes

New files:

- `sregym/generators/fault/inject_resource_contention.py`: `ResourceContentionFaultInjector`, a new fault category for resource contention. `inject_cpu_throttling` lowers a deployment's CPU request and limit together with `kubectl set resources`, then reads the limit back and raises if it did not apply. The request and limit are moved together because a limit below the existing request is rejected by the Kubernetes API. `recover_cpu_throttling` restores the original request and limit.
- `sregym/conductor/problems/cpu_throttling_brownout.py`: `CpuThrottlingBrownout`, parameterized by app and service, default limit `50m`. Diagnosis is graded by the existing `LLMAsAJudgeOracle`. Mitigation is graded by a `CompoundedOracle`.
- `sregym/conductor/oracles/cpu_throttling_mitigation.py`: `CpuThrottlingRatioOracle`, a performance-based mitigation oracle. It reads the deployment spec first, and if the CPU limit is gone it passes immediately. Otherwise it queries the CFS throttling ratio from Prometheus and passes when it stays below `0.10`.
- `tests/test_cpu_throttling_brownout.py`: cluster-free unit test for the structured root cause.

Modified:

- `sregym/conductor/problems/registry.py`: one import and three problem IDs: `cpu_throttling_brownout_hotel_reservation`, `cpu_throttling_brownout_social_network`, `cpu_throttling_brownout_astronomy_shop`.

To run the problem, start it from the SREGym CLI:

```
SREGym> start cpu_throttling_brownout_hotel_reservation
```

Wait time: the diagnostic query uses a 2-minute Prometheus rate window, so wait about 3 minutes after the problem reaches the diagnosis stage before running throttling queries. Running them earlier returns an empty result, because there is not yet enough data in the window.

Throttling ratio query used during investigation:

```
kubectl exec -n observe deploy/prometheus-server -c prometheus-server -- wget -qO- 'http://localhost:9090/api/v1/query?query=sum(rate(container_cpu_cfs_throttled_periods_total%7Bnamespace%3D%22hotel-reservation%22%2Cpod%3D~%22%5Efrontend-.%2A%22%7D%5B2m%5D))%2Fsum(rate(container_cpu_cfs_periods_total%7Bnamespace%3D%22hotel-reservation%22%2Cpod%3D~%22%5Efrontend-.%2A%22%7D%5B2m%5D))'
```

## Outcomes

Validated on a kind cluster with `cpu_throttling_brownout_hotel_reservation`. After injection:

- `frontend` has `limits.cpu: 50m` and `requests.cpu: 25m`.
- Every pod in the `hotel-reservation` namespace stays `Running`. This is the brownout property: degraded, not down.
- `kubectl top` shows `frontend` at about 39m, under its 50m limit. This is the misdirection. CPU usage looks fine because the container is throttled below its limit.
- The throttling ratio query returns about `0.685`, meaning `frontend` is throttled in roughly 68% of scheduling periods. This is a clear signal and separates well from the `0` ratio seen when no fault is active.

The injector prints a confirmation line on success and raises on failure, so a rejected injection fails loudly instead of reaching the diagnosis stage with no fault active.

## Mitigation

The fix is to raise or remove the CPU limit on the `frontend` deployment, for example:

```
kubectl set resources deployment/frontend -n hotel-reservation --limits=cpu=2
```

The mitigation oracle is a `CompoundedOracle` of `CpuThrottlingRatioOracle` and the stock `MitigationOracle`. The throttling oracle checks the deployment spec first, so removing the limit passes right away. If a limit is still present, it samples the throttling ratio and passes when the ratio stays below `0.10`. The `MitigationOracle` runs alongside as a liveness sanity check, so the agent cannot pass by deleting the service. The oracle waits 120 seconds for post-fix metrics to settle and then takes three samples, so the verdict takes about 3 minutes after submission.

### DIFF
--- a/sregym/conductor/oracles/cpu_throttling_mitigation.py
+++ b/sregym/conductor/oracles/cpu_throttling_mitigation.py
@@ -1,0 +1,115 @@
+"""Mitigation oracle for CPU-throttling brownout problems."""
+
+import json
+import subprocess
+import time
+import urllib.parse
+
+from sregym.conductor.oracles.base import Oracle
+_PROMETHEUS_URL = "http://localhost:9090"
+
+_THROTTLE_THRESHOLD = 0.10
+_SETTLE_BUFFER_SECONDS = 120
+_SAMPLE_COUNT = 3
+_SAMPLE_INTERVAL_SECONDS = 15
+
+
+class CpuThrottlingRatioOracle(Oracle):
+    """Passes when the faulty service is no longer CFS-throttled."""
+
+    importance = 2.0
+
+    def __init__(self, problem, faulty_service: str, threshold: float = _THROTTLE_THRESHOLD):
+        super().__init__(problem)
+        self.faulty_service = faulty_service
+        self.threshold = threshold
+
+    def _get_cpu_limit(self) -> str | None:
+        """Return the deployment's container[0] cpu limit, "" if none, None on error."""
+        ns = self.problem.namespace
+        cmd = (
+            f"kubectl get deployment {self.faulty_service} -n {ns} "
+            '-o jsonpath="{.spec.template.spec.containers[0].resources.limits.cpu}"'
+        )
+        try:
+            out = self.problem.kubectl.exec_command(cmd)
+        except Exception as exc:
+            print(f"⚠️  Failed to read cpu limit from deployment spec: {exc}")
+            return None
+        return (out or "").strip().strip("'\"")
+
+    def _promql(self) -> str:
+        ns = self.problem.namespace
+        selector = f'namespace="{ns}",pod=~"^{self.faulty_service}-.*",container!="",container!="POD"'
+        return (
+            f"sum(rate(container_cpu_cfs_throttled_periods_total{{{selector}}}[2m]))"
+            f" / "
+            f"sum(rate(container_cpu_cfs_periods_total{{{selector}}}[2m]))"
+        )
+
+    def _query_ratio(self) -> float | None:
+        url = f"{_PROMETHEUS_URL}/api/v1/query?query={urllib.parse.quote(self._promql())}"
+        cmd = [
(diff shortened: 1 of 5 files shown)


yimingsu01:
@Saadmrp1038 will review this PR in detail, but I would like to point out that not all faults in SREGym is fail-stop: we have metastable failure problems too, which is equipped with latency-based oracles which you claim we miss.

tianyin:
@TheDeadcoder I think @yimingsu01 meant that you should change the description (to correct the factual issue). He is not suggesting that you should close the issue.

TheDeadcoder:
No, I got it. After reviewing metastable failure problems part, I found very little novelty for this PR.

Issue #763, opened 2026-05-20
[Problem RFC] Expired TLS/SSL Certificate


gAngelov24:
Hi all, I am working on the evaluation task and would like to propose a new problem for the benchmark.

What immediately came to mind was the issue [Spotify](https://www.thousandeyes.com/blog/impacts-expired-tls-certificate) had a few years ago when their TLS certificate expired. These kind of stories are actually pretty common and cause a large amount of cloud outages (e.g. [Microsoft Teams](https://trackssl.com/microsoft-teams-ssl-expiry/), [Epic Games](https://www.keyfactor.com/blog/the-enemy-of-uptime-an-expired-ssl-certificate/), ...)

**How I plan to simulate the failure on SREGym**
1. Generate an intentionally expired TLS certificate
2. Create a Kubernetes `Secret` if type `tls` which contains the expired certificate
3. Modify and applications `Ingress` resource to terminate HTTPS traffic using the invalid `Secret`

**The problems expected runtime behavior**
In order for the agent to find this issue, it must look beyond the pods that `kubectl get pods` will show since those would all be in a `Running` and healthy state. The agent would need to test the external endpoint so that it can observe the SSL handshake failure. It could then trace the routing to the Ingress controller, and decode the TLS secret to diagnose the root cause.

Let me know if this sounds like a viable addition to the benchmark!

P.S.
I am pretty new to this stuff so if anything I said doesn't make sense or needs more thought please let me know!

HacksonClark:
That sounds like a great idea! Would be a welcome addition to the benchmark.

tianyin:
> I am pretty new to this stuff so if anything I said doesn't make sense or needs more thought please let me know!

It makes a lot of sense, and the writing of the proposal shows your maturity.

Saadmrp1038:
@gAngelov24 your idea sounds good! We have no fault like this yet. There is another TLS related fault `auth_miss_mongodb` that breaks the TLS at the internal service-to-database layer. Yours is at the ingress layer. 
Send a PR whenever it's ready. Waiting to see the fault in action!

tianyin:
I'm closing it as #784 is out. Thank you for the RFC. It's very helpful for us to understand the problem!

PR #484, opened 2026-01-17
Add `ruff` as formatter and linter

Saadmrp1038:
## Summary
Replaces Black and isort with Ruff for faster, unified linting and formatting.

## Changes
- **[.pre-commit-config.yaml](cci:7://file:///home/pial/Academics/UIUC/Projects/SREGym/.pre-commit-config.yaml:0:0-0:0)**: Replaced Black + isort hooks with `ruff-check` and `ruff-format`
- **[pyproject.toml](cci:7://file:///home/pial/Academics/UIUC/Projects/SREGym/pyproject.toml:0:0-0:0)**: Replaced `[tool.black]` and `[tool.isort]` with `[tool.ruff]` configuration

## Why Ruff?
Previously, we used `black` for formatting and `isort` for import sorting. While this kept code styled consistently, we had no linting meaning stuff like undefined variables, unused imports, or deprecated syntax patterns could slip through without warning.

Ruff is a single tool that replaces both `black` and `isort` while also adding linting capabilities we didn't have before. It catches:

- Undefined variables and unused imports (F/pyflakes rules)
- Common bugs and code smells (B/flake8-bugbear rules)
- Outdated Python patterns that can be modernized (UP/pyupgrade rules)
- Unnecessarily complex code that can be simplified (SIM/flake8-simplify rules)
- Unsorted or improperly grouped imports (I/isort rules)
- Style issues like incorrect whitespace, indentation, and line formatting (E/pycodestyle rules)

There are more rules [here](https://docs.astral.sh/ruff/rules/) if we want anything else.

### Required
Run `uv run pre-commit clean && uv run pre-commit install` to install the pre-commit hooks. You only need to do this once. Without this, the new hooks won't run when you commit.

### Recommended (for IDE integration)
The following steps are optional but recommended for a better developer experience:

1. Install the **Ruff** formatter from the [VS Code extension marketplace](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff). This shows lint errors directly in the IDE as you type.

2. Make Ruff your default formatter by opening your user `settings.json` (type `Open User Settings (JSON)` in the command palette) and adding:

```json
"[python]": {
    "editor.defaultFormatter": "charliermarsh.ruff",
    "editor.formatOnSave": true
},
"editor.codeActionsOnSave": {
    "source.fixAll": "explicit"
},
"editor.formatOnSaveMode": "modificationsIfAvailable"
```

This will show lint errors in the IDE and auto-fix them when saving (Ctrl/Cmd+S).

[...]

PR #784, opened 2026-05-26
Add expired_tls_hotel_reservation problem

gAngelov24:
## Real-World Failure
An expired TLS certificates are an extremely common cause of system outages. Unlike most failures, they are completely predictable since every certificate has a know expiry date; however, they still manage to take down major services. I previously mentioned in issue #763 some large cases where this has happened.

## How I Simulated The Failure
The fault is injected into the Hotel Reservation app on its `hotel-reservation` namespace. The simulation required two additional Kubernetes resources:
1. The Expired Certificate
    - The fault injector generates a new certificate at injection time using Python's `cryptography` library. At first I thought to use OpenSSL to generate the certificate however that library is strict about the date you set and wouldn't allow me to create an already expire certificate.
    - I also initially thought to write the certificate to the disk to avoid complex base64 string encoding, but I changed my implementation to base64 simply to keep everything local.
2. Frontend Ingress
    - I added a new Kubernetes Ingress resource (`frontend-ingress`) which is applied to the namespace. It references the expired secret via the standard Kubernetes TLS Ingress configuration.

Like I stated above, I used Pythons `cryptography` package to generate the expired x509 certificate, and to parse the `tls.crt` PEM bytes from the Kubernetes Secret in order to read the `not_valid_after_utc`.
In order to use it I ran:
```bash
uv add cryptography
```
If not installed, both the `inject_fault` and the oracle's `_get_cert_expiry` methods log an `ImportError` with the fix command rather than producing a confusing traceback.

## Problem Runtime
After the `inject_fault()` runs, the cluster has the following observable state:
**Kubernetes resources:**
```
kubectl get ingress -n hotel-reservation
NAME               CLASS   HOSTS         ADDRESS   PORTS     AGE
frontend-ingress   nginx   hotel.local             80, 443   Xs
 
kubectl describe ingress frontend-ingress -n hotel-reservation
# Shows: TLS terminates with secret "expired-frontend-cert"
# Annotation: nginx.ingress.kubernetes.io/ssl-redirect: "true"
 
kubectl get secret expired-frontend-cert -n hotel-reservation
NAME                    TYPE                DATA   AGE
expired-frontend-cert   kubernetes.io/tls   2      Xs
```

**Certificate:**
```
kubectl get secret expired-frontend-cert -n hotel-reservation \
  -o jsonpath='{.data.tls\.crt}' | base64 -d | openssl x509 -noout -dates
notBefore=<10 days ago>
notAfter=<1 day ago>
```

**Pod health:**
```
# All pods show Running/Ready
```

**HTTPS Behaviour:**
- Requests to `https://hotel.local` result in a failed TLS handshake
- the NGINX Ingress controller logs chow certificate validation errors
- HTTP requests are redirected to HTTPS (because of `ssl-redirect: "true"` in `frontend-ingress`) so there are no unencrypted fallback paths. Any path will be affected.

The challenge for an AI agent is that at first glance everything seems to be working:
1. `kubectl get pods -A` shows all green
2. Application logs show no errors at the app layer since TLS is terminated at the Ingress.
3. Service/Deployment configs don't show anything abnormal
<br>
In order to find the underlying issue, the agent must specifically inspect the Ingress TLS configuration then cross reference the secret, and finally decode and inspect the certificate inside the secret.

### Oracle Behavior
Before I get into Agent behavior, I want to talk about the new Oracle I had to design in order for this failure to work as expected. The default `MitigationOracle` checks pod health, which in this case is always fine since the expired TLS certificate does not affect pod health. 
Therefore I made `ExpiredTlsMitigationOracle` to inspect the frontend Ingress at submission time and checks the expiry of any TLS secret it references. It will accept any of the following as valid mitigations: deleting the Ingress, deleting the secret, removing the TLS block from the Ingress, or replacing the secret with a valid certificate.
It will reject the submission if any referenced secret still contains an expired certificate.
I tested it in both cases:
- `submit()` with failure active: `success: false`, `"Secret 'expired-frontend-cert' still contains a certificate that expired on <timestamp>"`
- `submit()` after removing the Ingress and secret: `success: true`, `"frontend-ingress has been deleted — mitigation accepted"`

### Human Agent (me):
While I may have had some insider information about the problem, I was able to diagnose the issue and resolve it. I ran the following command to diagnose the failure:
```bash
kubectl get ingress -n hotel-reservation
```
This showed me the `frontend-ingress` referencing the `expired-frontend-cert` you could then inspect the cert to verify it is expired. Then I ran:
```bash
kubectl delete ingress frontend-ingress -n hotel-reservation
kubectl delete secret expired-frontend-cert -n hotel-reservation
```
This way I deleted the Ingress and secret to mitigate the failure.

### AI Agent (Stratus, `gemini/gemini-3.1-flash-lite`)
The AI agent failed both stages across multiple attempts. During diagnosis, it checked the Jaeger for telemtry traces, found none, and submitted an incorrect root cause about services not sending data to Jaeger. Then, during mitigation, the agent read the frontend logs and found `connection refused` errors from the frontend trying to reach Consul. These errors are normal and go away on their own. It spent all 20 steps investing this error and never ran `kubectl get ingress -n hotel-reservation`
Note: Testing was also attempted with gemini/gemini-2.5-flash but hit the free-tier rate limit (5 RPM, 20 RPD). gemini/gemini-3.1-flash-lite was used for all completed runs.

## Additional Change
Outside of the actual problem, I noticed that whenever I pressed enter without a prompt in the CLI, it would crash. I added a check that way anytime you press enter, whether on accident or not, the CLI doesn't just crash. If that was a feature and not a bug let me know I will remove that addition.

/validate-problem expired_tls_hotel_reservation


Saadmrp1038:
@gAngelov24 Thanks for the PR! I will try to review it by tomorrow.

> If that was a feature and not a bug let me know I will remove that addition.

Definitely a bug. Good catch!

sregym/conductor/problems/manifests/frontend_ingress.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: frontend-ingress
+  namespace: hotel-reservation
+  annotations:
+    nginx.ingress.kubernetes.io/ssl-redirect: "true"
+spec:
+  tls:
+  - hosts:
+    - hotel.local
+    secretName: expired-frontend-cert
```
Saadmrp1038: We should use generic/realistic name here.

sregym/conductor/problems/expired_tls_hotel_reservation.py
```diff
@@ -0,0 +1,141 @@
...
+    inside the referenced secret to discover the root cause.
+    """
+    def __init__(self):
+        self.app = HotelReservation()
+        # app must be set before super().__init__() so the base class can read self.app.namespace
+        super().__init__(app=self.app, namespace=self.app.namespace)
+
+        self.kubectl = KubeCtl()
+        self.problem_id = "expired_tls_hotel_reservation"
+        self.secret_name = "expired-frontend-cert"
```
Saadmrp1038: We should use generic/realistic name here.

sregym/conductor/oracles/expired_tls_mitigation_oracle.py
```diff
@@ -0,0 +1,107 @@
+import datetime
+import logging
+
+from kubernetes import client
+
+logger = logging.getLogger(__name__)
+
+
+class ExpiredTlsMitigationOracle:
```
Saadmrp1038: This class should inherit from base `Oracle` class. Check the other oracles for reference.

sregym/conductor/problems/expired_tls_hotel_reservation.py
```diff
@@ -0,0 +1,141 @@
...
+            component="frontend-ingress",
+            namespace=self.namespace,
+            description=(
+                "The frontend Ingress resource is configured with a TLS secret named "
+                f"`{self.secret_name}` which contains an expired TLS certificate. "
+                "This breaks HTTPS connections."
+            ),
+        )
+        self.diagnosis_oracle = LLMAsAJudgeOracle(problem=self, expected=self.root_cause)
+        self.mitigation_oracle = ExpiredTlsMitigationOracle(problem=self)
```
Saadmrp1038, line 55: `self.app.create_workload()` is missing.

sregym/conductor/problems/expired_tls_hotel_reservation.py
Saadmrp1038, line 1: nit: We have ruff setup for this repo. You can check #484 for how to set it up locally. Will help catch simple lint errors and import ordering issues.

sregym/conductor/problems/expired_tls_hotel_reservation.py
```diff
@@ -0,0 +1,141 @@
...
+    def inject_fault(self) -> bool:
+        logger.info("Injecting Expired TLS Certificate fault...")
+
+        cert_pem, key_pem = self._generate_expired_cert()
+        self._create_tls_secret(cert_pem, key_pem)
+
+        KubeCtl().apply_configs(self.namespace, self.ingress_yaml_path)
+
+        logger.info("Injected expired TLS cert into Ingress.")
+        return True
```
Saadmrp1038, line 82: I think the fault is inert now. The `inject_fault` is not deploying the NGINX Ingress controller. That means no controller is processing the ingress. The expired cert exists but nothing is trying to use it, so there is no observable TLS failure. Check `ingress_misroute.py` for reference. `IngressNginx().deploy()` is needed to deploy the nginx ingress controller.

sregym/conductor/problems/manifests/frontend_ingress.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: Ingress
+metadata:
+  name: frontend-ingress
+  namespace: hotel-reservation
+  annotations:
+    nginx.ingress.kubernetes.io/ssl-redirect: "true"
+spec:
+  tls:
+  - hosts:
+    - hotel.local
```
Saadmrp1038: We should add `ingressClassName: nginx` so that ingress gets picked up by controller reliably. Check `ingress_misroute.py` for reference.

Saadmrp1038 (changes requested):
@gAngelov24 Thanks for the effort! I left some comments below. 

I am not sure if this fault is working correctly right now. The major issue I found is that `IngressNginx().deploy()` is missing from `inject_fault`.

> the NGINX Ingress controller logs chow certificate validation errors

What controller did you mean by this? I don't think one is getting deployed currently. There is no `ingress-nginx` namespace or any Ingress controller running in the cluster.

> Requests to https://hotel.local result in a failed TLS handshake

I wasn't able to reproduce this either. `hotel.local` doesn't resolve, and without a running Ingress controller, nothing is serving TLS on that hostname.

Can you please share the commands you used to test these claims and the logs they're based on? Not sure if I am missing something here.

---

sregym/conductor/problems/expired_tls_hotel_reservation.py
gAngel28: oh you are right, I spent so much time trying to get the oracle right, I forgot to actually test that TLS handshakes were actually failing :( I'll start working on the changes right now and try and have them done by tonight.

sregym/conductor/problems/manifests/frontend_ingress.yaml
gAngel28: will do thanks for the comments!

sregym/conductor/problems/expired_tls_hotel_reservation.py
Saadmrp1038: @gAngelov24 no rush! Take your time.

tianyin: @gAngel28 Please make sure that you test it well and make it solid. It takes a lot of time to review and test ourselves, and a bit underwhelming to see untested code.

gAngelov24:
Okay, I addressed all the issues you had commented on. I changed variable names to be more realistic, I made the oracle inherit from the base oracle, and I fixed the fault so that NGINX controller reads the expired certificate. 
I ran these commands:
Terminal 1:
```
uv run python cli.py
SREGym> start expired_tls_hotel_reservation
```
Terminal 2:
```
openssl s_client -connect localhost:8443 -servername hotel-reservation.local 2>&1 | openssl x509 -noout -dates -subject
kubectl logs -n ingress-nginx deployment/ingress-nginx-controller
```

Now, you should be able to see that the certificate being used by the NGINX is expired. The openssl command should output something like this:
```
notBefore=May 19 16:54:30 2026 GMT
notAfter=May 28 16:54:30 2026 GMT
subject=CN = hotel-reservation.local
```
The notAfter date in the past proves the expired cert is being served to clients rather than the NGINX default fallback.
You can also confirm the controller detects it with:
`kubectl logs -n ingress-nginx deployment/ingress-nginx-controller` 
And look for:
`SSL certificate for server "hotel-reservation.local" expired (2026-05-28 ...)`
`TLS handshake error from 10.244.0.1:15450: remote error: tls: bad certificate`

Saadmrp1038 (approved):

Saadmrp1038:
@gAngelov24 Thanks for addressing my comments! The fault is working correctly now.
Congrats on completing the eval task 🎉

---

Also, @gAngelov24 and @gAngel28 I got confused for a moment XD

tianyin:
WoW!! This is awesome! Congrats!!

Saadmrp1038:
On a side note, agents really suffer in this problem. They get lost in the app namespace chasing random stuff and never checks the `ingress-nginx-controller`. I had to tweak the prompt to get the agent to notice the TLS error logs in the `ingress-nginx` namespace. But once it noticed the error it quickly fixed it (generated new valid cert). It does reveal a major flaw in our current agentic flow. The agents don't systematically explore outside the app namespace even when "everything looks healthy" in the app-namesapce and ideally it should broaden the search.

tianyin:
@Saadmrp1038 I feel you have so much materials to write another research paper :))

> It does reveal a major flaw in our current agentic flow. The agents don't systematically explore outside the app namespace even when "everything looks healthy" in the app-namesapce and ideally it should broaden the search.

Bingo!

PR #786, opened 2026-05-26
Fix/wsl stratus connection refused

### LINKED ISSUE #785: Connection refused when running Stratus on WSL
When running Stratus on WSL, the agent container fails immediately with:
`HTTPConnectionPool(host='0.0.0.0', port=8000): Connection refused`

the command ran that gave this error was:
`uv run python main.py --agent stratus --model gemini/gemini-3.1-flash-lite --judge-model gemini/gemini-3.1-flash-lite --problem expired_tls_hotel_reservation`

gAngelov24:
## Fix: Stratus agent `Connection refused` on WSL/Linux

`API_HOSTNAME` was being forwarded into the agent container as `0.0.0.0`, which is a valid bind address but not a valid connect address. On WSL this causes the agent to fail immediately with `Connection refused` on port 8000.

Added an `elif` for Linux in `_build_env_flags` mirroring the existing macOS block, and added `--add-host=host.docker.internal:host-gateway` to the docker run args so the hostname resolves correctly inside the container.

Fixes #785

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -68,6 +68,8 @@ class HumanAgent:
         while True:
             inp = await self._prompt()
             cmd = inp.strip().split(maxsplit=1)
+            if not cmd:
+                continue
             if cmd[0].lower() == "exit":
                 sys.exit(0)
             if cmd[0].lower() == "options":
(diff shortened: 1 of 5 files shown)


Saadmrp1038:
@gAngelov24 create a branch from main and make the changes there. Otherwise this is also including the chagnes from your new fault too :)
You can also just revert the new fault related changes here whatever is easier for you.

gAngelov24:
> @gAngelov24 create a branch from main and make the changes there. Otherwise this is also including the chagnes from your new fault too :)

whoops sorry, I'll fix it

PR #790, opened 2026-05-26
A Kafka poison-pill head-of-line block problem

TheDeadcoder:
## Summary
This PR adds `kafka_poison_pill_hol_block`, which is a Kafka poison-pill head-of-line block problem

A common production incident for Kafka-based microservice platforms is poison pill. It is a single malformed record that a consumer cannot deserialize. Because order is preserved within a partition, that one record blocks every valid message behind it. The offset is never committed. Meanwhile the lag grows, and yet every pod still reports healthy. Real outages have been attributed to this, including a malformed record that blocked a payment pipeline for around 45 minutes. So far, I could not find fault of this kind in SREGym. A poison pill lives in the Kafka log, outside the control plane, so a restarted consumer replays the bad record. This PR adds that scenario.

Relevant References:
[Conduktor - Dead Letter Topics: Handling Poison Pills](https://www.conduktor.io/blog/dead-letter-topics-handling-poison-pills)
[Redpanda - Reliable message processing with a dead letter queue](https://www.redpanda.com/blog/reliable-message-processing-with-dead-letter-queue)
[Lydtech Consulting - Kafka Poison Pill](https://www.lydtechconsulting.com/blog-kafka-poison-pill.html)

## Code changes

3 files are added and one is edited:
- `sregym/generators/fault/inject_kafka.py` - fault injector. A self-contained producer and consumer are deployed into the `astronomy-shop` namespace, and one poison record is seeded into a dedicated topic at offset 20.
- `sregym/conductor/oracles/data_plane_progress.py` - `DataPlaneProgressOracle`, the mitigation oracle. Data-plane progress is graded from the consumer's logs.
- `sregym/conductor/problems/kafka_poison_pill_hol_block.py` - the problem definition.
- `sregym/conductor/problems/registry.py` - one import and one registry entry (see `registry.patch.md`).

After the problem starts and the diagnosis stage opens, an interval of about 1 to 2 minutes should be allowed before investigation begins, so the stalled consumer and the growing lag are clearly visible.

## Validation
``` bash
uv run python tests/integration/validate_problem.py --problem kafka_poison_pill_hol_block
```

## Expected behaviour

While the fault is active, the consumer logs repeat `unprocessable record at offset=20`, the committed offset stays at 20, and lag grows, while every pod remains `Running` and `Ready`. Diagnosis is scored by the existing LLM judge. The mitigation oracle prints its checks and ends with either a pass or a clear failure reason (`offset not advanced past poison record`, `data loss: valid records skipped`, or `not restart-resistant`). Results are written to the `cli.py` results panel, or to the run CSV when `main.py` is used. Inspection commands used here are:
```bash
kubectl logs deployment/orders-validator -n astronomy-shop --tail=20  # ... unprocessable record at offset=20 ...
kubectl get pods -n astronomy-shop -l app=orders-validator # 1 pod running
kubectl logs deployment/order-stream  -n astronomy-shop --tail=5   # producer still streaming
```

## Mitigation

The expected fix is to move the consumer group past the poison record without dropping the valid messages queued behind it. Since an offset reset requires the group to have no active members, the consumer is stopped first, the group offset is reset to 21 (one past the poison), and the consumer is started again:

```bash
kubectl scale deployment/orders-validator -n astronomy-shop --replicas=0
# now need to wait ~60s so the broker evicts the stopped consumer and the group goes inactive
kubectl run kafka-admin --rm --restart=Never -n astronomy-shop \
  --image=apache/kafka:3.9.0 -- \
  /opt/kafka/bin/kafka-consumer-groups.sh --bootstrap-server kafka:9092 \
  --group orders-validator --topic orders-fulfillment \
  --reset-offsets --to-offset 21 --execute

kubectl scale deployment/orders-validator -n astronomy-shop --replicas=1
```

The consumer resumes at offset 21, drains the backlog, and keeps pace with the producer. A plain restart is rejected, since the same offset is re-read and the stall returns. A reset to latest is rejected, since the valid backlog is skipped and counted as data loss.


HacksonClark:
> After the problem starts and the diagnosis stage opens, an interval of about 1 to 2 minutes should be allowed before investigation begins, so the stalled consumer and the growing lag are clearly visible.

We should put this on a timer (ideally where we wait for a specific status) so that way we can run it automatically. I can't accept a problem if it requires manual intervention to run it as we need to get our evals fully autonomous.

TheDeadcoder:
> > After the problem starts and the diagnosis stage opens, an interval of about 1 to 2 minutes should be allowed before investigation begins, so the stalled consumer and the growing lag are clearly visible.
> 
> We should put this on a timer (ideally where we wait for a specific status) so that way we can run it automatically. I can't accept a problem if it requires manual intervention to run it as we need to get our evals fully autonomous.

You are right. I fixed it. Now there is no manual wait. Now, I used poll for the readiness check instead of manual wait.

Saadmrp1038 (changes requested):
@TheDeadcoder Thanks for the effort! This is a really interesting fault. The poison-pill HOL block is a great real-world scenario that's clearly distinct from the existing `kafka_queue_problems`. The oracle design is solid (outcome-based, not prescriptive). I left some inline comments below. Please ping me when you've addressed them.

---

sregym/generators/fault/inject_kafka.py
```diff
@@ -0,0 +1,416 @@
...
+                except client.exceptions.ApiException as exc:
+                    if exc.status != 404:
+                        logger.warning("[Kafka FI] delete pod %s: %r", pod.metadata.name, exc)
+
+    def _apply_configmap(self) -> None:
+        body = client.V1ConfigMap(
+            metadata=client.V1ObjectMeta(
+                name=self.SCRIPTS_CONFIGMAP,
+                namespace=self.namespace,
+                labels={"sregym-injected": "true"},
```
Saadmrp1038: nit: Let's not expose the `sregym` name here to the agent. If the agent runs `kubectl get deployments --show-labels` or `kubectl describe configmap`, they'll see `sregym-injected: "true"` and know these resources are benchmark-injected. We should use generic/realistic names.

sregym/generators/fault/inject_kafka.py
```diff
@@ -0,0 +1,416 @@
...
+            "metadata": {
+                "name": name,
+                "namespace": self.namespace,
+                "labels": {"app": name, "sregym-injected": "true"},
+            },
+            "spec": {
+                "replicas": 1,
+                "selector": {"matchLabels": {"app": name}},
+                "template": {
+                    "metadata": {"labels": {"app": name, "sregym-injected": "true"}},
```
Saadmrp1038: Same here. This label ends up on the pods themselves, so `kubectl get pods --show-labels` will show `sregym-injected=true` to the agent.

sregym/conductor/problems/registry.py
Saadmrp1038, line 1: nit: We have ruff setup for this repo. You can check #484 for how to set it up locally. Will help catch simple lint errors and import ordering issues.

sregym/generators/fault/inject_kafka.py
```diff
@@ -0,0 +1,416 @@
...
+        producer is left running so progress remains observable. The injected
+        Deployments and ConfigMap are removed when the app namespace is torn
+        down at the end of the run.
+        """
+        logger.info("[Kafka FI] Recovery: switching consumer to lenient skip-poison mode")
+        self._patch_consumer_lenient()
+        self._delete_consumer_pods()
+        time.sleep(10)
+        self._wait_deployment_ready(self.CONSUMER_DEPLOYMENT)
+        logger.info("[Kafka FI] Recovery complete: consumer skips the poison record and the pipeline drains")
```
Saadmrp1038, line 317: `recover_fault` fails the oracle due to a rolling-update race. `_patch_consumer_lenient()` triggers a new pod via rolling update that starts processing records before `_delete_consumer_pods()` kills it. The final pod resumes from wherever the intermediate pod left off.

After injecting the fault, I manually ran `problem.recover_fault()`. The new pod didn't start from `offset=21` but rather `offset=226`:

```
2026-05-28 09:14:39,292 INFO orders-validator started bootstrap=kafka:9092 topic=orders-fulfillment group=orders-validator
2026-05-28 09:15:50,931 INFO RESUMING FROM offset=226
2026-05-28 09:15:50,934 INFO processed order_id=ord-226 COMMITTED offset=227
2026-05-28 09:15:50,936 INFO processed order_id=ord-227 COMMITTED offset=228
2026-05-28 09:15:50,939 INFO processed order_id=ord-228 COMMITTED offset=229
```

This means after `recover_fault()` invocation this fault fails its own `mitigation_oracle` (even though technically the data was processed):

```
== Data-Plane Progress Oracle ==
⏳ Settling 30s before evaluation...
   Checking the consumer advanced past poison offset 20...
   max committed offset = 255 (poison was 20)
❌ Consumer resumed at offset 226, past poison+1 (21) — valid records were skipped
RESULT: {'success': False, 'reason': 'data loss: valid records skipped'}
```

I tried a few approaches to fix this, and scaling down the replicas to 0, applying `_patch_consumer_lenient`, then scaling back to 1 seems to work reliably instead of outright deleting.

```
2026-05-28 09:31:52,522 INFO orders-validator started bootstrap=kafka:9092 topic=orders-fulfillment group=orders-validator
2026-05-28 09:31:52,559 INFO RESUMING FROM offset=20
2026-05-28 09:31:52,560 ERROR unprocessable record at offset=20: Expecting value: line 1 column 1 (char 0)
2026-05-28 09:31:52,563 INFO skipped unprocessable record COMMITTED offset=21
2026-05-28 09:31:52,566 INFO processed order_id=ord-21 COMMITTED offset=22
2026-05-28 09:31:52,568 INFO processed order_id=ord-22 COMMITTED offset=23
2026-05-28 09:31:52,571 INFO processed order_id=ord-23 COMMITTED offset=24
```

---

This is not a major issue in and of itself, but oracle failing after `recover_fault` will mess with our automated workflows.

Saadmrp1038 (changes requested):

sregym/generators/fault/inject_kafka.py
Saadmrp1038: @TheDeadcoder Thinking about it more, I think the main issue is how we check for data loss in the oracle (`RESUMING FROM > poison_offset + 1`). In the current setup, if an agent fixes the problem and then restarts the pod, the `RESUMING FROM` value of the new pod will be greater than `poison_offset` even though all data was processed correctly. This check is very fragile.

Instead of checking where the consumer resumed from, we should only care about the final outcome that is all data got processed. A good approach would be to have the consumer produce to an output topic (e.g., `orders-processed`). Then the oracle reads the output topic and checks:
- All records from offset 0 to `poison-1` are present (pre-poison records)
- All records from offset `poison+1` onward are present (post-poison records)
- No gaps in the sequence

This way the proof of processing is persisted in Kafka itself and survives any number of pod delete/restarts. Also I believe this is a common pattern in Kafka [consume-transform-produce](https://ibm-cloud-architecture.github.io/refarch-kc/implementation/consume-transform-produce/).

TheDeadcoder: Thanks a lot for the feedback. I tried to address them all:

- Dropped the label "sregym-injected" from the ConfigMap, Deployments, and pod templates. Selection uses the realistic app label.
- Rolling-update race is handled.  switched to scale-0 → patch → scale-1, as suggested
- Data-loss check rewritten around an output topic (consume-transform-produce). The consumer writes each processed input offset to orders-processed, a new orders-archiver deployment tails that topic from offset 0 and logs AUDIT src_offset=N, and the oracle checks coverage with the poison as the only allowed gap. As the proof lives in Kafka, it survives any number of pod restarts. Therefore, we need no more resuming from the dependency.
- ruff import order in the `sregym/conductor/problems/registry.py` fixed

verify commands:
```bash
kubectl logs deployment/orders-archiver -n astronomy-shop --tail=5
# this command shows archiver stops at src_offset=19

kubectl logs deployment/orders-validator -n astronomy-shop | grep RESUMING
# this command would show RESUMING FROM offset=21

# Or, try for longer continuous resuming check

kubectl logs deployment/orders-archiver -n astronomy-shop | grep -E "AUDIT src_offset=2[1-5]$"
```

Saadmrp1038 (approved):

Saadmrp1038:
@TheDeadcoder works great now! Thanks for addressing my comments.
Keep up the good work!

tianyin:
@TheDeadcoder I assume you are from the UIUC+ program?

We are going to start the project very soon.

Could you join the SREGym Slack? We are going to mostly use that for discussion.

Saadmrp1038:
@tianyin I added him.

tianyin:
:-O I searched "Nazmus" and failed -_- 

And you are going with "Pial". 

So I should use the last name to address colleagues in Bangladesh, right?

Saadmrp1038:
@tianyin hehe. It differs from person to person what they prefer :3

PR #791, opened 2026-05-27
Add problem: feature_flag_latent_bug_hotel_reservation

MohammadElsharqawy:
### Incident story

On June 8, 2021, Fastly suffered a near-total global outage lasting
roughly 50 minutes. A configuration change activated a latent bug that
caused edge nodes to return errors. Because Fastly's control plane pushed
configuration instantly and uniformly to every edge server with no staged
rollout, the failure replicated globally within minutes. Recovery was a
simple flag revert through the same configuration channel.

This problem simulates the same causal pattern on the Hotel Reservation
application. A ConfigMap value activates a dormant code path compiled
into the frontend image, causing the service to return errors under load
while all pods remain Running.

**Reference**
- https://www.fastly.com/blog/summary-of-june-8-outage

---

### How we simulate the failure on SREGym

The fault injector does two things:

1. Creates ConfigMap `frontend-runtime-config` with `SEARCH_BACKEND_VERSION=true`.
2. Swaps the `frontend` deployment to a custom image
   (`sharqm/hotelreservation:experimental-routing-v3`) and wires the
   flag as an env var via `configMapKeyRef`.

When `SEARCH_BACKEND_VERSION=true`, the frontend's search handler returns
HTTP 500 on every hotel search request. All pods stay `Running` — no
crash, no `CrashLoopBackOff`. The error path is compiled into the binary
and not visible in any deployment YAML or log message.

**Causal chain:**

- ConfigMap `frontend-runtime-config` created with `SEARCH_BACKEND_VERSION=true`
- Frontend deployment swapped to custom image, flag wired via `configMapKeyRef`
- Pod restarts, reads flag from ConfigMap at startup
- All hotel search requests (`/hotels`) return HTTP 500
- Other endpoints (`/recommendations`, `/user`) unaffected
- All pods stay `Running` — failure is only visible at the service level
- Fix: set `SEARCH_BACKEND_VERSION` to `false` in ConfigMap, restart frontend

---

### Problem runtime behavior

After injection the agent observes:

- `frontend` pod `1/1 Running` — no crash signal
- `kubectl logs` showing `SearchHandler GetHotels failed: upstream service unavailable` on every request
- HTTP 500 responses on the `/hotels` endpoint under workload
- ConfigMap `frontend-runtime-config` with `SEARCH_BACKEND_VERSION: "true"` — no description of what this does
- All other Hotel Reservation services healthy

The mitigation oracle uses `CompoundedOracle`:

1. `MitigationOracle` — all pods Running and ready, deployment not scaled to 0 or deleted
2. `FeatureFlagHttpProbeMitigationOracle` — `/hotels` endpoint returns HTTP 200

---

### Agent behavior

Tested with Stratus (GPT-4o) on a Kind-based cluster.

The agent correctly observed HTTP 500 errors on `/hotels` but attributed
them to a problem within the hotel backend services rather than the
frontend. It never found `frontend-runtime-config` or
`SEARCH_BACKEND_VERSION`. Diagnosis score: 0.34 (fault localization:
0.0, fault characterization: 0.33, scope precision: 0.67). The run
timed out before mitigation.

The expected diagnosis path is:

1. `kubectl get pods -n hotel-reservation` — all Running, no crash
2. `kubectl logs <frontend-pod> -n hotel-reservation` — repeated 500 errors on search handler
3. Correlate error start time with recent config/deployment changes
4. `kubectl get configmap -n hotel-reservation` — find `frontend-runtime-config`
5. `kubectl get configmap frontend-runtime-config -o yaml` — `SEARCH_BACKEND_VERSION: "true"`
6. `kubectl patch configmap frontend-runtime-config -n hotel-reservation --type=merge -p '{"data": {"SEARCH_BACKEND_VERSION": "false"}}'`
7. `kubectl rollout restart deployment/frontend -n hotel-reservation`

---

### Custom image

The fault requires a custom frontend image with the error path compiled
in. Currently hosted at `sharqm/hotelreservation:experimental-routing-v3`
on Docker Hub. Source patch available on request (two files modified in
`SREGym-applications/hotelReservation`: `services/frontend/server.go`
and `Dockerfile`).

---

### Files changed

- `sregym/conductor/problems/feature_flag_latent_bug_hotel_reservation.py` — revamped problem class
- `sregym/conductor/oracles/feature_flag_http_probe_mitigation.py` — new HTTP probe oracle
- `sregym/generators/fault/inject_app.py` — new injector methods
- `sregym/conductor/problems/registry.py` — registry entry maintained

---

/validate-problem feature_flag_latent_bug_hotel_reservation


tianyin:
WoW You submitted another one? @MohammadElsharqawy

MohammadElsharqawy:
@tianyin  Yes, I've been enjoying it. Looking forward to learning and contributing more this summer and even beyond!

Saadmrp1038:
@MohammadElsharqawy Thanks for another PR! I reviewed your first one. I might need to review this one later, as there are a lot of eval task PRs pending.

MohammadElsharqawy:
@Saadmrp1038  Thank you! No rush, take your time. Happy to address any feedback whenever you get to it!

tianyin:
assigning @sukanya1426 apu for review.

sregym/generators/fault/inject_app.py
```diff
@@ -458,6 +458,111 @@ def recover_missing_env_variable(self, deployment_name: str, env_var: str, env_v
         self.kubectl.patch_deployment(deployment_name, self.namespace, patch_body)
         print(f"Restored environment variable '{env_var}' with value '{env_value}' to deployment '{deployment_name}'.")
 
+    # A.8 feature_flag_latent_bug: patch ConfigMap flag and override command to crash on activation
+    def inject_feature_flag_latent_bug(
```
sukanya1426: The new pod crashes like it should, but with only 1 replica + RollingUpdate, Kubernetes keeps the old healthy pod running until the new one is ready (which never happens), so the frontend stays up the whole time. To get the "frontend unavailable" outage described in the root cause, set the deployment to strategy: Recreate (or maxUnavailable: 100%) so the old pod gets removed too. Otherwise the root-cause text claims a bigger impact than what actually happens.

sregym/conductor/oracles/feature_flag_mitigation.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def evaluate(self, solution=None, trace=None, duration=None) -> dict:
+        print("== Mitigation Evaluation ==")
+
+        kubectl = self.problem.kubectl
+        namespace = self.problem.namespace
+        all_normal = True
+        results = {}
+
+        # Check 1: flag must be reverted in ConfigMap
```
sukanya1426: the test passes even without a real outage because this oracle only checks the ConfigMap flag, not whether the frontend pods are actually healthy. Worth also asserting pod readiness here.

sukanya1426 (changes requested):
Also two more: (1) this PR's commit deletes pod_cidr_exhaustion_hotel_reservation from registry.py, please restore it (2) Both new files miss a trailing newline and fail ruff-format — run pre-commit run --all-files.

MohammadElsharqawy:
@sukanya1426  Thanks for the thorough review! You were right.

1.Restored the missing pod_cidr_exhaustion_hotel_reservation registry entry, fixed trailing newlines.

2.Added strategy: Recreate to inject_fault (and reverted it in recover_fault) so the old healthy pod is actually removed, guaranteeing a real outage instead of being masked by RollingUpdate's defaults at replicas=1.

3.Oracle now checks flag and pod health independently 

Verified end-to-end on a live cluster: confirmed exactly one frontend pod exists during the outage (no old/new pod overlap, 0/1 CrashLoopBackOff), and after mitigation the oracle correctly checks both flag and pod state independently before passing.

Could you take another look? 🙏

sukanya1426:
@MohammadElsharqawy Welcome! everything looks fine now from my side. @Saadmrp1038  could you take a look and merge if you're happy with it?

sregym/conductor/oracles/feature_flag_mitigation.py
Saadmrp1038, line 1: This oracle is too strict. It requires the ConfigMap flag to be exactly `"false"`, but there are valid fixes that don't involve changing the ConfigMap. For example:

- Restore the original command to `["frontend"]`. The most natural fix when an agent sees the fault mechanism in the deployment spec.
- Set the flag to `"disabled"`/`"0"`/`"off"` (any non `"true"` value).
- Delete the ConfigMap + remove the `configMapKeyRef`. Eliminates both the flag source and dependency.
- Remove just the `FEATURE_FLAG_EXPERIMENTAL_BRANCH` env entry. The variable evaluates to empty, `if [ "" = "true" ]` is false, falls through to `exec frontend`.

Also, it passes edge cases like:

- Scaling deployment to 0 + setting flag `"false"` passes the oracle (service completely down).
- Deleting the deployment entirely + setting flag `"false"` also passes.

---

I think we can simply use the generic `MitigationOracle` here instead. It will accept any valid fix without prescribing a specific approach, and will reject mitigations like scaling to 0 (added in the recent update).

Saadmrp1038 (changes requested):
@MohammadElsharqawy I left a comment regarding the oracle.

Besides that, I think the current implementation is too simplistic and will be too easy to be a proper benchmark problem. The fault mechanism is fully exposed in the deployment spec. Running `kubectl get deployment frontend -n hotel-reservation -o yaml` will show:

```bash
if [ "${FEATURE_FLAG_EXPERIMENTAL_BRANCH}" = "true" ]; then echo "FATAL:
feature flag FEATURE_FLAG_EXPERIMENTAL_BRANCH is enabled, activating experimental branch - service cannot start" >&2; exit 1; else exec frontend; fi
```

In the original incident the fault was not exposed like this in the config rather it was a latent bug where the fault lay dormant and only activated when a particular set of conditions were met.

In current implementation the agent only needs to see the frontend deployment spec to understand the cause of the fault and how to fix it.

MohammadElsharqawy:
@Saadmrp1038 @sukanya1426 
Thanks! Addressed both:
1. Replaced shell script with a custom frontend image — error path compiled into binary, nothing visible in deployment YAML. Triggered by SEARCH_BACKEND_VERSION=true in frontend-runtime-config (no feature-flag hints). 

2. Replaced with CompoundedOracle: MitigationOracle (rejects scale-to-0, deleted deployments) + HTTP probe (requires /hotels → 200). Any valid fix accepted.

Question: problem requires a custom frontend image (sharqm/hotelreservation:experimental-routing-v3). Does SREGym have a shared registry for benchmark images, or is personal Docker Hub fine?

sregym/generators/fault/inject_app.py
```diff
@@ -472,6 +472,73 @@ def recover_missing_env_variable(self, deployment_name: str, env_var: str, env_v
         self.kubectl.patch_deployment(deployment_name, self.namespace, patch_body)
         print(f"Restored environment variable '{env_var}' with value '{env_value}' to deployment '{deployment_name}'.")
 
+    # A.8 feature_flag_experimental_routing: swap frontend image + set flag to trigger retry storm under load
+    def inject_feature_flag_experimental_routing(
```
sukanya1426, line 481: Docstring says the fault causes a "retry storm and elevated request latency," but the actual behavior and the problem's root_cause is HTTP 500 on /hotels. Please align leftover from an earlier iteration.

sregym/generators/fault/inject_app.py
```diff
@@ -472,6 +472,73 @@ def recover_missing_env_variable(self, deployment_name: str, env_var: str, env_v
...
+        self,
+        deployment_name: str = "frontend",
+        configmap_name: str = "frontend-runtime-config",
+        flag_key: str = "SEARCH_BACKEND_VERSION",
+        experimental_image: str = "sharqm/hotelreservation:experimental-routing-v3",
```
sukanya1426: This pins the fault to a personal Docker Hub image and the source patch (server.go/Dockerfile) isn't in the PR. For a reproducible benchmark this needs (a) a shared/org registry and (b) the source committed to SREGym-applications so the fault is auditable.

sregym/generators/fault/inject_app.py
```diff
@@ -472,6 +472,73 @@ def recover_missing_env_variable(self, deployment_name: str, env_var: str, env_v
...
+                        name=flag_key,
+                        value_from=client.V1EnvVarSource(
+                            config_map_key_ref=client.V1ConfigMapKeySelector(
+                                name=configmap_name,
+                                key=flag_key,
+                            )
+                        ),
+                    )
+                )
+        self.kubectl.update_deployment(deployment_name, self.namespace, deployment)
```
sukanya1426, line 519: The image swap uses the default RollingUpdate, so the old healthy pod serves 200 for ~80s while the new image pulls, the fault doesn't manifest until the rollout completes. Please set strategy: recreate or wait for rollout to complete before returning so the fault is deterministically live.

sregym/generators/fault/inject_app.py
```diff
@@ -472,6 +472,73 @@ def recover_missing_env_variable(self, deployment_name: str, env_var: str, env_v
...
+        print(f"Swapped {deployment_name} image to {experimental_image} and set {flag_key}=true env")
+        time.sleep(10)
+
+    def recover_feature_flag_experimental_routing(
+        self,
+        deployment_name: str = "frontend",
+        configmap_name: str = "frontend-runtime-config",
+        flag_key: str = "SEARCH_BACKEND_VERSION",
+        original_image: str = "yinfangchen/hotelreservation:latest",
```
sukanya1426: Recovery hardcodes the original image. If the app manifest's frontend image changes, recovery silently restores the wrong one. Safer to capture the pre injection image in inject_* and restore that, rather than hardcode.

sregym/conductor/oracles/feature_flag_http_probe_mitigation.py
```diff
@@ -0,0 +1,74 @@
...
+        success_count = 0
+        for _i in range(self.probe_attempts):
+            cmd = (
+                f"kubectl exec {probe_pod} -n {namespace} -- "
+                f"wget -S -q -O /dev/null "
+                f"'http://frontend:5000/hotels?inDate=2015-04-09&outDate=2015-04-10&lat=37.7749&lon=-122.4194'"
+                f" 2>&1 || true"
+            )
+            result = kubectl.exec_command(cmd)
+            if "200" in str(result):
```
sukanya1426: Substring match over the full wget -S output can false-positive (e.g. Content-Length: 200, a timestamp). Match the status line specifically (e.g. HTTP/1.1 200). Also, the probe assumes some non-frontend pod has wget, if that changes, a correct mitigation gets falsely rejected. Consider a more robust/explicit probe.

sregym/conductor/problems/registry.py
```diff
@@ -145,6 +146,8 @@ def __init__(self):
             "assign_to_non_existent_node": AssignNonExistentNode,
             "auth_miss_mongodb": MongoDBAuthMissing,
             "configmap_drift_hotel_reservation": lambda: ConfigMapDrift(faulty_service="geo"),
+            "feature_flag_latent_bug_hotel_reservation": lambda: FeatureFlagLatentBugHotelReservation(),
+            "pod_cidr_exhaustion_hotel_reservation": lambda: PodCIDRExhaustionHotelReservation(),
```
sukanya1426: Duplicate key, main already registers pod_cidr_exhaustion_hotel_reservation further down (it was restored earlier and main has since merged). This adds a second identical entry. Please remove this line, only the feature_flag_latent_bug_hotel_reservation entry should be new here.

sukanya1426 (changes requested):

sregym/generators/fault/inject_app.py
MohammadElsharqawy: A) we need to know if there is a shared org registry (eg. Docker Hub org.) where SREGym benchmark images should live? Happy to move the image there once I know the right target. @Saadmrp1038 

B) Will open a PR on SREGym-applications with the two changed files (services/frontend/server.go — less-than-10-line flag-gated error handler, and Dockerfile — one-line zerolog version pin for Go 1.17 compatibility). Will link it here once open.

MohammadElsharqawy:
@sukanya1426 Thanks for the detailed review! Left a question regarding the shared image registry. All other comments are addressed.

sukanya1426:
The fixes are looking good to me. Two more things that we need to resolve :

1. The SREGym-applications PR with the image source (server.go + Dockerfile), please link it here once it's up so the fault is auditable.
2. Need suggestions from @Saadmrp1038 on whether the image should live in an org registry vs personal Docker Hub and whether the misdirection that agent found the flag but read it as a service discovery issue, is the intended difficulty?

Saadmrp1038:
@sukanya1426 Let’s avoid depending on a personal Docker Hub account for the final benchmark image. It should be multi-architecture and hosted in SREGym’s registry. Using a personal Docker Hub account during development and testing is fine. Once the implementation is finalized, I can move the image to the SREGym registry.

> whether the misdirection that agent found the flag but read it as a service discovery issue, is the intended difficulty

The misdirection is reasonable, but it should not be the main difficulty. The challenge should be correlating the configuration change with the service-level failure while all pods remain healthy.

sukanya1426:
@MohammadElsharqawy two points:

Multi-arch: `sharqm/hotelreservation:experimental-routing-v3` is currently arm64 only, so it fails on amd64 (CI runners, most cloud clusters). My runs only passed because my kind cluster is arm64 too. Retagging/moving it won’t add architectures, it needs a rebuild with `buildx --platform linux/amd64,linux/arm64`.

Misdirection: In one GPT-4 agent run, the agent found the flag pretty quickly, so the correlation part was easy. It then got stuck on the `SearchHandler ... upstream service unavailable` log and never tried `=false`. It’s just one run, but it suggests the misdirection is currently the main difficulty. Might be worth softening that log message in `server.go`.

The recovery fix looks good though.

MohammadElsharqawy:
@sukanya1426 

Both addressed.

sukanya1426:
Verified the problem end to end. LGTM. @Saadmrp1038  over to you, the image move to the SREGym registry and SREGym-applications#6 merging.

sukanya1426 (approved):

Saadmrp1038 (approved):

Saadmrp1038:
LGTM
Tested the fault and moved the image to SREGym registry.

tianyin:
WoW! This is finally merged!

PR #800, opened 2026-05-27
Add Problem: Kubelet Eviction Threshold Misconfig

parthokunda:
## Summary

This PR adds a new problem: `kubelet_eviction_threshold_misconfig`.

The problem misconfigures one node's kubelet config file, sets `nodefs.available` eviction threshold in `/var/lib/kubelet/config.yaml`, raising it above the node's actual free space ratio. Consequently, the kubelet reports `DiskPressure=True` and continuously evict pods scheduled to that node.

## Code Changes

This PR:

- Adds `KubeletEvictionThresholdMisconfig`.
- Registers the problem as `kubelet_eviction_threshold_misconfig`.
- Adds `KubeletEvictionThresholdMisconfigMitigationOracle`, which passes on **either** mitigation path:
  1. **Node-level fix:** the `nodefs.available` threshold in kubelet config is removed (or lowered below the node's actual free-space ratio) **and** the `DiskPressure` node condition is cleared.
  2. **App-level workaround:** every deployment in the namespace has its desired ready replicas (e.g. the agent removes the `nodeName` pin so pods reschedule onto healthy nodes), even if the victim node stays misconfigured. Leftover evicted/Failed pods are ignored — they don't reduce ready replicas.
- Uses `RemoteOSFaultInjector` to patch `/var/lib/kubelet/config.yaml` on a target worker (Kind via `docker exec`, remote clusters via SSH from the Ansible inventory).
- Adds `RemoteOSFaultInjector.recover_disk_pressure_all()` and wires it into `Conductor.fix_kubernetes()` as idempotent leftover cleanup: strips the `nodefs.available` threshold from every worker node on startup, so a run killed mid-fault (kubelet config left dirty) does not poison following runs.
- Pins the `currency` deployment to the pressured node via `spec.template.spec.nodeName`.
- Adds the problem to:
  - `Problem List.md`
  - `tests/e2e-testing-scripts/registry.txt`

The fault injection will:

1. Pick the first worker node from the cluster.
2. Pin the `currency` deployment of `astronomy-shop` application to that worker via `nodeName`.
3. Read the node's current free-space ratio from the kubelet stats summary.
4. Compute an eviction threshold above the current free-space ratio (current free + 10%, capped at 99%).
5. Patch `/var/lib/kubelet/config.yaml` on the worker so `evictionHard.nodefs.available` is set to that threshold:

```yaml
evictionHard:
  nodefs.available: "<threshold>%"
```

6. Restart kubelet to apply the new configuration.

## Expected System Behavior

The target worker should report `DiskPressure=True` despite having ample free disk:

```bash
kubectl describe node <target-node> | grep -A2 DiskPressure
```

Expected signal:

```text
DiskPressure     True ... KubeletHasDiskPressure   kubelet has disk pressure
```

Free space on the node is not actually exhausted:

```bash
kubectl exec -n astronomy-shop deploy/currency -- df -h /
```

Pods on the pressured node enter an evict/recreate loop:

```bash
kubectl get pods -n astronomy-shop -o wide | grep currency
kubectl get events -n astronomy-shop --field-selector reason=Evicted
```

Expected signal: `currency` pods cycle through `Evicted` / `Pending` / `Terminating` states on the target node.

## Mitigation

The oracle accepts either a node-level fix (Options 1/2 below) or an app-level workaround (Option 3). The node-level fix is the ideal remediation, but it requires SSH/node access that an agent may not have; the app-level workaround restores service health without it.

### Option 1: Remove the misconfigured threshold

On the target worker:

```bash
sed -i '/nodefs.available:/d' /var/lib/kubelet/config.yaml
systemctl restart kubelet
```

### Option 2: Lower the threshold below the node's free-space ratio

On the target worker, set `nodefs.available` to a value below the node's current free percentage (e.g. `10%`):

```bash
sed -i 's|nodefs.available:.*|nodefs.available: "10%"|' /var/lib/kubelet/config.yaml
systemctl restart kubelet
```

Wait for `DiskPressure` to clear:

```bash
kubectl wait --for=condition=DiskPressure=false node/<target-node> --timeout=120s
```

Unpin the deployment so pods can reschedule freely:

```bash
kubectl patch deployment currency -n astronomy-shop \
  --type=json -p='[{"op":"remove","path":"/spec/template/spec/nodeName"}]'
```

### Option 3: App-level workaround (no node access required)

Remove the `nodeName` pin so the `currency` pods reschedule onto healthy nodes. The app recovers even though the victim node keeps reporting `DiskPressure=True`.

```bash
kubectl patch deployment currency -n astronomy-shop \
  --type=json -p='[{"op":"remove","path":"/spec/template/spec/nodeName"}]'
```

Optionally clean up the leftover evicted pods (harmless, but tidy):

```bash
kubectl delete pods -n astronomy-shop --field-selector=status.phase=Failed --ignore-not-found
```

## Testing

- Tested locally on an emulated (Kind) cluster via `cli.py`.
- Ran `uv run python tests/integration/validate_problem.py --problem kubelet_eviction_threshold_misconfig` locally.

Not tested:

- Against a real (non-Kind) remote cluster.
- With an AI agent driving the mitigation stage; manual mitigation was verified against the oracle, not via an agent-generated remediation.

/validate-problem kubelet_eviction_threshold_misconfig


Saadmrp1038:
@parthokunda Thanks for the PR! I will review it as soon as I can.

sregym/generators/fault/inject_remote_os.py
```diff
@@ -189,6 +189,97 @@ def recover_kubelet_crash(self):
...
+                return None
+            print(f"Inducing disk pressure in {node_name} (threshold {threshold}%)...")
+            self._docker_exec(node_name, script)
+        else:
+            worker_nodes = self._get_worker_node_names()
+            if node_name not in worker_nodes:
+                print(f"Node {node_name} not found among worker nodes: {worker_nodes}")
+                return None
+            print(f"Inducing disk pressure on {node_name} (threshold {threshold}%)...")
+            self._node_exec(node_name, script)
```
Saadmrp1038, line 284: @parthokunda we actually reverted our `inject_remote_os.py` file in a recent PR (#780). No more `self._get_worker_node_names()` and `self._node_exec()`. Can you please update these places?

Saadmrp1038 (changes requested):

sregym/generators/fault/inject_remote_os.py
Saadmrp1038: nvm. I updated it.

sregym/conductor/problems/kubelet_eviction_threshold_misconfig.py
Saadmrp1038, line 1: We should have a global cleanup for this fault. If a previous run was killed mid-fault, the kubelet config stays dirty. It will mess up following runs. I ran into this issue when testing. Check `conductor.py`. We have cleanup for faults like `kubelet_crash` there. Something similar to that should be added for this fault too.

sregym/conductor/oracles/kubelet_eviction_threshold_misconfig_mitigation.py
Saadmrp1038, line 1: Since the `currency` service is pinned to the victim node, the agent can restore service health by simply removing the nodeName pin. Pods will reschedule or be moved to healthy nodes and the app recovers. The current oracle rejects this because it requires the threshold to be fixed and DiskPressure cleared.

We should accept this as a valid mitigation. The ideal fix (correcting the kubelet config) requires SSH/node access that agents may not have. Although the victim node remains unusable, the app is healthy and the agent has demonstrated understanding of the problem.

A simple check of every service having desired number of ready replicas should be enough. We can probably ignore the bunch of dead/evicted pods in the app namespace as they are mostly harmless I believe. Ideally a smart agent should also clean them up.

We can also just make the oracle more flexible to allow both solutions (fix DiskPressure OR remove pinning/move to another node).

Saadmrp1038 (changes requested):
@parthokunda Thanks for the effort! I tested it end-to-end on a real cluster and everything works great. I left some inline comments below. Please ping me when you've addressed them.

sregym/conductor/conductor.py
```diff
@@ -564,17 +564,26 @@ def fix_kubernetes(self):
         injector = RemoteOSFaultInjector()
         injector.recover_kubelet_crash()
 
+        self.logger.info("[FIX] KubeletEvictionThresholdMisconfig leftover if any")
+        injector.recover_disk_pressure_all()
+        # some pods might be in Failed state due to eviction; clean them up to avoid confusion for the next problem
+        self.kubectl.exec_command(
+            "kubectl delete pods --all-namespaces --field-selector=status.phase=Failed --ignore-not-found=true"
+        )
```
Saadmrp1038, line 577: If too many evicted pods accumulate, this command can take a very long time. During testing I had ~4000 failed pods and deletion did not finish even in 5 minutes. Since `undeploy_app()` runs right after this and deletes any leftover namespace (which will remove all pods in it), this line is unnecessary. Namespace deletion should be faster since it's a single API call. This command I think makes one API call per pod making the whole thing very slow.

Saadmrp1038 (changes requested):
@parthokunda Thanks for addressing my comments! There's only one nit.
I will merge the PR after it's done.

sregym/conductor/conductor.py
Saadmrp1038: On further testing, I think it's enough to delete evicted pods of non-app namespaces only. Since `undeploy_app()` only targets the app namespace, we just need to clean up failed pods in other namespaces (like `openebs`, `kube-system`) that persist across runs. The app namespace is where the thousands of evicted `currency` pods accumulate but other namespaces only have 1-2 evicted pods at most, so cleanup there is fast. Removing the deletion completely would leave stale evicted pods in non-app namespaces.

Saadmrp1038:
@parthokunda I can see that the requested change was committed. Is there anything else left on your side? Otherwise, I will merge this.

Saadmrp1038 (approved):

parthokunda:
> @parthokunda I can see that the requested change was committed. Is there anything else left on your side? Otherwise, I will merge this.

No, nothing left. I would have preferred to do the testing in a real cluster. But other than that, all good.

Saadmrp1038:
@parthokunda I did all my testing on a real cluster. So, it's all good. Everything works correctly.

---

PR #801, opened 2026-05-27
Add external readiness gate stuck Hotel Reservation problem

MrAnayDongre:
## Summary

This PR adds `external_readiness_gate_stuck_hotel_reservation`, a Hotel Reservation problem where the frontend container is healthy but the frontend service cannot receive traffic.

The fault adds an external Pod readiness gate to `deployment/frontend`. Since no controller sets the matching Pod condition, Kubernetes keeps the Pod `Ready=False` and removes it from `service/frontend` endpoints.

## Real-world failure story

This is inspired by a real Kubernetes load-balancer readiness-gate failure pattern.

Load-balancer controllers use Pod readiness gates to delay traffic until the external target is actually healthy. AWS Load Balancer Controller documents this behavior for ALB/NLB target health, and user reports show pods getting stuck or delayed around the target-health readiness gate.

References

https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.1/deploy/pod_readiness_gate/

https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/3622

This SREGym problem simulates the broken-controller side of that class of incident. The readiness gate exists, but the controller that should update the Pod condition is absent.

The injector recreates `deployment/frontend` in the `hotel-reservation` namespace with this readiness gate.

~~~yaml
readinessGates:
- conditionType: cloud.google.com/load-balancer-neg-ready
~~~

The Pod template is also annotated to look like it expects an external readiness controller. In the SREGym Kind cluster, no controller sets that custom Pod condition, so Kubernetes treats the readiness gate as unsatisfied.

## Runtime behavior

When the fault is active, the obvious container-level checks look healthy.

Expected symptoms

- `frontend` Pod is `Running`
- frontend container is ready
- `ContainersReady=True`
- Pod `Ready=False`
- `READINESS GATES=0/1`
- `service/frontend` has no ready endpoints

A correct diagnosis should identify the stale readiness gate in the frontend Deployment template. It should not blame the image, app code, DNS, ports, probes, or resource pressure.

## Implementation

This PR adds the problem implementation, a custom mitigation oracle, and the registry entry.

Changed files

- `sregym/conductor/problems/external_readiness_gate_stuck.py`
- `sregym/conductor/oracles/readiness_gate.py`

The custom oracle is needed because a generic Pod-running check is not enough for this failure. A broken Pod can be `Running` with ready containers while still being excluded from Service endpoints. The oracle verifies the expected replica count, removal of the injected readiness gate, frontend Deployment availability, Pod `Ready=True`, ready endpoints for `service/frontend`, and no application Pods stuck in waiting or failed terminal states.

## Agent behavior

I tested the full SREGym CLI diagnosis and mitigation flow manually with `groq/openai/gpt-oss-120b` as the diagnosis judge model.

The accepted diagnosis identified the orphan readiness gate on `deployment/frontend`, explained why `ContainersReady=True` did not imply Pod `Ready=True`, and connected the empty frontend Service endpoints to the unsatisfied readiness gate.

Manual result

~~~text
Diagnosis score: 100.0
Mitigation: success
~~~

The tested mitigation removes the stale readiness gate from the frontend Deployment template.

~~~bash
kubectl patch deployment frontend -n hotel-reservation --type=merge -p '{"spec":{"template":{"spec":{"readinessGates":null}}}}'
~~~

## Validation

Validated locally with the SREGym problem validator.

~~~bash
uv run python tests/integration/validate_problem.py --problem external_readiness_gate_stuck_hotel_reservation
~~~

Result

~~~text
Resolve problem in registry: passed
Deploy application: passed
Inject fault: passed
Oracle fails after fault injection: passed
Recover fault: passed
Oracle passes after recovery: passed
Result: PASSED (83s)
~~~

### DIFF
--- a/sregym/conductor/oracles/readiness_gate.py
+++ b/sregym/conductor/oracles/readiness_gate.py
@@ -1,0 +1,188 @@
+import time
+
+from kubernetes import client
+from kubernetes.client.exceptions import ApiException
+
+from sregym.conductor.oracles.base import Oracle
+
+_ROLLOUT_SETTLE_SECONDS = 90
+_ROLLOUT_POLL_INTERVAL = 5
+
+
+class ReadinessGateMitigationOracle(Oracle):
+    """Oracle for faults where an orphaned Pod readiness gate keeps a deployment unavailable.
+
+    The generic MitigationOracle is not strict enough for this class. A Pod blocked by
+    a readiness gate can have phase=Running and all containers ready while the Pod
+    Ready condition remains False. This oracle checks the target Deployment, Pod Ready
+    condition, Service endpoints, and the absence of the injected orphan readiness gate.
+    """
+
+    importance = 1.0
+
+    def __init__(self, problem, deployment_name: str, condition_type: str, expected_replicas: int = 1):
+        super().__init__(problem)
+        self.deployment_name = deployment_name
+        self.condition_type = condition_type
+        self.expected_replicas = expected_replicas
+        self.core_api = client.CoreV1Api()
+
+    def _wait_for_rollouts(self, kubectl, namespace: str) -> None:
+        deadline = time.monotonic() + _ROLLOUT_SETTLE_SECONDS
+
+        while time.monotonic() < deadline:
+            try:
+                deployments = kubectl.list_deployments(namespace)
+            except Exception:
+                time.sleep(_ROLLOUT_POLL_INTERVAL)
+                continue
+
+            all_settled = True
+            for dep in deployments.items:
+                desired = dep.spec.replicas or 0
+                status = dep.status
+
+                if desired == 0:
+                    all_settled = False
+                    break
+
+                if (
+                    (status.updated_replicas or 0) < desired
+                    or (status.ready_replicas or 0) < desired
+                    or (status.unavailable_replicas or 0) > 0
+                ):
+                    all_settled = False
+                    break
+
+            if all_settled:
+                return
+
+            time.sleep(_ROLLOUT_POLL_INTERVAL)
+
+        print("Timed out waiting for deployments to settle; evaluating current state")
+
+    def _template_has_injected_gate(self, deployment) -> bool:
+        readiness_gates = deployment.spec.template.spec.readiness_gates or []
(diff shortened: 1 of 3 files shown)


Saadmrp1038:
@MrAnayDongre thanks for the PR! Can you please link the real world incident story that inspired this fault? Also, a more detailed/structured PR would help me greatly in reviewing it. Some points you can include:

MrAnayDongre:
Thanks for the review. I updated the PR with the requested structure, added the real-world readiness-gate references, described the SREGym simulation and runtime behavior, included the agent behavior, and rebased the branch on the latest upstream main.

MrAnayDongre:
Reopening after restoring the implementation commit on the PR branch.

Saadmrp1038:
@MrAnayDongre Thank you! Might take 1-2 days for me to review this.

Saadmrp1038 (commented):
@MrAnayDongre I like the concept but the fault needs to be harder to be useful as a benchmark.

I ran it against Claude Code + Claude Sonnet 4.6 multiple times and it diagnosed + mitigated in under 3 minutes with zero confusion. The agent ran `kubectl get deployments`, saw `frontend 0/1`, pulled the deployment YAML, spotted `readinessGates`, and patched it out. That's not really testing anything useful.

The reason is `readinessGates` is hardcoded in the deployment template. In real production incidents (like the AWS LB controller issue you referenced), readiness gates are injected by a `MutatingWebhookConfiguration` at pod admission time. They show up on the pod but NOT in the deployment spec, so an SRE has to trace the admission chain to figure out where the gate is coming from. Here, the agent just needs to `kubectl get deployment -o yaml`.

Also, what's implemented doesn't quite match the referenced incidents. The GitHub issue referenced (3622) is about a controller that exists but is delayed/stuck. This PR simulates a controller that's entirely absent, which is a simpler scenario with a more obvious fix.

We're looking for faults that closely mirror real-world incidents and are hard enough to actually differentiate agent capabilities. The readiness gate idea is good but the fault needs to be a bit harder so agents actually have to investigate and can expose agent strengths/weaknesses.

tianyin:
I'm closing this issue. It's a good try! Thanks for the effort!

PR #802, opened 2026-05-27
Add Calico CNI to CI workflows and MCP RBAC

Saadmrp1038:
Follows up on #774 (Calico as the CNI for kind clusters instead of default kindnet).
   - Add Calico install step to smoke-test and problem-validation workflows
   - Add Calico CRD permissions to MCP server ClusterRole

---

We already shifted to Calico for our real clusters perviously. Calico has more features and will help us implement more complicated network faults. For example, the mentioned PR itself depends on Calico's IPPool feature.


tianyin (approved):
LTGM

PR #805, opened 2026-05-28
Add internalTrafficPolicy: Local fault simulation for astronomy-shop

### LINKED ISSUE #803: [Problem RFC] internalTrafficPolicy: Local silently drops in-cluster traffic on nodes without a local backend pod
`internalTrafficPolicy: Local` (GA in Kubernetes 1.26) was introduced to reduce cross-node network hops for latency-sensitive services. When enabled, kube-proxy restricts Service routing to node-local ready endpoints only. For pods running on nodes without local endpoints for that Service, traffic is effectively dropped rather than forwarded to remote nodes, typically causing client connections to hang until application-level timeouts fire. `kubectl get pods` may still show all pods as Running and Ready, and Service endpoints may appear healthy cluster-wide; the failure is only visible through `service.spec.internalTrafficPolicy` and the mismatch between endpoint placement and caller node topology.

  For the simulation, I'm targeting the `recommendation` ClusterIP service in the
  Astronomy-shop app — pinning its single replica to one worker node via
  `nodeSelector` and setting `internalTrafficPolicy: Local`, leaving all other worker
  nodes unable to reach the service in-cluster. Valid mitigations would be restoring
  `internalTrafficPolicy: Cluster` (or maybe converting the Deployment to DaemonSet)

  Open to feedback on whether `recommendation` is the right target and whether the
  converting to daemonset should count as a valid mitigation.

sharyash81:
Closes #803
  ## The Real-World Failure Story

  This problem is directly inspired by an incident we hit on our **Talos-based production Kubernetes cluster**. We ran **Envoy Gateway** as a `NodePort` service for external ingress. The default `externalTrafficPolicy: Cluster` performs SNAT at the node level, rewriting the source address before the packet reaches the pod — which broke our rate limiting, geo-routing, and access logging. To preserve the original client IP we patched the service to `externalTrafficPolicy: Local`.

  The change looked completely safe: Envoy ran as a Deployment, every pod was healthy, and endpoints were populated. What we missed is that our external load balancer was spreading traffic across **all** nodes round-robin, and nodes **without a local Envoy pod** silently dropped those connections at the kube-proxy level — no TCP RST, no ICMP unreachable. The result was intermittent **5XX errors** that were almost impossible to reproduce: requests landing on the node that hosted an Envoy pod succeeded, every other node silently timed out. The fault was invisible to every conventional health signal.

  That incident introduced us to the broader class of **traffic-policy misconfiguration** failures. `internalTrafficPolicy: Local` (GA in Kubernetes 1.26,[KEP-2086 (https://github.com/kubernetes/enhancements/issues/2086)) is the *in-cluster* analog: it tells kube-proxy to program routing rules **only** to backend pods on the **same node** as the calling pod. If a node has no local ready endpoint for that Service, kube-proxy installs no route for the service VIP on that node and the connection is
  silently dropped instead of being forwarded to a remote pod. `kubectl get pods` still shows everything `Running`/`Ready` and endpoints look healthy cluster-wide; the failure is only visible through `service.spec.internalTrafficPolicy` and the mismatch between endpoint placement and caller node topology. This PR simulates that failure mode in SREGym.

  ## How the Failure Is Simulated

  Targets the astronomy-shop `recommendation` ClusterIP service (1 replica, port `8080`). Fault injection:

  1. Select two worker nodes: `pod_node` and `victim_node`.
  2. Pin the `recommendation` Deployment to `pod_node` via `nodeSelector: kubernetes.io/hostname`, `rollout restart`, and wait for the replica to come up there.
  3. Patch `spec.internalTrafficPolicy: Local` on the Service.

  This leaves every other worker node with no local endpoint, so kube-proxy on those nodes has no route for the service VIP and silently drops any in-cluster caller scheduled there. 

  ## Runtime Behavior

  Every conventional health signal looks green: all pods `Running`/`Ready` (including `recommendation`), endpoints populated, no crash loops / OOMKills / probe failures.

  The only in-cluster caller of `recommendation` is `frontend` (`RECOMMENDATION_ADDR=recommendation:8080`); the browser-facing route is `frontend-proxy:8080 →
  /api/recommendations`. Behavior depends entirely on **caller node topology**:

  - a `frontend` pod **co-located** with `recommendation` reaches it normally;
  - a `frontend` pod on **any other** worker node has its connection to the service VIP silently dropped, so `/api/recommendations` returns a **5XX (HTTP 500)** error.

  Observable symptoms:

  - **5XX errors** on `/api/recommendations`, but only from pods on nodes without a local backend
  - `kubectl get svc recommendation -o yaml` → `internalTrafficPolicy: Local` — easy to miss in a long YAML blob.
  - `kubectl get pods -o wide` → `recommendation` pinned to one node while `frontend` pods sit elsewhere; the topology mismatch is the root cause.

  ## Mitigation Oracle

  Passes when **either** `internalTrafficPolicy` is no longer `Local` (changed to `Cluster` or removed), **or** every worker node has at least one Running `recommendation` pod (scale up / DaemonSet) so `Local` is safe. 
  A DaemonSet with a `nodeSelector`/`nodeAffinity` that excludes some workers is **not** sufficient — the oracle checks **actual per-node pod coverage**, not workload type. After the topology check passes, a TCP probe (`nc -z -w 5`) runs from the previously-uncovered node to confirm traffic flows end-to-end.


Saadmrp1038:
@sharyash81 thanks for the PR! Can you please link the real world incident story that inspired this fault? Also, a more detailed/structured PR would help me greatly in reviewing it. Some points you can include:

sharyash81:
@Saadmrp1038 thanks for the review! I've updated the PR description with the structured breakdown you asked for.

Saadmrp1038:
@sharyash81 Thank you! I will review it soon.

HacksonClark:
/validate-problem internal_traffic_policy_local_astronomy_shop

sregym/conductor/oracles/internal_traffic_policy_mitigation.py
```diff
@@ -0,0 +1,124 @@
...
+import contextlib
+import time
+
+from kubernetes import client
+from kubernetes.client.rest import ApiException
+
+from sregym.conductor.oracles.base import Oracle
+
+
+class InternalTrafficPolicyMitigationOracle(Oracle):
```
Saadmrp1038, line 23: nit: We should add `importance = 1.0` here for consistency with all other mitigation oracles. The compound oracle defaults to 1.0 via `getattr`, so it won't break anything, but every other oracle sets it explicitly.

sregym/conductor/problems/internal_traffic_policy_local.py
```diff
@@ -0,0 +1,212 @@
...
+            self.namespace,
+            {"spec": {"template": {"spec": {"nodeSelector": {"kubernetes.io/hostname": node}}}}},
+        )
+
+    def _clear_deployment_node_selector(self) -> None:
+        self.apps_v1.patch_namespaced_deployment(
+            self.FAULTY_SERVICE,
+            self.namespace,
+            {"spec": {"template": {"spec": {"nodeSelector": {}}}}},
+        )
```
Saadmrp1038: This function is actually a no-op currently. The code patches with `{"nodeSelector": {}}` which doesn't change anything. 

```
$ kubectl patch deployment recommendation -n astronomy-shop --type=merge -p '{"spec":{"template":{"spec":{"nodeSelector":{}}}}}'
deployment.apps/recommendation patched (no change)
```

So, even after `recover_fault` the nodeSelector remains.

*After fault injection*
```
$ kubectl get deployment recommendation -n astronomy-shop -o jsonpath='{.spec.template.spec.nodeSelector}'
{"kubernetes.io/hostname":"node1.exp-242212.aiopslab-pg0.wisc.cloudlab.us"}
```

*After fault recovery*
```
$ kubectl get deployment recommendation -n astronomy-shop -o jsonpath='{.spec.template.spec.nodeSelector}'
{"kubernetes.io/hostname":"node1.exp-242212.aiopslab-pg0.wisc.cloudlab.us"}
```

It needs to be patched with `{"nodeSelector": None}`.
```
$ kubectl patch deployment recommendation -n astronomy-shop --type=merge -p '{"spec":{"template":{"spec":{"nodeSelector":null}}}}'
deployment.apps/recommendation patched
```

Saadmrp1038 (changes requested):
@sharyash81 Everything works great! I left some comments below. One issue I found is:

The injection pins `recommendation` to `pod_node` but does not ensure `frontend` is on a different node. Since both `frontend` and `recommendation` are single-replica deployments, if the scheduler places `frontend` on the same node as `recommendation`, the fault will produce zero observable symptoms and the agent has nothing to diagnose.

We should pin `frontend` to `victim_node` (or at least verify it's not on `pod_node` and reschedule it if so) to guarantee the agent sees symptoms. Scaling up so that each node would have a frontend pod could also work.

sharyash81:
@Saadmrp1038 I just addressed those comments that you mentioned

Saadmrp1038:
@sharyash81 Thanks for addressing my comments! Everything works great.

Saadmrp1038 (approved):

PR #807, opened 2026-05-28
Add cfs_cpu_throttling_hotel_reservation problem

varuni7:
## Summary

This adds a new Hotel Reservation benchmark problem: `cpu_throttling_hotel_reservation`.

The scenario is based on a production incident at Omio where aggressively low CPU limits caused silent latency degradation that went undetected for months. There was also a known Kubernetes issue (https://github.com/kubernetes/kubernetes/issues/67577) where containers were being throttled far more aggressively than their configured limits due to a kernel bug, since fixed in Linux 5.4. The general throttling problem with tight CPU limits still remains for multi-threaded workloads and is the cause of much debate in the Kubernetes community around whether CPU limits should be used at all. The interesting part is that `kubectl top` shows CPU usage well below the limit, so the diagnosis direction is "CPU looks fine, but why is the service slow?" which is the opposite of the existing `astronomy_shop_ad_service_high_cpu` and `edge_request_filter_cpu_saturation` problems that show artificially high CPU usage.

References:
- https://medium.com/omio-engineering/cpu-limits-and-aggressive-throttling-in-kubernetes-c5b20bd8a718
- https://home.robusta.dev/blog/stop-using-cpu-limits
- https://github.com/kubernetes/kubernetes/issues/67577

## Expected System Behavior

After fault injection, `geo` runs with `limits.cpu: 50m`, all pods stay Running with no restarts, and `kubectl top` shows CPU well below the limit, looks healthy. The real signal is in `kubectl describe deployment geo` (`Limits: cpu: 50m`) and `kubectl exec deploy/geo -- cat /sys/fs/cgroup/cpu/cpu.stat` (`nr_throttled` climbing). The Prometheus `ContainerCPUThrottling` alert also fires since `hotel-reservation` is already in the alert selector.

## Mitigation

Open to suggestions. Raising the limit or removing it entirely both work. The oracle passes if `limits.cpu > 100m` or if the limit is removed.

## Validation

```bash
uv run python tests/integration/validate_problem.py --problem cpu_throttling_hotel_reservation
```

/validate-problem cpu_throttling_hotel_reservation


tianyin:
@varuni7 I love this problem, as someone worked on CPU management. The problem you describe is very real. @pratiksampat did some really interesting research on that (you can read his work https://tianyin.github.io/pub/kscaler.pdf if you're interested). 

> due to a kernel bug, since fixed in Linux 5.4

Given that SREGym does not touch the kernel, that behavior is hard to simulate. It would be an extremely interesting problem tho.

> The general throttling problem with tight CPU limits still remains for multi-threaded workloads and is the cause of much debate in the Kubernetes community around whether CPU limits should be used at all.

:) 

I definitely think limit is necessary, but you are right that it's hard to use it correctly with coase-grained knobs.
 
Thank you for the work!

sharyash81:
@tianyin @varuni7 nice work on this! I was actually looking at another Problem RFC for UIUC++ summer research (other than myself) and came across this one.

This throttling issue is so sneaky — we ran into it two years ago in production too. Figuring out the actual root cause (throttling before hitting the limit) took me a lot of time since it was affecting our production, so I wrote this article breaking it down and covering in-depth root cause: https://medium.com/@sharyash81/solving-cpu-throttling-issue-in-golang-applications-before-hitting-the-cpu-limit-in-kubernetes-7d8f40da6477

Thought you might find it interesting 🙌

Saadmrp1038:
@sharyash81 the blog was a nice read. It will actually help me review this PR :3

sharyash81:
> @sharyash81 the blog was a nice read. It will actually help me review this PR :3

Glad you found it helpful =D

sregym/conductor/problems/cpu_throttling.py
```diff
@@ -0,0 +1,57 @@
...
+            component=f"deployment/{self.faulty_service}",
+            namespace=self.namespace,
+            description=(
+                f"The `{self.faulty_service}` deployment has a CPU limit of {_CPU_LIMIT} millicores, "
+                "which is far too low for its normal operation. The Linux CFS scheduler throttles the "
+                "container whenever it exceeds its quota within a 100ms scheduling window, causing "
+                "severe latency increases on search paths that depend on this service. "
+                "Crucially, `kubectl top pods` shows CPU usage well below the limit — the throttling "
+                "is silent at the kubectl level but visible in the container's cgroup stats "
+                "(`/sys/fs/cgroup/cpu/cpu.stat`: high `nr_throttled`) and in the Prometheus "
```
Saadmrp1038: nit: The cgroup path `/sys/fs/cgroup/cpu/cpu.stat` is cgroup v1. On cgroup v2 clusters (increasingly common; even my local machine is v2) the path is `/sys/fs/cgroup/cpu.stat`. Consider mentioning both or using a generic phrasing like "the container's cgroup cpu.stat".

sregym/conductor/problems/cpu_throttling.py
```diff
@@ -0,0 +1,57 @@
...
+from sregym.service.apps.hotel_reservation import HotelReservation
+from sregym.service.kubectl import KubeCtl
+from sregym.utils.decorators import mark_fault_injected
+
+_CPU_LIMIT = "50m"
+
+
+class CpuThrottling(Problem):
+    def __init__(self, app_name: str = "hotel_reservation", faulty_service: str = "geo"):
+        self.app_name = app_name
```
Saadmrp1038: nit: Having `app_name` param is not needed here since we always create `HotelReservation()` app. We can just remove the param here and hardcode `app_name` to `hotel_reservation`.

sregym/conductor/problems/cpu_throttling.py
Saadmrp1038, line 1: nit: Naming the fault `cfs_cpu_throttling_hotel_reservation` feels more appropiate.

Saadmrp1038 (changes requested):
@varuni7 The code looks good! I left some nit comments below.

During my testing I noticed that after fault injection with 50m cpu limit the CPU usage shown by `kubectl top` is pinned at the limit. Most likely 50m is too low for `geo` to work normally.

```
$ kubectl top pod -n hotel-reservation | grep geo
geo-6b5c578658-xj8sn                      49m          16Mi
```

Almost all cpu cycles get throttled:

```
$ kubectl exec deploy/geo -n hotel-reservation -- cat /sys/fs/cgroup/cpu.stat
usage_usec 10109855
user_usec 8316088
system_usec 1793767
core_sched.force_idle_usec 0
nr_periods 2044
nr_throttled 2044
throttled_usec 3350730038
nr_bursts 0
burst_usec 0
```

To truly get "CPU looks fine, but why is the service slow?" I think a little tweaking is needed to this value. I did some light experimenting and 125m (and around this number) seemed like a good value. The CPU usage has enough
headroom:

```
$ kubectl top pod -n hotel-reservation | grep geo
geo-5678fb8994-xrgkc                      109m          9Mi
```

The value mostly fluctuates around ~105m. That means about 85% utilization. But more importantly a lot of throttling happens:

```
$ kubectl exec deploy/geo -n hotel-reservation -- cat /sys/fs/cgroup/cpu.stat
usage_usec 28628948
user_usec 18324073
system_usec 10304874
core_sched.force_idle_usec 0
nr_periods 2680
nr_throttled 1151
throttled_usec 146224508
nr_bursts 0
burst_usec 0
```

That's ~40% throttle rate. You can also test with a bit larger cpu limit values (130m, 135m etc.) and see which cpu usage/throttle rate is appropriate.

Let me know what you think.

sregym/conductor/problems/cpu_throttling.py
varuni7: hi @Saadmrp1038 , thanks for the review! 

I am running a few experiments and yes, as you mentioned the sweet spot seems to be around 100 to 140 .
Will update with a detailed report

varuni7:
> @varuni7 I love this problem, as someone worked on CPU management. The problem you describe is very real. @pratiksampat did some really interesting research on that (you can read his work https://tianyin.github.io/pub/kscaler.pdf if you're interested).
> 
> > due to a kernel bug, since fixed in Linux 5.4
> 
> Given that SREGym does not touch the kernel, that behavior is hard to simulate. It would be an extremely interesting problem tho.
> 
> > The general throttling problem with tight CPU limits still remains for multi-threaded workloads and is the cause of much debate in the Kubernetes community around whether CPU limits should be used at all.
> 
> :)
> 
> I definitely think limit is necessary, but you are right that it's hard to use it correctly with coase-grained knobs.
> 
> Thank you for the work!

thanks  for the review @tianyin !

Loved reading the paper, directly leveraging kernel-level metrics is interesting , one question

If we pack hundreds of containers onto a single server, does tracking all that history inside the kernel eventually bog down the CPU scheduler?
I was wondering what's the breaking point where the monitoring overhead starts hurting performance

> Given that SREGym does not touch the kernel, that behavior is hard to simulate. It would be an extremely interesting problem tho.

agreed, I mentioned the kernel bug just as a reference/context  as it was interesting to me, we hit into this in a lot of otherwise production scenarios as well 

> I definitely think limit is necessary, but you are right that it's hard to use it correctly with coase-grained knobs.

true XD 

I came across tools which do a workload prediction and tune the cpu limits accordingly , https://github.com/robusta-dev/krr , these recommenders seem interesting!

pratiksampat:
> Loved reading the paper, directly leveraging kernel-level metrics is interesting , one question
> 
> If we pack hundreds of containers onto a single server, does tracking all that history inside the kernel eventually bog down the CPU scheduler? I was wondering what's the breaking point where the monitoring overhead starts hurting performance

That's a good question. The observation loop is a few hundred CPU cycles at most since all it does is get the time, and append an entry for the runqueue. The lightness of it all is what allows it to be in-kernel and keep things scalable. While, I have stress-tested this without things breaking, it still might warrant a deeper study in the future.

The decision/recommendation loop is what takes the most time instead. Since the data we now extract is richer and the history smaller, we get away with using simple statistics for recommending quota-period inside the kernel while being extremely fast while doing so. However, I'm sure there may be more accurate ways to models this in userspace but at the cost of reactivity when this data is extracted via eBPF/trace buffers.

varuni7:
| limit | kubectl top | nr_periods | nr_throttled | throttle_pct | throttled_usec |
|-------|-------------|------------|--------------|--------------|----------------|
| 50m   | 48m         | 486        | 412          | 84.7%        | 37,249,146     |
| 75m   | 68m         | 464        | 170          | 36.6%        | 5,722,757      |
| 100m  | 81m         | 474        | 106          | 22.3%        | 4,765,373      |
| 110m  | 82m         | 466        | 74           | 15.8%        | 3,755,126      |
| 125m  | 82m         | 480        | 69           | 14.3%        | 4,449,633      |
| 130m  | 82m         | 472        | 33           | 6.9%         | 1,468,273      |
| 135m  | 79m         | 471        | 11           | 2.3%         | 432,881        |
| 150m  | 77m         | 6844       | 297          | 4.3%         | 18,425,251     |
| 200m  | 72m         | 526        | 9            | 1.7%         | 568,095        |

going with 125m

varuni7:
@Saadmrp1038 I've addressed the reviews about the app_name and cgroup path 
please let me know if any other changes are required

Saadmrp1038 (changes requested):
@varuni7 It seems the results vary based on the type of device it is being run on. 

For my cloudlab clusters, 125m resulted in a ~40% throttle rate. 
For your setup, it only caused a 15% throttle rate, which is not much tbh. 

Well, it does make sense that 125m of different cores will not show the same performance. I think we might need a dynamic mechanism here that will try to balance the CPU usage and throttle rate to maximize throttle rate without capping CPU usage.

varuni7:
@Saadmrp1038 how about this 
I first measure geo's baseline cpu demand on the current hardware, then binary search for a limit that lands in a target throttle rate band (say 40-45%). cache the result per cluster so this calibration only runs once. The way, th limit ends up being relative to actual cpu performance rather than a hardcoded value.

Saadmrp1038:
@varuni7 Sounds about right. No need to hard target throttle ~40-45%. Try to balance so as not to cap the CPU usage to the limit. Otherwise it will be easy for the agent to detect.

Saadmrp1038:
@varuni7 I am reviewing the PR right now. One question -- what throttle rate do you get with this new mechanism?

sregym/conductor/problems/registry.py
```diff
@@ -130,6 +131,7 @@ def __init__(self):
             "assign_to_non_existent_node": AssignNonExistentNode,
             "auth_miss_mongodb": MongoDBAuthMissing,
             "configmap_drift_hotel_reservation": lambda: ConfigMapDrift(faulty_service="geo"),
+            "cpu_throttling_hotel_reservation": lambda: CpuThrottling(faulty_service="geo"),
```
Saadmrp1038: nit: We should name the fault `cfs_cpu_throttling_hotel_reservation`.

sregym/conductor/oracles/cpu_throttling_mitigation.py
Saadmrp1038, line 1: With the new calibration mechanism this oracle doesn't work anymore. The calibration on my cluster produces limit=135m. But the oracle uses a hardcoded `_MIN_ACCEPTABLE_CPU_MILLICORES = 100`. Since 135m > 100m, the oracle passes without any fix.

sregym/conductor/oracles/cpu_throttling_mitigation.py
```diff
@@ -0,0 +1,94 @@
...
+            print(f"Deployment '{self.faulty_service}' scaled to {desired} — scaling to 0 is not a valid fix")
+            return {"success": False}
+
+        containers = deployment.spec.template.spec.containers
+        for container in containers:
+            limits = (container.resources and container.resources.limits) or {}
+            cpu_limit_str = limits.get("cpu") if isinstance(limits, dict) else getattr(limits, "cpu", None)
+            if cpu_limit_str is None:
+                print(f"Container '{container.name}' has no CPU limit, removing it is NOT a valid fix")
+                return {"success": False}
```
Saadmrp1038: Feels like removing the `cpu_limit` should be a valid fix.

Saadmrp1038 (changes requested):

sregym/conductor/oracles/cpu_throttling_mitigation.py
varuni7: on what throttle rate do you get with this new mechanism?

around 30% throttle , with a 40m limit

sregym/conductor/oracles/cpu_throttling_mitigation.py
varuni7: ah I see, my env always gave limit between 30 - 60 m , so wasnt able to catch this

sregym/conductor/oracles/cpu_throttling_mitigation.py
varuni7: Feels like removing the cpu_limit should be a valid fix.

I had removed it because of https://github.com/SREGym/SREGym/pull/807#issuecomment-4567717847

should I add it back ?

Saadmrp1038: > around 30% throttle , with a 40m limit

Sounds good enough. Your 40m vs my 135m -- there's quite a huge difference!

> I had removed it because of https://github.com/SREGym/SREGym/pull/807#issuecomment-4567717847

For the purpose of this fault, removing the limit feels like quite a logical mitigation. Instead of guessing at what CPU limit will bring the throttle rate down, the agent just removes it. Might not be the ideal mitigation, but still reasonable.

varuni7: fair , done! added back

Saadmrp1038 (commented):
@varuni7 The calibration is working nicely.

During testing with the agent I found one problem. The agent never looks up the throttle rate and simply reasons that since the cpu limit of `geo` service is much lower compared to all its peers (who have 1000m), it raises the `geo` service's cpu limit to 1000m to mitigate the fault. Basically instead of investigating properly the agent is simply pattern matching and submitting a reasonable guess.

To make the fault more believable and not make `geo` stand out so much, the calibration should set cpu limit values for other services too.

What should be done:
1. Calibration should be done for all services
2. For `geo` the calibration sets a value that doesn't pin cpu usage to the limit but still causes significant throttling (~25-40%)
3. For other services the calibration sets a value with more headroom so they don't throttle meaningfully (<10%)

For example, geo gets `baseline * 1.15` headroom (triggers throttling) while other services could get `baseline * 1.5` or `baseline * 2.0` (no meaningful throttling). That way all services have varied realistic limits, and the only way to pinpoint geo as the problem is to actually check throttle rates via cgroup stats or Prometheus and not just compare deployment specs.

With so many services getting new limits, a verification step after injection would be good. Sample each service again after the limits are applied, confirm that the cpu usage and throttle rate of all the services is within desired range. This prevents the calibration from creating false signals that can confuse the agent.

---

Without this I think the fault might be too easy. Over multiple runs the agent always succeeded in both diagnosis and mitigation by simply guessing based on pattern matching.

Let me know what you think.

varuni7:
`To make the fault more believable and not make geo stand out so much, the calibration should set cpu limit values for other services too.
`
agreed , addressed in the latest commit

varuni7:
will add the verification step next , thinking of 

after injecting check nr_throttled/nr_periods from cpu.stat and assert 

- faulty service throttle rate is above some minimum (e.g  >5%)  
- non-faulty services are below some maximum

Saadmrp1038:
@varuni7 I think the verification step is needed. During testing, after fault injection jaeger was at 80.8% throttle rate! Recreating all deployments likely put extra pressure on jaeger and its CPU usage exceeded the limit (limit 35m, usage 41m). As a result, it got heavily throttled.

sregym/generators/fault/inject_virtual.py
```diff
@@ -343,6 +345,307 @@ def recover_resource_request(self, microservices: list[str]):
...
+            for container in containers:
+                resources = container.setdefault("resources", {})
+                if service in microservices:
+                    resources.setdefault("requests", {})["cpu"] = "10m"
+                else:
+                    resources.setdefault("requests", {})["cpu"] = limit
+                resources.setdefault("limits", {})["cpu"] = limit
+            modified_yaml_path = self._write_yaml_to_file(service, deployment_yaml)
+            self.kubectl.exec_command(f"kubectl delete deployment {service} -n {self.namespace}")
+            self.kubectl.exec_command(f"kubectl apply -f {modified_yaml_path} -n {self.namespace}")
```
Saadmrp1038: Let's use `kubectl patch` instead of delete+apply here. Gives a more smoother transition. As deleting and recreating all services can cause the startup hiccups again (pods restarting a couple of times before stabilizing etc.).

Saadmrp1038 (changes requested):
@varuni7 Sorry for the delayed review. I tested with the new calibration system. It is working really great!
I didn't find any major issues yet. I will do some agent runs to see how it goes. If everythings alright I will merge the PR.

Saadmrp1038 (approved):

PR #813, opened 2026-05-29
Fix `k8s_proxy` hanging on parallel kubectl requests

Saadmrp1038:
The k8s filtering proxy uses Python's `HTTPServer` which is single-threaded and processes one request at a time. When an agent fires multiple kubectl commands in parallel, requests queue up and eventually hang/timeout.

Replacing `HTTPServer` with `ThreadingHTTPServer` seems to be working.

Thanks to @selen0phile for reporting this!

### DIFF
--- a/sregym/service/k8s_proxy.py
+++ b/sregym/service/k8s_proxy.py
@@ -22,7 +22,7 @@ import os
 import ssl
 import tempfile
 import threading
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
@selen0phile Can you please test if this works now?

tianyin (approved):
LGTM

PR #814, opened 2026-05-29
Add cronjob_sidecar_blocks_completion_hotel_reservation problem

selen0phile:
# Add cronjob_sidecar_blocks_completion_hotel_reservation problem

## 1. The failure

A Kubernetes Job is only marked `Complete` when every container in the pod terminates. That's a problem when the pod has a sidecar designed for long-running service work, like a service-mesh proxy (`istio-proxy`, `linkerd-proxy`), a telemetry sidecar (`fluent-bit`, `fluentd`, `otel-collector`, `datadog-agent`), or a credential broker (`vault-agent`, Cloud SQL Proxy). The sidecar never receives a termination signal when the primary container exits, so the pod stays `Running` indefinitely and the Job stays `active`. Because `successfulJobsHistoryLimit` only applies to *finished* Jobs, the accumulated active Jobs are not reaped. Over hours the namespace fills up with hundreds of stuck Jobs and Pods.

The Kubernetes project itself documented this and it was significant enough to motivate native sidecar containers in Kubernetes 1.28 ([KEP-753](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/753-sidecar-containers/README.md)). The legacy pattern is still what the majority of production manifests, Helm charts, and operator-managed deployments use, so the bug continues to reproduce exactly as reported in [kubernetes/kubernetes#64056](https://github.com/kubernetes/kubernetes/issues/64056).

### Sources

- [Kubernetes blog (2023-08-25): "Introducing native sidecar containers"](https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/). Quote: *"in the case of a Job, if the main container finishes, but the sidecar container continues running, it prevents the Job from correctly determining whether the Pod has terminated successfully."*
- [KEP-753](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/753-sidecar-containers/README.md), the enhancement proposal.
- [kubernetes/kubernetes#64056](https://github.com/kubernetes/kubernetes/issues/64056). Production user with a fluentd sidecar, confirmed root cause and fix.
- [istio/istio#11045](https://github.com/istio/istio/issues/11045), literally titled "Jobs are Broken (they Never Complete)".
- [Linkerd blog: "The Proxy Died First"](https://linkerd.io/2026/05/18/the-proxy-died-first/).
- [Red Hat Customer Portal #6280111](https://access.redhat.com/solutions/6280111).
- [TeamSnap Engineering: "Properly Running Kubernetes Jobs with Sidecars"](https://medium.com/teamsnap-engineering/properly-running-kubernetes-jobs-with-sidecars-ddc04685d0dc).

## 2. Simulation

- **App:** Hotel Reservation
- **Injection layer:** Direct K8s API. No Helm changes, no node access.

A CronJob named `audit-log-archiver` is deployed into the Hotel Reservation namespace. Its pod template contains two regular containers, which is the legacy, pre-KEP-753 pattern:

- **Primary container `archiver`**: simulates a realistic audit-log archival step. Logs progress, sleeps 1 second, exits 0.
- **Sidecar container `fluent-bit-sidecar`**: simulates a fluent-bit log forwarder bound to the canonical fluent-bit forward port. Its command is `exec nc -lk -p 24224 -e cat`. This is a network listener daemon, not a visible shell loop. An agent reading the spec sees "a network service listening on port 24224", exactly what a real fluent-bit binary looks like in production. Patching the sidecar's command to `sleep 10` would obviously break the listener, so the agent has to apply a structural fix to the lifecycle rather than a workaround on the container command.

Schedule is `* * * * *`. Each scheduled Job runs the primary, the primary exits, and the sidecar keeps the pod alive forever. After three to four cycles the symptom is unambiguous: several Jobs stuck at `COMPLETIONS: 0/1` whose pods show `archiver` Terminated/Completed and `fluent-bit-sidecar` Running.

`inject_fault` returns only once at least two Jobs from the CronJob exist, so the agent always sees the accumulation pattern.

### How `recover_fault` works

`recover_fault` applies the same fix the oracle requires from the agent: it deletes the broken CronJob and recreates it with the sidecar moved to `initContainers` and `restartPolicy: Always` set. Then it deletes the pre-recovery accumulated Jobs by name (the ones snapshotted before the fix went in), so legitimate new Jobs the now-fixed CronJob spawns are left alone. Idempotent against an already-recovered state.

### Oracle policy

The mitigation oracle accepts exactly one fix and rejects everything else.

**Accepted:** convert the sidecar to the K8s 1.28+ native sidecar pattern. Move the sidecar container from `spec.jobTemplate.spec.template.spec.containers` into `spec.jobTemplate.spec.template.spec.initContainers`, and set `restartPolicy: Always` on it. The kubelet then SIGTERMs the sidecar after the primary container exits, and the pod reaches `Succeeded`. This is the architecturally correct fix per KEP-753, and the only one that fixes the lifecycle bug without putting a time bomb on the workload or deleting its functional purpose.

**Rejected:**

| Workaround | Why rejected |
|---|---|
| Add `activeDeadlineSeconds` to the jobTemplate | Time-bombs the workload rather than fixing the lifecycle bug. Real archival runs that exceed the deadline get killed mid-flight. |
| Remove the sidecar from the jobTemplate | Silently deletes audit-log forwarding to the SIEM. |
| Delete the CronJob | Removes the workload, not the bug. |

The oracle does both spec-level and runtime validation:

1. **Spec check:** `spec.jobTemplate.spec.template.spec.initContainers` has at least one entry with `restartPolicy: Always`. Fails fast on the wrong shape.
2. **Active Jobs bounded:** no more than 2 Jobs owned by the CronJob are currently `active`. Catches the case where the spec was fixed but the accumulated Jobs weren't cleaned up.
3. **App still healthy:** every Deployment in the namespace reports `ready_replicas == spec.replicas`. Catches collateral damage.
4. **Runtime proof:** at least one Job owned by the now-fixed CronJob has reached `status.succeeded >= 1` within 150s. This is what makes "spec looks right" different from "fix actually works", a wrong patch shape (sidecar moved without `restartPolicy`, or `restartPolicy` on the wrong container) will still pass step 1 but fail step 4 because the pod never reaches Succeeded.

When the oracle rejects, the message names the accepted shape so an agent reading the verdict can pivot.

## 3. Runtime behavior

```
$ kubectl get cronjob -n hotel-reservation audit-log-archiver
NAME                 SCHEDULE    SUSPEND   ACTIVE   LAST SCHEDULE   AGE
audit-log-archiver   * * * * *   False     4        15s             4m12s

$ kubectl get jobs -n hotel-reservation -l app.kubernetes.io/name=audit-log-archiver
NAME                          STATUS    COMPLETIONS   DURATION   AGE
audit-log-archiver-29666848   Running   0/1           4m         4m
audit-log-archiver-29666849   Running   0/1           3m         3m
audit-log-archiver-29666850   Running   0/1           2m         2m
audit-log-archiver-29666851   Running   0/1           1m         1m
```

The Job-owned pods show the smoking gun in their container status:

```
$ kubectl get pod -n hotel-reservation <pod> -o json | jq '.status.containerStatuses[] | {name, state, ready}'
{ "name": "archiver",           "state": { "terminated": { "reason": "Completed", "exitCode": 0 } }, "ready": false }
{ "name": "fluent-bit-sidecar", "state": { "running":    { "startedAt": "..." } },                  "ready": true  }
```

Primary completed, sidecar still running, Job stays `active`, CronJob keeps creating more.

### Unbounded growth

`schedule: "* * * * *"` produces one new active Job per minute, indefinitely. Because the Jobs are stuck in `active` phase (and `successfulJobsHistoryLimit: 3` / `failedJobsHistoryLimit: 1` only reap finished Jobs, which ours never become), the active set grows without bound: about 60 Jobs/hour, 1,440 Jobs/day, 43,000 Jobs/month if left running. This is why the cited postmortems describe the failure as a slow-motion incident rather than an instant outage. The cluster looks fine for hours or days, then crosses the etcd-degradation / pod-CIDR-exhaustion / apiserver-LIST-overload threshold and falls over all at once.

## 4. Agent behavior

Two frontier agents were tested end-to-end against this problem.

| Agent | Diagnosis | TTL | Mitigation outcome | TTM |
|---|---|---|---|---|
| Stratus (gpt-5) | 100/100 | 117s | Removed the sidecar from the jobTemplate (JSON patch). Oracle rejected with the pivot message; agent did not pivot to the native sidecar pattern within its step budget. | 298s |
| Claude Code (sonnet-4-6) | 89/100 | n/a | First move was the same: remove the sidecar from the jobTemplate. | n/a |

### Stratus diagnosis (verbatim, 100/100)

> *"Root cause: The audit-log-archiver CronJob is misconfigured so its Jobs never complete. Each Job's Pod has two containers: the main archiver (busybox) finishes and exits 0 immediately, but the fluent-bit-sidecar container runs `nc -lk -p 24224` indefinitely. Because at least one container remains running, the Pod never terminates and the Job never reaches the required completion. With the CronJob scheduled every minute and concurrencyPolicy=Allow, new Jobs keep being created while previous Jobs remain Active, resulting in multiple NotReady audit-log-archiver Pods and accumulating Active Jobs (e.g., audit-log-archiver-29667608/09/10 show 1 Active, 0 Succeeded). Evidence: `kubectl describe pod` shows archiver container Terminated (Completed) and sidecar Running; `kubectl describe job` shows 1 Active / 0 Succeeded / 0 Failed; `kubectl describe cronjob` shows multiple Active Jobs. The persistent fault is the CronJob's sidecar preventing Job completion."*

Stratus quoted the literal sidecar command (`nc -lk -p 24224`) and the actual Job names from the cluster, so it read the spec carefully rather than pattern-matching on a shell loop. Judge checklist scored 1.0 on Fault Localization, Fault Characterization, and Scope Precision.

### Claude Code diagnosis (verbatim, 89/100)

> *"A CronJob in the hotel-reservation namespace has a sidecar container that blocks job completion. In Kubernetes Jobs and CronJobs, all containers including sidecars must exit for the Job to be considered complete. The sidecar container (likely a service mesh proxy or monitoring agent) keeps running indefinitely after the main container finishes, preventing the CronJob from ever reaching a Completed state. This causes Jobs to accumulate and the CronJob history to fill with perpetually-running jobs..."*

Different reasoning chain, same conclusion. Claude Code described the sidecar generically ("likely a service mesh proxy or monitoring agent") rather than reading the literal command, which is why D2 Fault Characterization is lower (0.67) while Fault Localization (1.0) and Scope Precision (1.0) are full credit. The 89 vs 100 gap is meaningful, a copied answer would have scored 100. Two agents, two independent reasoning chains, both correctly identify the bug from kubectl observations alone.

### Mitigation lands on "remove the sidecar"

Both agents' first kubectl patch was the same: remove the sidecar container from the jobTemplate. That's the lowest-friction operation that resolves the symptom, and it's exactly the shape the oracle is designed to reject. The oracle returns the pivot message naming the accepted shape (`initContainers` + `restartPolicy: Always`), which is what gives a capable agent a path forward. Stratus's actual patch (JSON 6902):

```bash
kubectl -n hotel-reservation patch cronjob audit-log-archiver --type=json -p '[
  {"op":"add",   "path":"/spec/suspend","value":true},
  {"op":"remove","path":"/spec/jobTemplate/spec/template/spec/containers/1"},
  {"op":"add",   "path":"/spec/jobTemplate/spec/ttlSecondsAfterFinished","value":300}
]'
kubectl -n hotel-reservation delete jobs -l app.kubernetes.io/name=audit-log-archiver
kubectl -n hotel-reservation patch cronjob audit-log-archiver --type=merge -p '{"spec":{"suspend":false}}'
```

Rejected with (verbatim from the agent's `/submit` response):

> *"CronJob 'audit-log-archiver' is not in an acceptable post-fix state. The only accepted fix for this lifecycle bug is the K8s 1.28+ native sidecar pattern: move the sidecar container from spec.jobTemplate.spec.template.spec.containers to spec.jobTemplate.spec.template.spec.initContainers, and set restartPolicy=Always on it. activeDeadlineSeconds and removing the sidecar are not accepted -- they either time-bomb the workload or delete its functional purpose."*

Stratus made two patch attempts, both Path 2 variants, and did not pivot to the native sidecar shape before TTM expired. A stronger agent reads the rejection and pivots. The diagnosis half of this problem is solvable today, the mitigation half is the harder bar.

## Testing

- [x] `tests/integration/validate_problem.py --problem cronjob_sidecar_blocks_completion_hotel_reservation` passes locally on a kind cluster (K8s 1.32, Calico CNI):

  ```
  ================================================================
  PROBLEM VALIDATION — cronjob_sidecar_blocks_completion_hotel_reservation
  ================================================================
    ✅  Resolve problem in registry: CronJobSidecarBlocksCompletionHotelReservation · app `Hotel Reservation`
    ✅  Deploy application: `Hotel Reservation` deployed to namespace `hotel-reservation`
    ✅  Inject fault: inject_fault() completed without error
    ✅  Oracle fails after fault injection: oracle reported failure after 1 check(s)
    ✅  Recover fault: recover_fault() completed without error
    ✅  Oracle passes after recovery: oracle reported success after 1 check(s)
  ----------------------------------------------------------------
    Result: PASSED  (210s)
  ================================================================
  ```

- [x] Oracle rejects each of the following:

  | Attempted fix | Verdict | Reason |
  |---|---|---|
  | Delete owned Jobs, spec untouched | ❌ | Spec still has the lifecycle trap |
  | `kubectl patch cronjob -p '{"spec":{"suspend":true}}'` | ❌ | Doesn't address the lifecycle bug |
  | Add `activeDeadlineSeconds=30` to jobTemplate | ❌ | Time-bombs the workload, doesn't fix the lifecycle |
  | Remove the sidecar container | ❌ | Deletes the sidecar's functional purpose |
  | `kubectl delete cronjob audit-log-archiver` | ❌ | Removes the workload, not the bug |

- [x] Oracle accepts the native sidecar fix:

  | Fix | Verdict | Verification |
  |---|---|---|
  | Move sidecar to `initContainers` with `restartPolicy: Always`, plus delete accumulated Jobs | ✅ | Spec check, active Jobs ≤ 2, app healthy, plus runtime proof: a Job from the now-fixed CronJob reaches `status.succeeded >= 1` within 150s |

- [x] Re-injection idempotency. `inject_fault` defensively deletes any leftover CronJob.
- [x] Recovery idempotency. `recover_fault` returns cleanly against an already-recovered state.
- [x] No `sregym` strings in cluster-applied manifests (`grep -rE "sregym|CronJobSidecar"` returns 0).

/validate-problem cronjob_sidecar_blocks_completion_hotel_reservation


Saadmrp1038:
@selen0phile Thanks for the PR! I will review it soon.

sregym/conductor/oracles/cronjob_sidecar_mitigation.py
Saadmrp1038, line 1: nit: We have ruff setup for this repo. You can check #484 for how to set it up locally. Will help catch simple lint errors and import ordering issues.

Saadmrp1038 (changes requested):
@selen0phile This is a _very_ detailed and well-tested PR! Thanks for taking the time and being thorough. Everything works perfectly. 

There are just some minor lint errors. Ping me after you address them. I will merge the PR.

sregym/conductor/oracles/cronjob_sidecar_mitigation.py
selen0phile: fixed the linting issues, can you check if it's okay? @Saadmrp1038

Saadmrp1038 (approved):

Saadmrp1038:
@selen0phile Congrats on your first PR merged here 🎉
Keep up the good work!

---

PR #818, opened 2026-05-29
Add node_routing_table_wipeout problem

keremakcali:
Adds a new SREGym problem simulating the removal of per-pod kernel routes from a single worker node's routing table, using Hotel Reservation as the workload.

The motivating incident is the Datadog March 8, 2023 outage. A systemd security patch on Ubuntu 22.04 restarted `systemd-networkd`, which flushed IP routing rules that Cilium had installed on the affected nodes. Pods on those nodes lost network connectivity, and the nodes themselves became unreachable. About 60% of Datadog's instances across every region were affected.
Reference: https://www.datadoghq.com/blog/engineering/2023-03-08-deep-dive-into-platform-level-impact/

This is a simplified analog of the Datadog scenario. On a single worker node it deletes the per-pod host routes (`10.244.x.y dev vethXXX scope link`) that the CNI writes at pod creation to deliver traffic to each local pod.

The fault leaves inter-node pod-CIDR routes (`10.244.x.0/24 via <node-ip>`) alone. The CNI's reconciler periodically reinstalls these inter-node routes if they go missing, so flushing them too would converge to the same end state within seconds. Per-pod host routes are not part of that reconcile path — they are only rewritten when a pod is created.

The fault is implemented as a new problem class `NodeRoutingTableWipeout` at `sregym/conductor/problems/node_routing_table_wipeout.py` and registered in `registry.py`. The route flush itself is exposed as a new `flush_pod_routes(node)` method on `RemoteOSFaultInjector`, which runs:

```
ip route show | awk '/^10\.244\.[0-9]+\.[0-9]+ dev veth/ {print $1}' | xargs -r -n 1 ip route del
```

on the target node.

Victim node selection intersects worker nodes with nodes that have hotel-reservation pods; this guarantees `ServiceEndpointDown` fires deterministically. Recovery deletes every pod on the victim node across every affected namespace — not just hotel-reservation — because the route flush is node-scoped and any pod that was on the node when the routes were deleted needs its veth route reinstalled by the CNI at recreation.

Diagnosis uses `LLMAsAJudgeOracle` with a structured root cause string identifying the kernel routing table on the affected node as the origin of the failure. Mitigation uses `AlertOracle` against the existing `ServiceEndpointDown` rule, which already covers hotel-reservation.

## The problem runtime behavior

Verified the behavior on a 4-node kind cluster (1 control-plane, 3 workers). The fault is injected simply by executing the command above on the victim node. The victim node stays `Ready`. `ip route show` on the affected node confirms the missing veth entries. Pods on the victim node cannot complete TCP connections in either direction. Pods that don't depend on other services over the network stay `Running`/`Ready`; pods that do can panic or enter `CrashLoopBackOff`. Within about a minute, `ServiceEndpointDown` fires for every Service backed by a pod on the victim node. Cross-node monitoring may surface incidental noise in other namespaces.

## The agent behavior

Tested with Stratus on Claude Haiku 4.5. The agent diagnosed startup race conditions between services and MongoDB, blaming missing readiness probes and dependency ordering. In the mitigation stage the agent spent a lot of time deleting specific MongoDB pods that happened not to be on the victim node, without broadening to a node-wide restart.

The agent followed the most visible signals — `CrashLoopBackOff` on the reservation pod and MongoDB connection panics in pod logs — and never traced the failure down to the node or routing layer. This may reflect a difficulty in the problem: the visible symptoms point at the application while the fault sits below the Kubernetes abstraction.

/validate-problem node_routing_table_wipeout


Saadmrp1038:
@keremakcali Thanks for the PR! Might take a few days to review this.

sregym/generators/fault/inject_remote_os.py
```diff
@@ -189,6 +189,27 @@ def recover_kubelet_crash(self):
 
         self._wait_for_worker_nodes("Ready")
 
+    def flush_pod_routes(self, node: str):
+        """Delete per-pod host routes from the kernel routing table on a specific worker node."""
+        cmd = "ip route show | awk '/^10\\.244\\.[0-9]+\\.[0-9]+ dev veth/ {print $1}' | xargs -r -n 1 ip route del"
```
Saadmrp1038, line 218: We would need to use `sudo` to make it work on real cluster.

sregym/generators/fault/inject_remote_os.py
```diff
@@ -189,6 +189,27 @@ def recover_kubelet_crash(self):
...
+            worker_info = self._get_remote_worker_info()
+            if not worker_info:
+                return
+            if node not in worker_info:
+                print(f"Node {node} not found in inventory")
+                return
+            user = worker_info[node]
+            print(f"Flushing pod routes on {node}...")
+            self._ssh_exec(node, user, cmd)
+            print(f"Pod routes flushed on {node}")
```
Saadmrp1038, line 235: Tested on a real cluster. The method takes a Kubernetes node name (e.g. `node2.exp-242212.aiopslab-pg0.wisc.cloudlab.us`) and looks it up in `worker_info`, which is keyed by ansible SSH hostnames (e.g. `c220g5-120130.wisc.cloudlab.us`). These never match, so the method prints "Node not found in inventory" and returns without doing anything. Check what faults like `inject_kubelet_crash` and `pod_cidr_exhaustion_hotel_reservation` do.

Saadmrp1038 (changes requested):
@keremakcali Thanks for the effort! Since you tested on kind, there are some bugs in real cluster setup. I left some inline comments below.

I think a **major mismatch** happened here. A few days ago we changed our kind setup to Calico CNI (#774) replacing the default kindnet (which is not really something you would use in production and lacks a lot of features). I think you tested this fault on the old kind setup.  

As it is now the fault does not work on new Calico based setup. It only works for the previous bridge CNI (kindnet). Calico's Felix agent actively monitors and re-adds per-pod routes via its [RouteRefreshInterval](https://docs.tigera.io/calico/latest/reference/felix/configuration) (default 90s).

Bottom line is, I am not sure if this fault is feasible in a Calico based setup. You might need to do some more testing.

---

Saadmrp1038:
@keremakcali when you created the RFC we did not have Calico in kind :)
Can you please test if it's possible to implement this fault in the new setup?

You would need to clean up you old kind cluster and create a new one with the [setup_kind_cluster.sh](https://github.com/SREGym/SREGym/blob/main/kind/setup_kind_cluster.sh) script.

keremakcali:
@Saadmrp1038 thanks a lot for the review! I came to realize this issue today as well. I should have done more research on how other CNIs handle this problem before submitting. I will do some testing as soon as I can and hopefully figure something out :")

Saadmrp1038:
@keremakcali feel free to pivot to some other faults if you feel stuck on this one after trying.

keremakcali:
Upon testing I decided not to go forward with this problem and worked on a different fault class instead. The new PR is: #847. 
Closing this now.

Issue #812, opened 2026-05-28
[Problem RFC] HPA Control Loop Fails Due to Missing CPU Requests


MABDz:
Hi! This is an RFC I’ve written for a new SREGym benchmark problem as part of my evaluation task for UIUC++ SRSE. It primarily focuses on a Kubernetes HorizontalPodAutoscaler (HPA) control-loop failure. I’ll try to explain the problem I’ve tried to model via a scenario. Consider an HPA which is correctly configured for a user-facing service, but its target pods are missing an effective CPU request. Because a CPU-based HPA calculates usage as a percentage of the pod’s request, it can't compute the metric ([source](https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/)). The HPA drops into a broken state, reporting `<unknown>/60%`, `ScalingActive=False`, and`FailedGetResourceMetric`.

The problem itself is based on a real Kubernetes autoscaling failure mode. [This AWS re:Post thread](https://repost.aws/questions/QUXy4-A8bmQN-6aoiWpX4tcA/eks-metrics-failed-to-get-cpu-utilization) introduced me to the existence of the problem. From here I was then able to find a bunch of other posts and documentation referencing variations of the same issue, some of which I have attached below:
1.  https://github.com/kubernetes/kubernetes/issues/95237
2.  https://docs.vertica.com/26.2.x/en/containerized/troubleshooting/vas/
3.  https://github.com/dapr/dapr/issues/7813

For the SREGym implementation, I’m thinking about using the Hotel Reservation app from the Quiz, specifically targeting its frontend deployment. The fault injection here would deploy a standard HPA (e.g., `frontend-capacity`), and then remove both `requests.cpu` and `limits.cpu` from the frontend container spec. I’m phrasing this as “effective CPU request” because in Kubernetes, if a CPU limit remains while the CPU request is removed, the admitted pod can still end up with an effective CPU request equal to the limit, which would prevent the intended HPA failure from appearing as my testing revealed. I’ve manually validated this on the `hotel-reservation/frontend` service. Removing the effective CPU request reliably triggers the `<unknown>/60%` failure state, and putting them back immediately restores the HPA to `ValidMetricFound`.

Thus a valid mitigation must restore a computable HPA metric. A hacky workaround like manually scaling the deployment shouldn't pass, as it leaves the autoscaler completely broken.

[...]

PR #820, opened 2026-05-30
New Problem | HPA missing effective CPU request 

MABDz:
This PR adds a new SREGym problem, `hpa_missing_effective_cpu_request_hotel_reservation` as part of my evaluation task for UIUC++ SRSE.

It models a HorizontalPodAutoscaler control-loop failure on the Hotel Reservation app. The `frontend` deployment stays Running/Ready, but its HPA can't compute CPU utilization because the target pods have no effective CPU request, so it reports `<unknown>/60%` with `ScalingActive=False` / `FailedGetResourceMetric`. The interesting part for an agent is that `kubectl get pods` looks completely healthy, the broken loop is only visible if you go look at the autoscaler itself.

Files:

```text
Problem List.md
sregym/conductor/problems/hpa_missing_effective_cpu_request.py
sregym/conductor/oracles/hpa_control_plane_mitigation.py
sregym/conductor/problems/registry.py
```

Related RFC: #812

/validate-problem hpa_missing_effective_cpu_request_hotel_reservation

## 1. Real-world failure story

A CPU-based HPA computes utilization as a percentage of each pod's CPU *request*. If the target pods have no effective CPU request, there's nothing to divide against, so the autoscaler can't make a decision and parks in a failed-metric state. The trap is that nothing crashes: the workload keeps serving, the HPA object exists, and the cluster looks fine until load actually arrives and the service never scales.

I first ran into this in an [AWS re:Post thread](https://repost.aws/questions/QUXy4-A8bmQN-6aoiWpX4tcA/eks-metrics-failed-to-get-cpu-utilization), then found the same `FailedGetResourceMetric` / missing-CPU-request condition reported across unrelated stacks (a Dapr sidecar case, and kubernetes/kubernetes#95237 where a duplicate container name produced it), which is what convinced me it's a recurring class rather than a one-off.

References:

* Kubernetes HPA docs (the request-relative computation): https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/
* GKE HPA troubleshooting (the operational symptom): https://docs.cloud.google.com/kubernetes-engine/docs/troubleshooting/horizontal-pod-autoscaling
* kubernetes/kubernetes 95237 (the exact failed-metric condition): https://github.com/kubernetes/kubernetes/issues/95237
* Kubernetes requests/limits docs: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

The problem targets the Hotel Reservation app in the `hotel-reservation` namespace, specifically:

```text
Deployment/frontend
HorizontalPodAutoscaler/frontend-capacity
```

The visible HPA name is intentionally generic (`frontend-capacity`) so the benchmark doesn't leak the fault category through the resource name.

1. Confirms that no namespace `LimitRange` would inject a default CPU request or CPU limit.
2. Waits for metrics-server to report pod metrics for the target deployment.
3. Saves a sanitized snapshot of the original `frontend` deployment to `/tmp`.
4. Removes both `resources.requests.cpu` and `resources.limits.cpu` from the `frontend` container spec.
5. Applies the modified deployment and waits for rollout.
6. Creates a CPU-utilization HPA named `frontend-capacity` with `minReplicas: 1`, `maxReplicas: 5`, target CPU utilization 60%.
7. Waits until the HPA reports `FailedGetResourceMetric` due to `missing request for cpu`.

The injector removes both the CPU request and the CPU limit because removing only `requests.cpu` is not reliable here. While building this I found that with the CPU limit still present, Kubernetes admits the pod with an effective CPU request equal to that limit, and the HPA computes happily (the intended failure never appears). Removing both fields ensures the frontend pods truly lack an effective CPU request.

This temporarily leaves the frontend container without a CPU limit, but the problem doesn't depend on sustained high load, so the blast radius on the kind cluster is low.

Recovery reapplies the original deployment snapshot (restoring the request) and leaves the HPA in place so the mitigation oracle can confirm the autoscaling loop is healthy again. Namespace cleanup removes the injected HPA during teardown.

## 3. Problem runtime behavior and mitigation oracle

After injection, pod health looks normal:

```bash
```

The frontend pod is `Running`/`Ready` and the rest of the app keeps serving. The diagnostic signal is on the HPA:

```bash
kubectl get hpa -n hotel-reservation
```

Expected output:

```text
NAME                REFERENCE              TARGETS             MINPODS   MAXPODS   REPLICAS
frontend-capacity   Deployment/frontend    cpu: <unknown>/60%  1         5         1
```

Then:

```bash
kubectl describe hpa frontend-capacity -n hotel-reservation
```

Expected current condition:

```text
ScalingActive   False   FailedGetResourceMetric
the HPA was unable to compute the replica count: failed to get cpu utilization: missing request for cpu in container hotel-reserv-frontend
```

A correct diagnosis identifies the broken autoscaling loop:

> The frontend HPA is configured to use CPU utilization, but the frontend pods are missing an effective CPU request. Since HPA CPU utilization is computed relative to requested CPU, the HPA can't compute the metric and reports `<unknown>/60%`, `ScalingActive=False`, and `FailedGetResourceMetric` / `missing request for cpu`. The pods stay Running/Ready, but the autoscaler can't make scaling decisions.

The shortcut mitigation is manual scaling, which should not pass:

```bash
kubectl scale deployment/frontend -n hotel-reservation --replicas=3
```

It bumps the replica count but leaves the HPA broken, and the oracle rejects it because the HPA still can't compute its CPU metric.

The canonical mitigation:

```bash
kubectl set resources deployment/frontend -n hotel-reservation \
  --requests=cpu=100m,memory=128Mi
kubectl rollout restart deployment/frontend -n hotel-reservation
kubectl rollout status deployment/frontend -n hotel-reservation --timeout=300s
```

After HPA reconciliation, the fixed state is:

```text
cpu: <real value>%/60%
ScalingActive=True
ValidMetricFound
```

### Oracle behavior

The default pod-health oracle can't grade this, the pods stay Running/Ready while the HPA is broken, so this PR adds `HPAControlPlaneMitigationOracle`.

The oracle is effect-based and doesn't require one specific YAML edit. Restoring CPU requests is the canonical fix, but a valid raw CPU `averageValue` HPA configuration is also accepted as long as the HPA stays correctly wired and healthy.

It passes only when the current cluster state satisfies all of:

1. `Deployment/frontend` exists.
2. Frontend pods are Running/Ready.
3. An HPA exists and targets `Deployment/frontend`.
4. The HPA has a valid CPU metric configuration.
5. The HPA has a valid scaling range.
6. The current CPU metric is readable, not `<unknown>`.
7. `ScalingActive=True`.
8. The current reason is not `FailedGetResourceMetric`.

The oracle does not treat old HPA Events as ground truth. During testing, `kubectl describe hpa` kept showing historical `missing request for cpu` Events even after the fix, so the oracle reads the current object state instead:

```text
.status.conditions
.status.currentMetrics
.spec.scaleTargetRef
```

It also does not require replicas to climb above 1. Actual scale-up depends on load, node timing, HPA sync timing, and the CPU request value the agent picks, so grading on it would make the problem flaky. The deterministic bar is that the HPA can compute its metric and is currently active.

Finally, it uses a short settle loop, up to 60 seconds, requiring 2 consecutive healthy reads, to absorb the post-rollout window where pods are Ready but metrics-server and the HPA controller haven't yet repopulated `status.currentMetrics`.

## 4. Agent behavior and validation

**Manual (cli.py).** The diagnosis was judged correct (100/100), confirming the root-cause wiring. The canonical fix passed after the oracle observed the expected transient empty-metrics state followed by two healthy reads. The manual-scale shortcut was rejected as intended.

**Stratus.** Two runs, `stratus` on `gpt-5.1` (judge `gpt-5.1`, 1800s timeout):

| Run | Diagnosis | Score | Mitigation | TTL | TTM |
|-----|-----------|-------|------------|-------|--------|
| 1 | Failed | 0.11 | Failed | 46.8s | 285.8s |
| 2 | Failed | 0.22 | Failed | 41.4s | 290.6s |

In both runs Stratus inspected pods, deployments, logs, traces, the dependency graph, and Prometheus metrics, saw everything healthy at the app layer, and concluded there was no active fault, it never looked at the HPA or the missing CPU request. Both mitigations ended on `Timed out after 60s waiting for 2 consecutive healthy HPA polls … no current CPU resource metric is reported`. That's the intended behavior: the fault isn't reachable through pod health, app logs, or traces, so an agent has to inspect autoscaling state directly.

---
Looking forward to your feedback! Please let me know if any further details or adjustments are needed.


Saadmrp1038:
@MABDz Thanks for the PR! I will review it soon.

MABDz:
Hi! Had a small update. While testing with agents other than Stratus I noticed a slight issue with false negatives (fault localization) in the Diagnosis Oracle. To fix this I have broadened the expected diagnosis framing so correct agents are not penalized for localizing the fault to the frontend deployment resource configuration, as long as they also connect it to the `frontend-capacity` HPA CPU-utilization failure.

I have reran multiple agents to check that the diagnosis judge is no longer producing false negatives for correct HPA-aware answers, while still rejecting unrelated or too-broad diagnoses.

| Agent | Run | Diagnosis result | Score / composite | Mitigation result | TTL | TTM | Diagnosis summary |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| Stratus | 1 | Failed | 0.00 | Failed | 44.5s | 387.4s | Misdiagnosed the issue as Jaeger / OpenTelemetry service miswiring. It did not mention `Deployment/frontend`, `frontend-capacity`, CPU requests, or HPA metric failure. |
| Stratus | 2 | Failed | 0.22 | Failed | 42.7s | 271.3s | Concluded the application was healthy and treated the only issue as an RBAC / observability limitation. It did not inspect or identify the HPA failure. |
| Codex | 1 | Passed | 1.00 | Passed | 76.4s | 234.8s | Correctly identified that the frontend container has no CPU request, causing `frontend-capacity` to fail CPU-utilization calculation with `FailedGetResourceMetric`. |
| Codex | 2 | Passed | 1.00 | Passed | 89.5s | 174.1s | Correctly localized the fault to the `Deployment/frontend` resource configuration / autoscaling path and noted that backend services were healthy. |
| Claude Code | 1 | Failed | 0.00 | Failed | 1458.8s | 1569.0s | Misdiagnosed the issue as MongoDB startup ordering / missing init containers across multiple services. It did not identify the frontend HPA or CPU request issue. |
| Claude Code | 2 | Passed | 1.00 | Passed | 650.3s | 947.4s | Correctly identified that `Deployment/frontend` is missing CPU requests, causing `frontend-capacity` to fail computing CPU utilization and remain unable to autoscale. |

The updated diagnosis framing now behaves as intended since as can be seen above correct answers that connect the missing frontend CPU request to the HPA control-loop failure pass, while answers that only inspect application health, blame unrelated telemetry/MongoDB components, or never inspect the HPA still fail.  Please let me know if any further details or adjustments are needed!

sregym/conductor/problems/hpa_missing_effective_cpu_request.py
```diff
@@ -0,0 +1,432 @@
...
+
+    HPA_NAME = "frontend-capacity"
+    HPA_CPU_TARGET_PERCENT = 60
+    HPA_MIN_REPLICAS = 1
+    HPA_MAX_REPLICAS = 5
+    HPA_DEFAULT_CPU_REQUEST = "100m"
+    HPA_DEFAULT_MEMORY_REQUEST = "128Mi"
+    HPA_DEFAULT_CPU_LIMIT = "1"
+    HPA_SNAPSHOT_SUFFIX = "hpa_effective_cpu_request_original"
+    HPA_ROLLOUT_ANNOTATION = "sregym.io/frontend-capacity-rollout-at"
```
Saadmrp1038: nit: Let's use a more realistic/normal name here for `HPA_ROLLOUT_ANNOTATION`

Saadmrp1038 (changes requested):
@MABDz Thanks for doing such a thorough implementation and testing! Everything works great. I left a small nit comment. Ping me after you address it and I will merge the PR.

The agent run data is also really insightful. It seems Codex performs really well. What models did you use for the different agents?

sregym/conductor/problems/hpa_missing_effective_cpu_request.py
MABDz: Good catch! I'll update it to use `kubectl.kubernetes.io/restartedAt`. Hope that's better.

Saadmrp1038 (approved):

MABDz:
@Saadmrp1038 Thanks a lot for the review!

I've addressed the nit by replacing the custom rollout annotation with the standard `kubectl.kubernetes.io/restartedAt` annotation. I've also reran the local problem validation after the change just in case and it passes successfully.

As for the models I used the following:
Stratus: `gpt-5.1` 
Codex: `gpt-5.5` 
ClaudeCode: `Sonnet `

The judge model was `gpt-5.1` for all runs. Also just as a note `gpt-5.1` is used here for Stratus here because I didn't realize that Codex (which I ran after the initial few Stratus runs) doesn't support `gpt-5.1` anymore :P I reran stratus on `gpt-5.1` in the latter runs for consistency sake, but it would be interesting to see how it performs with `gpt-5.5` as well.

Saadmrp1038:
@MABDz Congrats on your first PR merged here 🎉

---

Issue #795, opened 2026-05-27
[Problem RFC] PriorityClass Preemption Displaces a Critical Service


munimthahmid:
I am thinking of adding a new SREGym benchmark problem around Kubernetes `PriorityClass` preemption. The idea is to model a case where a high-priority maintenance workload lands on the same node as an important application service. Since the maintenance workload has a higher `PriorityClass` and a large CPU request, Kubernetes may preempt the lower-priority service pod to make room for it.

This is based on a real production failure mode. Grafana documented a Hosted Prometheus outage caused by Kubernetes Pod Priority behavior, where production ingesters were preempted after a priority rollout/configuration issue: https://grafana.com/blog/how-a-production-outage-was-caused-using-kubernetes-pod-priorities/

For SREGym, I am thinking of using the Hotel Reservation app and targeting a user-facing service such as `search`. To keep the scenario deterministic, the fault would place the victim service on a selected worker node, then deploy a high-priority `maintenance-worker` pod onto that same node with a dynamically sized CPU request.

The expected symptom is that the victim service loses ready capacity or its replacement pod stays Pending while the high-priority workload occupies the node. Kubernetes events should show scheduling/preemption or insufficient CPU signals, and frontend requests that depend on the victim service should degrade.

A correct diagnosis should identify the high-priority maintenance workload and Kubernetes priority/preemption behavior as the root cause, rather than blaming the application service itself.

A valid mitigation would restore the service by removing or reducing the harmful maintenance pressure, for example by scaling down the maintenance workload, lowering its priority, reducing its CPU request, or moving it away from the affected node.

Does this idea sound reasonable as a new SREGym benchmark problem? I would appreciate feedback on whether this failure mode fits the benchmark, and whether there is a cleaner way to model the preemption behavior.

Saadmrp1038:
@munimthahmid sounds good. There are some implementation details that you would need to figure out. Like a node can have 192 cpu cores. So, dynamically allocating 190 cores to the priority service might be unreasonable. Maybe we can try to enforce a namespace level cpu limit and adjust based on that.
You should try to implement this. The details would become clearer when actively working on it.

tianyin:
Love the comments from @Saadmrp1038

Saadmrp1038:
Implemented by another contributor (#821). 
Luckily, @munimthahmid hadn't started implementation yet. Would've been wasted effort otherwise.

PR #821, opened 2026-05-30
Add priority preemption cascade problem

thromel:
## Summary

This PR adds `priority_preemption_cascade_hotel_reservation`, a Hotel Reservation benchmark problem based on Grafana Labs' Kubernetes Pod Priorities outage in Hosted Prometheus.

The failure is a scheduler-policy bug. A tenant workload gets a medium PriorityClass, preempts an existing production pod with priority 0, and then blocks the replacement production pod from reclaiming capacity because the replacement inherits the same unsafe default priority. The visible symptom is an under-replicated `reservation` service, but the root cause is PriorityClass policy and scheduler preemption.

## Real-world failure story

Grafana's incident writeup describes a Cortex rollout where new ingesters had medium priority while existing production ingesters had no priority. Under node pressure, Kubernetes preempted the existing ingesters for the new workload. Replacement production pods then had the wrong priority relationship and could not recover the cluster cleanly.

Related links:

- Grafana Labs incident writeup: [How a Production Outage Was Caused Using Kubernetes Pod Priorities](https://grafana.com/blog/how-a-production-outage-was-caused-using-kubernetes-pod-priorities/)
- Grafana config PR referenced by the writeup: [Add critical pod priority to monitoring pods](https://github.com/grafana/jsonnet-libs/pull/147)
- Kubernetes docs: [Pod Priority and Preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/)

## Fault model

The problem starts from a normal Hotel Reservation deployment, then makes the scheduler state explicit:

1. `reservation` gets a capped memory request while its running pod still has priority 0.
2. `reservation` is pinned to the selected pressure node so its replacement cannot escape to another worker.
3. `platform-medium` is created with `globalDefault: true`. This is the bad platform policy.
4. `production-critical` is created with a higher value, but it is not applied to `reservation` during injection.
5. Other Hotel Reservation deployments are assigned `platform-medium`, so `reservation` is the intended preemption victim.
6. `analytics-batch/report-cache-shard-*` padding workloads are pinned to the same node with 8Gi or 16Gi memory requests and the same `platform-medium` priority as the tenant workload. They make large-memory nodes look like a busy cluster without making `reservation` or the tenant request hundreds of GiB.
7. `analytics-batch/tenant-ingester` is pinned to the same node as `reservation` and requests enough memory that it can only schedule by preempting the lower-priority `reservation` pod.

The injector waits for actual evidence: the tenant workload is ready, `reservation` is unavailable, a scheduler preemption event exists, and the replacement `reservation` pod has inherited `platform-medium`.

## What reviewers should see at runtime

When the fault is active:

- `analytics-batch/tenant-ingester` is Running on the target node.
- `analytics-batch/report-cache-shard-*` pods are Running on the same node with `platform-medium` priority and 8Gi or 16Gi requests.
- `reservation` is under-replicated or unavailable.
- scheduler events show preemption and failed scheduling pressure.
- `platform-medium` exists and is still `globalDefault: true`.
- `production-critical` exists but is not applied to the `reservation` deployment template.

Useful checks:

```bash
kubectl get events -A --sort-by=.lastTimestamp | grep -Ei "preempt|failedscheduling|reservation|tenant-ingester"

kubectl get pod -n hotel-reservation \
  -l io.kompose.service=reservation \
  -o custom-columns=NAME:.metadata.name,PHASE:.status.phase,NODE:.spec.nodeName,PRIORITY:.spec.priority,CLASS:.spec.priorityClassName,NOMINATED:.status.nominatedNodeName

kubectl get pod -n analytics-batch \
  -o custom-columns=NAME:.metadata.name,PHASE:.status.phase,NODE:.spec.nodeName,PRIORITY:.spec.priority,CLASS:.spec.priorityClassName,REQ:.spec.containers[*].resources.requests.memory

kubectl get priorityclass
kubectl describe node <target-node>
```

A good diagnosis should avoid app-level explanations like bad image, broken service selector, DNS, or reservation code. The useful evidence is in scheduler events, pod priorities, and node memory requests.

## Expected mitigation

The intended fix is to repair the priority policy while keeping the tenant workload present:

- make `platform-medium` no longer the global default;
- protect `deployment/reservation` with a PriorityClass whose value is higher than `platform-medium`;
- use `production-critical` for that protection, or use another explicit higher-valued class such as `reservation-high-priority`;
- let the `reservation` rollout recover and regain ready endpoints.

The mitigation oracle rejects shortcuts that would make a generic health check pass while missing the fault:

- deleting `analytics-batch/tenant-ingester`;
- scaling the tenant workload to zero;
- deleting the PriorityClasses instead of fixing the policy;
- scaling any Hotel Reservation deployment below one replica;
- reducing the memory requests used to create scheduler pressure.

## Code changes

- `sregym/conductor/problems/priority_preemption_cascade.py`: problem definition, scheduler setup, realistic request sizing, capacity padding, preemption evidence checks, and cleanup for app templates plus PriorityClasses.
- `sregym/conductor/oracles/priority_preemption_mitigation.py`: mitigation oracle for policy recovery and shortcut rejection.
- `sregym/conductor/problems/registry.py`: registers `priority_preemption_cascade_hotel_reservation`.
- `Problem List.md`: adds the new problem entry.
- `tests/problems/test_priority_preemption_cascade.py` and `tests/oracles/test_priority_preemption_mitigation.py`: focused coverage for injection assertions, padding behavior, cleanup, and oracle checks.

No new Python dependencies or RBAC changes are included.

## Validation

Requested CI validation:

```bash
/validate-problem priority_preemption_cascade_hotel_reservation
```

Local checks run after the latest linked-comment fix:

```bash
python3 -m py_compile \
  sregym/conductor/problems/priority_preemption_cascade.py \
  tests/problems/test_priority_preemption_cascade.py

uv run pytest -q \
  tests/problems/test_priority_preemption_cascade.py \
  tests/oracles/test_priority_preemption_mitigation.py

uv run --with ruff ruff check \
  sregym/conductor/problems/priority_preemption_cascade.py \
  tests/problems/test_priority_preemption_cascade.py \
  tests/oracles/test_priority_preemption_mitigation.py

uv run --with black black --check \
  sregym/conductor/problems/priority_preemption_cascade.py \
  tests/problems/test_priority_preemption_cascade.py \
  tests/oracles/test_priority_preemption_mitigation.py
```

The focused unit suite passed with 29 tests.

A targeted local lifecycle on `kind-sregym-calico-spike` also passed before the latest review-fix commits: app deploy, fault injection, oracle failure while the fault was active, recovery, oracle success after mitigation, and cleanup of app namespace, tenant namespace, PriorityClasses, and Hotel Reservation PVs.

No agent-run result is claimed here. The evidence for this PR is the lifecycle validation and the scheduler state described above.


Saadmrp1038:
@thromel Thanks for the PR! I will review it soon.

Saadmrp1038:
@thromel It looks like another contributor opened an RFC for this exact problem 3 days ago(#795). Since they've completed their eval task and haven't started implementation of the fault yet, there's no wasted effort on either side.

That said, please do a quick check for open issues/PRs before creating a new one. With so many people actively implementing faults right now, duplicate efforts can be really unfortunate :)

Thanks!

thromel:
> @thromel It looks like another contributor opened an RFC for this exact problem 3 days ago(#795). Since they've completed their eval task and haven't started implementation of the fault yet, there's no wasted effort on either side.
>
> That said, please do a quick check for open issues/PRs before creating a new one. With so many people actively implementing faults right now, duplicate efforts can be really unfortunate :)
>
> Thanks!

@Saadmrp1038  Thanks for the review, and sorry about that. I should have checked for existing RFCs/issues before starting the implementation.

Fortunately, it sounds like there wasn't any overlapping implementation work yet, but I'll make sure to do a more thorough check next time to avoid duplicate effort.

Thanks for the heads-up!

docs/problem-rfcs/priority_preemption_cascade_hotel_reservation.md
Saadmrp1038, line 1: We probably want some kind of metadata layer over the problems to expose some data like this later but for now it's fine. We should drop this. Only one problem in the docs looks weird :)

sregym/conductor/problems/priority_preemption_cascade.py
```diff
@@ -0,0 +1,706 @@
...
+class PriorityPreemptionCascadeHotelReservation(Problem):
+    """Inject an unsafe global PriorityClass plus a tenant pressure workload."""
+
+    PLATFORM_PRIORITY_CLASS = "platform-medium"
+    PRODUCTION_PRIORITY_CLASS = "production-critical"
+    PRESSURE_NAMESPACE = "analytics-batch"
+    PRESSURE_DEPLOYMENT = "tenant-ingester"
+    PRESSURE_LABEL = "tenant-ingester"
+    PROBLEM_LABEL_KEY = "sregym.io/problem"
+    PROBLEM_LABEL_VALUE = "priority-preemption-cascade"
```
Saadmrp1038: nit: Let's use a more realistic/normal name here and not expose the `sregym` term to the agent.

sregym/conductor/problems/priority_preemption_cascade.py
```diff
@@ -0,0 +1,706 @@
...
+
+        body = client.V1PriorityClass(
+            metadata=client.V1ObjectMeta(
+                name=name,
+                labels={self.PROBLEM_LABEL_KEY: self.PROBLEM_LABEL_VALUE},
+            ),
+            value=value,
+            global_default=global_default,
+            preemption_policy="PreemptLowerPriority",
+            description="SREGym priority preemption cascade simulation.",
```
Saadmrp1038: nit: Let's use a more realistic/normal description here.

sregym/conductor/oracles/priority_preemption_mitigation.py
```diff
@@ -0,0 +1,241 @@
...
+            )
+            return {"success": False}
+
+        template_priority = target_deployment.spec.template.spec.priority_class_name
+        if template_priority != production_priority:
+            print(
+                f"❌ Deployment '{target}' is not explicitly protected with "
+                f"priorityClassName '{production_priority}'"
+            )
+            return {"success": False}
```
Saadmrp1038: This feels too strict. If an agent creates a new PriorityClass (like `reservation-high-priority`) with a value higher than `platform-medium` and assigns it to the reservation deployment, that feels like a valid policy fix. Any unwanted side-effects would result in oracle failure anyway. Is there a reason we would want to mark this as fail?

Saadmrp1038 (changes requested):
@thromel Everything works great! I left some comments below. Mostly minor things.
Ping me after you address them. I will merge the PR.

Saadmrp1038:
@thromel 
After some agent runs I actually found another issue. Real clusters can have a lot of free memory. The nodes I am using each have 192GB of memory. So, the memory requests look something like below.
 - Node allocatable: ~192 Gi
 - reservation request: ~56 Gi
 - tenant-ingester request: ~159 Gi
 
From the agent's perspective this seems excessively large and it tries to modify the memory request instead of investigating deeper and finding the priority issue. Can we do something about it? Maybe deploy some padding pods (with believable framing and each taking 8Gi/16Gi memory) in some namespace to request memory in reasonable amount. Then allocate realistic amount of memory to `reservation` and `tenant-ingestor`. This makes the scenario more believable.

If there's some other way to achieve this that's also okay.

thromel:
Good catch. I pushed a fix in dcfb4b96: `reservation` is capped at 8Gi, and the injector now adds pinned `report-cache-shard-*` padding pods with 8Gi/16Gi requests in `analytics-batch`. The tenant request stays realistic on large nodes, while the priority setup still forces the preemption issue onto `reservation`.

thromel:
@Saadmrp1038 I addressed the review comments, please check again

Saadmrp1038:
@thromel so fast! Thank you. Checking again.

Saadmrp1038 (approved):

Saadmrp1038:
@thromel Thanks for adressing my comments so fast! Looks great now.

Saadmrp1038:
@thromel Are you in our slack channel? I can't seem to find you.
If not you can join here - https://join.slack.com/t/sregym/shared_invite/zt-3z4ce764m-XRProMeLK468GZ2HDNrfUw

PR #825, opened 2026-05-30
Add service protocol mismatch problem

afridi044:
## Summary

This PR adds a new SREGym problem: `service_protocol_mismatch_hotel_reservation`.

The problem simulates a Kubernetes Service protocol mismatch where the `recommendation` Service in the Hotel Reservation application is accidentally changed from TCP to UDP. The pods remain Running and endpoints still exist, but TCP callers cannot reliably reach the service through the Kubernetes Service.

---

## Real-World Failure Story

This models a realistic Kubernetes/Helm manifest misconfiguration where a Service port protocol is accidentally changed during deployment. Because workload pods and endpoints still look healthy, the issue is not obvious from basic pod status checks.

---

## How the Failure Is Simulated on SREGym

The fault injection patches the `recommendation` Service in the `hotel-reservation` namespace, changing the Service port protocol from TCP to UDP:

```yaml
protocol: UDP
```

The recovery path patches the protocol back to TCP:

```yaml
protocol: TCP
```

---

## Problem Runtime Behavior

After injection:

- Hotel Reservation pods remain **Running**
- The `recommendation` endpoint still exists
- The `recommendation` Service has `protocol: UDP` on port `8085`
- TCP callers may **fail or time out** when reaching `recommendation` through the Service

This forces the agent to inspect the Kubernetes Service spec — not only pod status or endpoints.

---

## Agent Behavior

A successful agent should:

1. Determine the failure is **not** caused by pod scheduling, pod readiness, or missing endpoints.
2. Inspect the `recommendation` Service spec.
3. Identify that the Service protocol is incorrectly set to `UDP` instead of `TCP`.
4. Patch the Service protocol back to `TCP` and verify that the mitigation oracle passes.

---

## Mitigation

Patch the Service protocol back to TCP:

```bash
kubectl patch service recommendation -n hotel-reservation \
  --type=json \
  -p='[{"op":"replace","path":"/spec/ports/0/protocol","value":"TCP"}]'
```

---

## Additional Runtime Connectivity Validation

I also validated that the fault has real runtime impact, not only a Service-spec difference.

After fault injection, the `recommendation` Service was confirmed to have protocol `UDP`:

```bash
kubectl get svc recommendation -n hotel-reservation -o yaml | sed -n '/ports:/,/selector:/p'
```

Observed output:

```yaml
ports:
- name: "8085"
  port: 8085
  protocol: UDP
  targetPort: 8085
selector:
```

The application pods were still Running, and the `recommendation` endpoint still existed:

```bash
kubectl get endpoints recommendation -n hotel-reservation -o wide
```

Observed endpoint:

```text
recommendation   10.244.110.168:8085
```

Then I tested TCP connectivity from a temporary in-cluster Python client pod:

```bash
kubectl run tcp-socket-before -n hotel-reservation \
  --rm -i --restart=Never \
  --image=python:3.12-alpine \
  -- python -c "import socket; s=socket.create_connection(('recommendation',8085),5); print('TCP_CONNECTED'); s.close()"
```

Before mitigation, the TCP socket connection timed out:

```text
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.12/socket.py", line 865, in create_connection
    raise exceptions[0]
  File "/usr/local/lib/python3.12/socket.py", line 850, in create_connection
    sock.connect(sa)
TimeoutError: timed out
pod hotel-reservation/tcp-socket-before terminated (Error)
```

This confirms that TCP callers could not reach `recommendation:8085` through the Kubernetes Service while the Service protocol was incorrectly set to UDP.

After that, I patched the Service protocol back to TCP:

```bash
kubectl patch service recommendation -n hotel-reservation \
  --type=json \
  -p='[{"op":"replace","path":"/spec/ports/0/protocol","value":"TCP"}]'
```

Observed output:

```text
service/recommendation patched
```

Then I ran the same Python socket test again:

```bash
kubectl run tcp-socket-after -n hotel-reservation \
  --rm -i --restart=Never \
  --image=python:3.12-alpine \
  -- python -c "import socket; s=socket.create_connection(('recommendation',8085),5); print('TCP_CONNECTED'); s.close()"
```

After mitigation, the TCP socket connection succeeded:

```text
TCP_CONNECTED
pod "tcp-socket-after" deleted from hotel-reservation namespace
```

This before/after comparison confirms the runtime behavior:

- With `protocol: UDP`, pods and endpoints look healthy, but TCP clients time out.
- After restoring `protocol: TCP`, TCP connectivity to `recommendation:8085` is restored.

---

## Validation

/validate-problem service_protocol_mismatch_hotel_reservation

Local validation was run with:

```bash
uv run python tests/integration/validate_problem.py --problem service_protocol_mismatch_hotel_reservation
```

**Result: PASSED**

Validation confirmed:

- Problem resolves from registry
- Hotel Reservation app deploys successfully
- Fault injection completes
- Mitigation oracle detects the injected fault
- `recover_fault()` restores the Service protocol
- Mitigation oracle passes after recovery


Saadmrp1038:
@afridi044 Thanks for the PR.
This feels very similar to `target_port_misconfig_social_network`. Both are single-field Service spec misconfigurations where pods and endpoints look healthy, and the fix is a one-line kubectl patch.

We are looking for more real-world oriented failures. Incidents that caused outages in production and aren't too trivial for the agent to solve. I think you will need to pivot to another fault.

---

Also, the code structure is not correct. The oracle should live in the oracles folder; don't cram both the problem and the oracle into the same file.

Saadmrp1038:
Closing this PR. Please feel free to try again! You can open an RFC to validate your idea first. Also, I recommend going through some of the recently merged `problem` labeled PRs to get an idea.

tianyin:
Thank you for trying @afridi044 !! 

As @Saadmrp1038 mentioned, we are hoping that the failure scenario could reflect real-world failures, not synthetic ones. Otherwise, it is too simple :))

Give another try!

We encourage you to discuss the problem in https://github.com/SREGym/SREGym/issues first before you implement it (to ensure the problem is legit).

Issue #827, opened 2026-05-30
[Problem RFC] Calico route-reflector selector drift partitions cross-node pod networking


thromel:
I would like to propose a new Hotel Reservation problem based on Reddit's Pi-Day outage.

During Reddit's Kubernetes 1.23 to 1.24 upgrade, a Calico route-reflector topology still depended on the legacy `node-role.kubernetes.io/master` node label. As the cluster moved to the newer `node-role.kubernetes.io/control-plane` labeling convention, the Calico `BGPPeer` selector stopped matching the intended control-plane route-reflector nodes. In a Calico BGP topology with `nodeToNodeMeshEnabled: false`, losing the route-reflector peerings can partition cross-node pod and service traffic while many application containers still appear `Running`. The visible symptoms may look like DNS, application, or admission-controller noise, but the root cause is stale CNI/BGP peer selection after a Kubernetes label migration.

References:

- Reddit Engineering: https://www.reddit.com/r/RedditEng/comments/11xx5o0/you_broke_reddit_the_piday_outage/
- Kubernetes 1.24 removals and deprecations: https://kubernetes.io/blog/2022/04/07/upcoming-changes-in-kubernetes-1-24/
- Calico BGP route-reflector configuration: https://docs.tigera.io/calico/latest/networking/configuring/bgp

I checked for overlap before opening this. I did not find an existing issue or PR for the Reddit Pi-Day outage, Calico route-reflector selector drift, `BGPPeer` selector drift, or the `node-role.kubernetes.io/master` to `node-role.kubernetes.io/control-plane` migration.

The closest existing SREGym work looks different:

- #774 covers Calico IPAM/pod CIDR exhaustion. Pods fail because they cannot get IPs.
- #818 covers node-local routing table wipeout. A worker loses per-pod kernel routes.
- #795 and #821 cover the Grafana Kubernetes Pod Priorities outage. That one is scheduler preemption policy, not CNI/BGP routing.

## Required cluster mode

This problem requires a disposable multi-node Kind cluster with the default CNI disabled and Calico installed in a BGP-based mode where cross-node pod routes depend on Calico BGP peerings. A VXLAN-only Calico configuration is not sufficient, because route-reflector peer loss would not necessarily break pod traffic.

Minimum intended topology:

```text
kind-control-plane   route reflector, no app workloads
kind-worker          app/probe node A
kind-worker2         app/probe node B
```

## Proposed simulation

[...]

PR #828, opened 2026-05-31
Add Calico route-reflector label drift problem

thromel:
## Summary

This PR adds `calico_route_reflector_label_drift_hotel_reservation`, a Hotel Reservation benchmark problem based on Reddit's Pi-Day Kubernetes outage.

The failure is a CNI routing-policy bug. Calico is running in route-reflector mode with node-to-node mesh disabled, but its `BGPPeer` still selects route reflectors through the old `node-role.kubernetes.io/master` label. When that label disappears during a Kubernetes label migration, cross-node pod and service traffic breaks even though many pods still look healthy.

This follows RFC #827. The route-reflector setup is done at runtime, so this PR does not change `kind/setup_kind_cluster.sh` or `scripts/ansible/setup_cluster.yml`.

## Real-world failure story

Reddit's incident writeup describes a Kubernetes 1.23 to 1.24 upgrade where Calico route reflectors still depended on the legacy `node-role.kubernetes.io/master` label. After the control-plane label migration, Calico BGP peer selection no longer matched the intended route-reflector nodes. The outage looked like broad service instability, but the root cause was stale CNI/BGP peer selection.

Related links:

- Reddit Engineering writeup: [You Broke Reddit: The Pi-Day Outage](https://www.reddit.com/r/RedditEng/comments/11xx5o0/you_broke_reddit_the_piday_outage/)
- RFC: #827
- Kubernetes docs: [Kubernetes 1.24 removals and deprecations](https://kubernetes.io/blog/2022/04/07/upcoming-changes-in-kubernetes-1-24/)
- Calico docs: [Configure BGP peering](https://docs.tigera.io/calico/latest/networking/configuring/bgp)

## Fault model

The problem starts from a normal Hotel Reservation deployment on a multi-node kind cluster with Calico BGP available. Fault injection then makes the network topology explicit:

1. Calico `BGPConfiguration/default` is set to `nodeToNodeMeshEnabled: false`.
2. The control-plane node is annotated with `projectcalico.org/RouteReflectorClusterID`.
3. `BGPPeer/stale-master-route-reflectors` selects `has(node-role.kubernetes.io/master)`.
4. The old `node-role.kubernetes.io/master` label is added first so the topology starts healthy.
5. Hotel Reservation frontend and reservation pods are pinned to different worker nodes.
6. A `platform-health` probe is created with one client, one same-node server, and one cross-node server.
7. The actual fault removes the old master label, so the BGPPeer selector no longer matches a route-reflector node.

The injector waits for direct network evidence: same-node pod traffic still works, while cross-node Pod IP and Service/DNS probes fail.

## What reviewers should see at runtime

When the fault is active:

- `BGPConfiguration/default` has node-to-node mesh disabled.
- `BGPPeer/stale-master-route-reflectors` still exists but no longer selects the route-reflector node.
- `platform-health/east-west-client` can reach the same-node server.
- the same client cannot reach the cross-node server by Pod IP or Service DNS.
- Hotel Reservation pods may still be Running, so pod phase alone is not enough to diagnose this.

Useful checks:

```bash
kubectl get bgpconfiguration default -o yaml
kubectl get bgppeer stale-master-route-reflectors -o yaml

kubectl get node --show-labels | grep -E "control-plane|master|RouteReflector"

kubectl get pod -n platform-health -o wide
kubectl get svc -n platform-health

kubectl -n platform-health exec deploy/east-west-client -- \
  sh -c "wget -T2 -qO- http://east-west-server.platform-health.svc.cluster.local/health"
```

A good diagnosis should avoid app-level explanations like bad image, broken service selector, DNS-only failure, or Hotel Reservation code. The useful evidence is in Calico BGP configuration, node labels, route-reflector annotations, and cross-node probe behavior.

## Expected mitigation

The intended fix is to restore a valid route-reflector peer selection while keeping route-reflector mode enabled:

- keep `nodeToNodeMeshEnabled: false`;
- keep at least one `BGPPeer`;
- make the BGPPeer select an actual route-reflector node, either through `node-role.kubernetes.io/control-plane`, a deliberately restored legacy label, or a stable custom label such as `route-reflector=true`;
- ensure every selected peer node has `projectcalico.org/RouteReflectorClusterID`;
- let cross-node Pod IP, Service/DNS, and Hotel Reservation probes recover.

The mitigation oracle rejects shortcuts that would hide the fault:

- enabling node-to-node mesh instead of fixing route-reflector peer selection;
- deleting the synthetic probe;
- deleting Calico or BGPPeer resources without restoring cross-node routing;
- scaling Hotel Reservation deployments down;
- forcing app or probe pods onto one node;
- leaving stale unmatched legacy selectors behind;
- using a broad peer selector that also matches nodes without `projectcalico.org/RouteReflectorClusterID`.

## Code changes

- `sregym/conductor/problems/calico_route_reflector_label_drift.py`: problem definition, runtime BGP setup, probe topology, cross-node fault evidence, recovery, and cleanup.
- `sregym/conductor/oracles/calico_route_reflector_mitigation.py`: mitigation oracle for route-reflector recovery and shortcut rejection.
- `sregym/conductor/conductor.py`: global cleanup for interrupted same-cluster runs.
- `mcp_server/k8s/clusterrole.yaml`: node patch/update permissions and Calico BGP resource permissions for MCP use.
- `sregym/conductor/problems/registry.py`: registers `calico_route_reflector_label_drift_hotel_reservation`.
- `Problem List.md`: adds the new problem entry.
- `docs/problem-rfcs/calico_route_reflector_label_drift_hotel_reservation.md`: RFC and operator-facing notes.
- `tests/problems/test_calico_route_reflector_label_drift.py`, `tests/oracles/test_calico_route_reflector_mitigation.py`, `tests/conductor/test_calico_route_reflector_cleanup.py`, and `tests/mcp/test_mcp_rbac.py`: focused coverage for injection, cleanup, oracle checks, and RBAC.

No Python dependencies are added.

## Validation

Requested CI validation:

```bash
/validate-problem calico_route_reflector_label_drift_hotel_reservation
```

Local checks run on the clean PR branch:

```bash
python3 -m py_compile \
  sregym/conductor/problems/calico_route_reflector_label_drift.py \
  sregym/conductor/oracles/calico_route_reflector_mitigation.py \
  sregym/conductor/conductor.py

uv run pytest -q \
  tests/oracles/test_calico_route_reflector_mitigation.py \
  tests/problems/test_calico_route_reflector_label_drift.py \
  tests/conductor/test_calico_route_reflector_cleanup.py \
  tests/mcp/test_mcp_rbac.py

uv run --with ruff ruff check \
  sregym/conductor/conductor.py \
  sregym/conductor/problems/registry.py \
  sregym/conductor/problems/calico_route_reflector_label_drift.py \
  sregym/conductor/oracles/calico_route_reflector_mitigation.py \
  tests/problems/test_calico_route_reflector_label_drift.py \
  tests/oracles/test_calico_route_reflector_mitigation.py \
  tests/conductor/test_calico_route_reflector_cleanup.py \
  tests/mcp/test_mcp_rbac.py

uv run --with black black --check \
  sregym/conductor/conductor.py \
  sregym/conductor/problems/registry.py \
  sregym/conductor/problems/calico_route_reflector_label_drift.py \
  sregym/conductor/oracles/calico_route_reflector_mitigation.py \
  tests/problems/test_calico_route_reflector_label_drift.py \
  tests/oracles/test_calico_route_reflector_mitigation.py \
  tests/conductor/test_calico_route_reflector_cleanup.py \
  tests/mcp/test_mcp_rbac.py
```

The focused unit suite passed with 31 tests.

A targeted local lifecycle on `kind-sregym-calico-spike` also passed: app deploy, fault injection, oracle failure while the fault was active, recovery, oracle success after mitigation, and cleanup. Command used:

```bash
  --problem calico_route_reflector_label_drift_hotel_reservation \
  --summary /tmp/calico-route-reflector-validation-final.md \
  --inject-timeout 300 \
  --recover-timeout 600 \
  --poll-interval 15
```

That run passed in 1210s.

Local environment note: the lifecycle run needed `/run/udev` materialized inside the kind nodes for OpenEBS NDM. That was a local cluster setup issue, not a problem-code change.

No agent-run result is claimed here. The evidence for this PR is the lifecycle validation and the Calico state described above.


Saadmrp1038:
@thromel 
> the lifecycle run needed /run/udev materialized inside the kind nodes for OpenEBS NDM.

Are you using mac?

thromel:
> @thromel 
> 
> > the lifecycle run needed /run/udev materialized inside the kind nodes for OpenEBS NDM.
> 
> 
> 
> Are you using mac?

Yes, I am running this locally on mac.

Saadmrp1038:
@thromel I see. Thanks for the info.
Another person also faced this `/run/udev` issue on mac. Looking into this.

Saadmrp1038:
@thromel just a heads-up since you already have a previous PR and there are already so many pending PRs, it will probably take 1 week+ for me to get to this one. I am treating the first PR as the eval task.

tianyin:
@munimthahmid will help review it.

sregym/conductor/problems/calico_route_reflector_label_drift.py
```diff
@@ -0,0 +1,630 @@
...
+    """Inject stale Calico route-reflector selection after node-label migration."""
+
+    LEGACY_MASTER_LABEL = "node-role.kubernetes.io/master"
+    CURRENT_CONTROL_PLANE_LABEL = "node-role.kubernetes.io/control-plane"
+    ROUTE_REFLECTOR_CLUSTER_ID = "244.0.0.1"
+    ROUTE_REFLECTOR_CLUSTER_ID_ANNOTATION = "projectcalico.org/RouteReflectorClusterID"
+    BGP_PEER_NAME = "stale-master-route-reflectors"
+    PROBLEM_LABEL_KEY = "sregym.io/problem"
+    PROBLEM_LABEL_VALUE = "calico-route-reflector-label-drift"
+    NODE_MARKER_ANNOTATION = "sregym.io/calico-route-reflector-label-drift"
```
munimthahmid: These agent-visible names/labels leak the problem/root cause. Please replace them with generic names/labels.

sregym/conductor/conductor.py
```diff
@@ -551,6 +553,77 @@ async def submit(self, solution: str | None) -> dict:
...
+                return
+
+            kubectl.exec_command(f"kubectl delete namespace {self._q(problem.PROBE_NAMESPACE)} --ignore-not-found")
+            kubectl.exec_command(f"kubectl delete bgppeer {self._q(problem.BGP_PEER_NAME)} --ignore-not-found")
+
+            if bgp_config_labeled:
+                kubectl.exec_command("kubectl delete bgpconfiguration default --ignore-not-found")
+            elif bgp_config:
+                patch = json.dumps({"spec": {"nodeToNodeMeshEnabled": True}})
+                kubectl.exec_command(f"kubectl patch bgpconfiguration default --type=merge -p {self._q(patch)}")
```
munimthahmid: This creates an issue when a run is interrupted and the cluster previously had `nodeToNodeMeshEnabled: false`; directly setting it to `true` changes the cluster config. Since the in-memory snapshot is gone after interruption, please consider a design that stores the original config somewhere before modifying `BGPConfiguration/default`.

Problem List.md
```diff
@@ -47,7 +47,6 @@
...
 | resource_request_too_large                          | Cloud Management System Failure     | New      | Pods in `mongodb-rare` will never be scheduled because the container requires memory that excess limit in every node                                                                                                              | Container          | App           |
 | resource_request_too_small                          | Cloud Management System Failure     | New      | Pods in `mongodb-rare` will never be scheduled because the container requires memory that excess limit in every node                                                                                                              | Container          | App           |
-| hpa_missing_effective_cpu_request_hotel_reservation | Misconfiguration Failure            | New      | The `frontend` HPA cannot compute CPU utilization because target pods lack an effective CPU request, causing `ScalingActive=False` while pods remain Running/Ready                                                                 | Cluster Management | App           |
```
munimthahmid, line 52: This removes the existing `hpa_missing_effective_cpu_request_hotel_reservation` problem from the list. Could you please restore it?

Problem List.md
```diff
@@ -108,4 +108,3 @@
...
 | init_container_dependency_hang_astronomy_shop       | Misconfiguration Failure            | New      | An init container on the `frontend` deployment loops on `nslookup` against a non-existent service (`legacy-config-service`), so every new pod stays in `Init:0/1` and the rollout stalls indefinitely                          | Cluster Management | App           |
-| kubelet_eviction_threshold_misconfig                | Cloud Management System Failure     | New      | `currency` Deployment hard-bound to a worker node via `nodeName`; kubelet `nodefs.available` eviction threshold raised above the node's actual free-space ratio (computed dynamically from kubelet stats summary at injection time), causing the kubelet to evict the pinned pod repeatedly while the node carries the `disk-pressure` taint. Disk itself is healthy — root cause is misconfigured threshold. | Cluster Management | App           |
```
munimthahmid, line 115: This removes the existing `kubelet_eviction_threshold_misconfig` problem from the list. Could you please restore it?

sregym/conductor/problems/calico_route_reflector_label_drift.py
```diff
@@ -0,0 +1,630 @@
...
+                "Calico BGP CRDs are required for route-reflector label drift. " f"{self.CLUSTER_REQUIREMENTS}"
+            )
+        if not self._calico_bgp_dataplane_available():
+            raise RuntimeError(
+                "Calico BGP dataplane is not available; calico-node must expose birdcl for this "
+                f"route-reflector label drift problem. {self.CLUSTER_REQUIREMENTS}"
+            )
+        self._select_nodes()
+        self._capture_bgp_configuration()
+        self._capture_route_reflector_node_state()
```
munimthahmid, line 733: On reinjection, these capture calls can record BGP and node state left by the previous injection as the original cluster baseline. It would be better to check the problem ownership label and `NODE_MARKER_ANNOTATION` before capturing, so problem-owned residue is cleaned instead of later being restored as pre-existing state.

sregym/conductor/problems/calico_route_reflector_label_drift.py
```diff
@@ -0,0 +1,630 @@
...
+            node.metadata.name
+            for node in nodes
+            if (node.metadata.annotations or {}).get(self.NODE_MARKER_ANNOTATION) == "true"
+        ]
+
+    def _delete_support_resources(self):
+        # Calico watches these CRDs and node labels; cleanup avoids another
+        # DaemonSet restart because recover_fault already restarts calico-node.
+        with contextlib.suppress(Exception):
+            self._run(f"kubectl delete namespace {self._q(self.PROBE_NAMESPACE)} --ignore-not-found")
```
munimthahmid: As `PROBE_NAMESPACE` is set to `platform-health`, which is a generic namespace name, an unrelated namespace with that name may already exist. So it willl be better to refuse injection when it already exists without this problem's ownership label, and delete it only when that ownership label is present.

sregym/conductor/problems/calico_route_reflector_label_drift.py
```diff
@@ -0,0 +1,630 @@
...
+        should_delete_bgppeer = self._bgp_config_preexisted is not None or self._problem_created_bgppeer_exists()
+        if should_delete_bgppeer:
+            with contextlib.suppress(Exception):
+                self._run(f"kubectl delete bgppeer {self._q(self.BGP_PEER_NAME)} --ignore-not-found")
```
munimthahmid, line 670: The mitigation oracle accepts replacing the original `BGPPeer` with a differently named valid peer. However, recovery and cleanup only handle the fixed original name and do not delete the replacement.

Therefore, the agent-created resource remains in the cluster after the benchmark ends.

The cleanup should remember which `BGPPeer` resources existed before the run, then remove any new ones created during the run.

mcp_server/k8s/clusterrole.yaml
munimthahmid, line 1: Although agents already have access to the legacy `Endpoints` API, they may try to inspect `EndpointSlice` resources because EndpointSlice is the current Kubernetes API for Service backends. I would prefer adding read-only access to `discovery.k8s.io/endpointslices`.

docs/problem-rfcs/calico_route_reflector_label_drift_hotel_reservation.md
munimthahmid, line 1: As @Saadmrp1038 mentioned. Currently we don't have documentation for other problems. So we can drop this.

munimthahmid (changes requested):
@thromel The failure implementation and mitigation logic look mostly correct. I left a few inline comments on minor issues. Could you please address them?

After addressing them, could you also run the problem end to end with an agent and share how the agent performs, including the diagnosis and mitigation results and any timeouts or difficulties encountered? That would be greatly appreciated.

thromel:
I pushed the review fixes and resolved the threads.

What changed:
- Agent-visible names are generic now, so the problem text does not give away the Calico route-reflector failure.
- The problem snapshots the original `BGPConfiguration/default`, node labels/annotations, and existing BGPPeers before mutation. Cleanup can now restore the cluster state that was already there.
- Reinjection cleanup removes problem-owned residue before baseline capture.
- `platform-checks` is treated as unsafe if it already exists without this problem's ownership label. Cleanup only deletes that namespace when this problem created it.
- Cleanup also removes BGPPeers created during the run, including valid differently named replacement peers accepted by the mitigation oracle.
- Added read-only `discovery.k8s.io/endpointslices` RBAC for the Kubernetes MCP role.
- Fixed the Codex container install issue found during the e2e run. The install script now checks `codex exec` and installs the missing Linux ARM64 optional package when npm skips it.

Checks:
- Local focused tests: 42 passed for the Calico problem, conductor cleanup, mitigation oracle, and MCP RBAC tests.
- `ruff check`, `ruff format --check`, and `git diff --check` passed.
- Rebuilt `sregym-agent-base:latest`; Codex install smoke test passed with `codex-cli 0.141.0`.
- GitHub checks are green on `4ee73a3f`: `lint`, `resolve`, `validate`, `smoke-test`, and `codeql-custom`.
- There are no failing PR checks right now. The only non-success entry is a skipped `Claude Code` run from comment activity.

Codex run:
- Ran `codex` with `gpt-5.5` on `calico_route_reflector_label_drift_hotel_reservation` with `--agent-timeout 3600`.
- Result file: `results/0622_0426/codex_ALL_results.csv`.
- The run finished end to end. Codex diagnosed the `BGPPeer/cluster-peer-policy` selector drift: the peer selector still targeted deprecated `node-role.kubernetes.io/master`, while the actual route-reflector node only had `node-role.kubernetes.io/control-plane`.
- Mitigation succeeded. CSV timing: `TTL=1620.64s`, `TTM=1731.90s`, `Mitigation.success=True`.
- The first Codex attempt exposed the missing Linux ARM64 Codex package in the agent image; that fix is included here. During local setup, Loki `chunks-cache` also needed the same cluster-only memory reduction used in earlier local e2e runs.
- I regraded the saved diagnosis with `openai/gpt-5.5` using the clipboard credentials: `verdict=True`, `composite_score=0.89`, `D1=1.00`, `D2=1.00`, `D3=0.67`.

sregym/conductor/problems/calico_route_reflector_label_drift.py
thromel: Fixed in `8be76b3c`: agent-visible support resource names and ownership labels were changed to generic names/labels (`platform-checks`, `cluster-peer-policy`, `support-check`, etc.).

sregym/conductor/conductor.py
thromel: Fixed in `8be76b3c` and extended in `e6b5283b`: the original `BGPConfiguration/default` state is persisted in the hidden state ConfigMap before mutation, and cleanup restores that persisted state instead of blindly enabling node-to-node mesh.

Problem List.md
thromel: Fixed in `8be76b3c`: restored `hpa_missing_effective_cpu_request_hotel_reservation` in `Problem List.md`.

Problem List.md
thromel: Fixed in `8be76b3c`: restored `kubelet_eviction_threshold_misconfig` in `Problem List.md`.

sregym/conductor/problems/calico_route_reflector_label_drift.py
thromel: Fixed in `8be76b3c` and hardened in `e6b5283b`: reinjection now cleans problem-owned residue before baseline capture, refuses fixed-name BGPPeer collisions, and tracks original BGPPeers before mutation.

sregym/conductor/problems/calico_route_reflector_label_drift.py
thromel: Fixed in `e6b5283b`: injection refuses an existing unowned `platform-checks` namespace, and cleanup deletes the namespace only when it carries this problem's ownership label.

sregym/conductor/problems/calico_route_reflector_label_drift.py
thromel: Fixed in `e6b5283b`: the problem persists the set of BGPPeers present before injection and cleanup removes any new BGPPeer resources created during the run, including differently named replacement peers.

mcp_server/k8s/clusterrole.yaml
thromel: Fixed in `8be76b3c`: added read-only `discovery.k8s.io/endpointslices` access to the Kubernetes MCP ClusterRole.

docs/problem-rfcs/calico_route_reflector_label_drift_hotel_reservation.md
thromel: Fixed in `8be76b3c`: removed the standalone RFC document for this problem.

sregym/conductor/problems/calico_route_reflector_label_drift.py
```diff
@@ -0,0 +1,761 @@
...
+    """Inject stale Calico route-reflector selection after node-label migration."""
+
+    LEGACY_MASTER_LABEL = "node-role.kubernetes.io/master"
+    CURRENT_CONTROL_PLANE_LABEL = "node-role.kubernetes.io/control-plane"
+    ROUTE_REFLECTOR_CLUSTER_ID = "244.0.0.1"
+    ROUTE_REFLECTOR_CLUSTER_ID_ANNOTATION = "projectcalico.org/RouteReflectorClusterID"
+    BGP_PEER_NAME = "cluster-peer-policy"
+    PROBLEM_LABEL_KEY = "sregym.io/managed-by"
+    PROBLEM_LABEL_VALUE = "network-case"
+    NODE_MARKER_ANNOTATION = "sregym.io/network-case-node"
```
munimthahmid: these also leak the name sregym. use a neutral name.

munimthahmid (changes requested):
@thromel Everything looks good except for one minor issue. Could you take a look when you have a chance?

I ran three agent evaluations on this scenario, and the overall takeaway is that it is not easy for the agents, which is good:

1. Claude Sonnet 4.6 spent more than 40 minutes diagnosing the issue, chasing symptoms, and never converged on a root cause before I stopped it(it was burning tokens :) ).
2. One Codex GPT-5.5 run only "succeeded" by reading the leaked harness problem ID from mounted `/logs` artifacts and jumping straight to the fix, so its mitigation success is contaminated (#905 ).
3. Another Codex GPT-5.5 run got the diagnosis mostly right, correctly identifying the stale `master` vs `control-plane` label mismatch in the Calico BGPPeer selector, but its mitigation was rejected because it enabled full node-to-node mesh instead of preserving the intended route-reflector topology.

sregym/conductor/problems/calico_route_reflector_label_drift.py
thromel: Fixed in 4be3396d. The agent-visible ownership metadata now uses neutral `platform.example.com/...` keys, and I added a regression test to prevent reintroducing `sregym` in those keys.

munimthahmid (approved):
Thanks a lot, @thromel, for being patient and fixing all the issues. I ran one more end-to-end run, and it's working properly. Looks good to me. @Saadmrp1038 can have a look and merge it if everything is fine

Saadmrp1038:
Works Great!

PR #829, opened 2026-05-31
Add dev_shm_exhaustion_hotel_reservation problem

### LINKED ISSUE #823: [Problem RFC] Worker Crashes with "No Space Left on Device" Despite a Healthy Node Disk
Hello! I am submitting this RFC as part of the UIUC++ evaluation task for Agentic Site Reliability Engineering (Tianyin Xu et al., UIUC). I went through the existing problems and found a gap I think would make a great addition to SREGym.

## Motivation

I want to propose a failure that I think is one of the most misleading things an SRE can run into in Kubernetes: a pod crash-loops with "No space left on device" while the node disk is nearly empty with plenty of free space. The real cause has nothing to do with the disk. Every container gets a `/dev/shm` shared-memory filesystem that the container runtime silently caps at **64 MiB**, and when an application writes more than that, the kernel returns ENOSPC ("No space left on device"). The disk error is a red herring.

Here is the story of how an agent (or a human SRE) would experience this failure:

The pod goes into `CrashLoopBackOff`. The logs say:

```
dd: error writing '/dev/shm/scratch': No space left on device
```

The natural next step is to check the disk. But this is what comes back:

```
Resource       Used              Free            Shown in kubectl?
-----------    ---------------   -------------   -----------------
Node disk      mostly empty      hundreds of GB  Yes (describe node)
/dev/shm       64 MiB  (100%)    0 bytes         No  <-- silent limit
```

The disk is fine. There is no `DiskPressure` condition. No eviction. Nothing wrong anywhere. And yet the pod keeps crashing. It starts to feel like something is deeply broken that is not visible anywhere.

I find this interesting because the error message is not wrong. There really is no space left. It is just on a completely different resource that Kubernetes never surfaces.

I found similar type of failure well documented in production across many real systems:

- **PostgreSQL inside Kubernetes** crashes parallel workers with: `could not resize shared memory segment "/PostgreSQL.xxxx": No space left on device`, even when the node disk is completely healthy. (https://www.osso.nl/blog/2021/postgresql-inside-kubernetes-no-space-left-on-device/)
- **PyTorch / ML DataLoaders** share tensor batches between worker processes through `/dev/shm`. Workers die with `bus error` or ENOSPC ("No space left on device") when the batch size exceeds 64 MiB. (https://www.sobyte.net/post/2022-04/k8s-pod-shared-memory/)
- **Chromium / Puppeteer** uses `/dev/shm` for renderer IPC and crashes so often in containers that the `--disable-dev-shm-usage` flag exists just to work around this limit. (https://peter.sh/experiments/chromium-command-line-switches/#disable-dev-shm-usage)
- **OpenCV / FFmpeg** media pipelines use `/dev/shm` as scratch during decode and resize operations, and hit ENOSPC ("No space left on device") on large frames. (https://docs.docker.com/engine/containers/run/#runtime-constraints-on-resources)

The 64 MiB default traces back to Kubernetes commit [cbf1cb81](https://github.com/kubernetes/kubernetes/commit/cbf1cb81a90e3e74656a4c6647fae09a8cfee5f5). It is documented, but only if someone already knows to look for it.

## Proposal

I am thinking of adding a `dev_shm_exhaustion_hotel_reservation` problem to the Hotel Reservation benchmark. Hotel images need resizing, and a photo-processing worker (`media-processor`) writing scratch data to `/dev/shm` is a completely believable component in this app. I am using a generic name so the fault is not leaked to the agent.

**Fault injection:** Deploy a worker whose container runs:

```sh
dd if=/dev/zero of=/dev/shm/scratch bs=1M count=128 && tail -f /dev/null
```

The pod template omits a memory-backed `emptyDir` at `/dev/shm`, so the container falls back to the 64 MiB default. Writing 128 MiB fails at exactly 64 MiB with ENOSPC ("No space left on device") and the deployment enters `CrashLoopBackOff`. 

**The intended fix:** Mount an `emptyDir` with `medium: Memory` (e.g. `sizeLimit: 256Mi`) at `/dev/shm` in the worker's pod template.

## Repro

I verified this on my kind cluster (k8s v1.32, 4 nodes) using plain `kubectl` in a throwaway namespace:

| Claim | Observed |
|---|---|
| `/dev/shm` defaults to 64 MiB | `shm 64.0M 0 64.0M 0% /dev/shm` |
| Node disk is healthy | `/` has plenty of free space, nearly empty |
| Writing more than 64 MiB fails | `dd: error writing '/dev/shm/scratch': No space left on device` (died at exactly 64 MiB) |
| Pod enters CrashLoopBackOff | confirmed via `kubectl get pods` |

## Why Agents Struggle

When I thought about what an agent would do here, the sequence is pretty predictable:

1. Agent sees `CrashLoopBackOff` and pulls the logs.
2. Logs say `No space left on device`. Agent concludes: disk is full.
3. Agent checks node conditions and disk. Everything looks healthy. Plenty of free space. No `DiskPressure`.
4. Agent is stuck. The error clearly says no space, but the disk is fine.

The real answer is that there are two completely separate "out of space" resources: the node filesystem (huge, nearly empty) and the `/dev/shm` tmpfs (tiny, always full). Kubernetes only surfaces the first one. The second one is invisible until someone looks inside the container with `df -h /dev/shm` or checks the pod spec for a missing emptyDir volume.

Would love to hear any thoughts before moving forward with the implementation. 😊

## Sources

[1] **OSSO — "PostgreSQL inside Kubernetes / no space left on device"**
https://www.osso.nl/blog/2021/postgresql-inside-kubernetes-no-space-left-on-device/

[2] **SoByte — "Setting up the shared memory of a Kubernetes Pod"**
https://www.sobyte.net/post/2022-04/k8s-pod-shared-memory/

[3] **Chromium `--disable-dev-shm-usage` flag**
https://peter.sh/experiments/chromium-command-line-switches/#disable-dev-shm-usage

[4] **Docker documentation — Runtime options: `--shm-size`**
https://docs.docker.com/engine/containers/run/#runtime-constraints-on-resources

[5] **TimescaleDB issue #5686 — "could not resize shared memory segment"**
https://github.com/timescale/timescaledb/issues/5686

SuMayaBee:
This PR adds a new SREGym problem that reproduces a production Kubernetes failure class where a container crash-loops reporting "No space left on device" while the node disk is nearly empty. The real fault is shared-memory (`/dev/shm`) tmpfs exhaustion, not disk pressure.

The new benchmark problem is:

```text
dev_shm_exhaustion_hotel_reservation
```

Every container gets a `/dev/shm` shared-memory filesystem. The container runtime silently caps it at **64 MiB** (`size=65536k`), a default that traces back to Kubernetes commit [cbf1cb81](https://github.com/kubernetes/kubernetes/commit/cbf1cb81a90e3e74656a4c6647fae09a8cfee5f5). Applications that use POSIX shared memory write to `/dev/shm`; when they write more than 64 MiB, the kernel returns ENOSPC ("No space left on device"), the container exits non-zero, and the deployment enters `CrashLoopBackOff`.

The operational trap is the error message. "No space left on device" reads exactly like a full disk, but the node disk is nearly empty, there is no `DiskPressure` condition, and no kubelet eviction. The actual culprit is a tiny in-RAM tmpfs that Kubernetes never surfaces in any standard metric or condition.

This failure is well documented in production:

- [PostgreSQL parallel queries](https://www.osso.nl/blog/2021/postgresql-inside-kubernetes-no-space-left-on-device/) crash with `could not resize shared memory segment ... No space left on device` even when the node disk is healthy.
- [PyTorch DataLoaders](https://www.sobyte.net/post/2022-04/k8s-pod-shared-memory/) share tensor batches through `/dev/shm` and die with `bus error` or ENOSPC ("No space left on device") when batch sizes exceed 64 MiB.
- [Chromium renderer IPC](https://peter.sh/experiments/chromium-command-line-switches/#disable-dev-shm-usage) crashes so often in containers that `--disable-dev-shm-usage` exists solely to work around this limit.
- [OpenCV/FFmpeg media pipelines](https://docs.docker.com/engine/containers/run/#runtime-constraints-on-resources) use `/dev/shm` as scratch during decode/resize and hit ENOSPC ("No space left on device") on large frames.

In the Hotel Reservation app the believable failure site is a `media-processor` worker resizing hotel images, since image libraries routinely use `/dev/shm` for intermediate buffers.

The new problem lives in:

```text
sregym/conductor/problems/dev_shm_exhaustion_hotel_reservation.py
```

Fault injection deploys a single `media-processor` worker Deployment (`busybox:1.36`) inside the `hotel-reservation` namespace, whose container runs:

```sh
dd if=/dev/zero of=/dev/shm/scratch bs=1M count=128 && tail -f /dev/null
```

The pod template intentionally omits a memory-backed `emptyDir` at `/dev/shm`, so the container falls back to the runtime's 64 MiB default. Writing 128 MiB fails at exactly 64 MiB with ENOSPC ("No space left on device"), the container exits non-zero, and the Deployment enters `CrashLoopBackOff`.

The injection is pod-spec-level only (no host `sysctl`, no privileged node access, no cluster-wide side effects); the entire fault lives inside one namespace-scoped Deployment. The generic name `media-processor` keeps the fault category from leaking to the agent, and `recover_fault()` deletes the worker and waits for the 404, making teardown idempotent.

I verified the mechanism on my kind cluster (k8s v1.32, 4 nodes) before writing the code:

| Claim | Observed |
|---|---|
| `/dev/shm` defaults to 64 MiB | `shm 64.0M 0 64.0M 0% /dev/shm` |
| Node disk is healthy | `/` has plenty of free space, nearly empty |
| Writing more than 64 MiB fails | `dd: error writing '/dev/shm/scratch': No space left on device` (died at exactly 64 MiB) |
| Pod enters CrashLoopBackOff | confirmed via `kubectl get pods` |
| Fix works | `Running 1/1`, `/dev/shm` = 256 MiB after adding the `emptyDir` volume |

## 3. Runtime Behavior and Investigation

After injection, the Hotel Reservation application pods stay healthy and the node has no `DiskPressure`. The only anomaly is `media-processor` in `CrashLoopBackOff`.

Full investigation walkthrough:

```bash
# 1. Spot the crash-looping worker (restart count climbing)

# 2. Read the logs -> the misleading "No space left on device"
kubectl logs -n hotel-reservation -l app=media-processor --previous

# 3. Confirm exit code 1 / Terminated, not OOMKilled or evicted
kubectl describe pod -n hotel-reservation -l app=media-processor

# 4. Rule out the disk: node has no DiskPressure and ample free space
kubectl describe node <node> | grep -A6 Conditions
kubectl get nodes -o wide

# 5. The smoking gun -> /dev/shm is 64 MiB and 100% full, while the disk is fine
kubectl exec -n hotel-reservation deploy/media-processor -- df -h /dev/shm /
```

```text
media-processor pod: CrashLoopBackOff, restart count climbing
container logs:       dd: error writing '/dev/shm/scratch': No space left on device
describe pod:         Last State Terminated, exit code 1 (not OOMKilled, not Evicted)
describe node:        DiskPressure False, disk allocatable nearly empty
df -h /dev/shm:       64.0M used 64.0M (100%)  <-- root cause
df -h /:              mostly free                 <-- disk is not the problem
```

The key distinction from existing disk/OOM problems: the pod is never evicted, the node disk is never involved, and there is no memory pressure. The error comes from an in-RAM tmpfs that Kubernetes does not surface in any metric or condition.

## 4. Expected Diagnosis

```text
Root cause:
The media-processor Deployment writes more than 64 MiB to /dev/shm, but its pod template does not mount a memory-backed emptyDir at /dev/shm. The container falls back to the runtime's default 64 MiB tmpfs. Writes beyond 64 MiB fail with ENOSPC, so the container exits non-zero and the Deployment enters CrashLoopBackOff, even though the node filesystem has ample free disk space.
```

Likely wrong turns (red herrings the symptom invites): node disk pressure / full filesystem, OOM kill, ephemeral-storage eviction, or a bad image/command.

## 5. Expected Mitigation

The fix is to mount a memory-backed `emptyDir` at `/dev/shm` in the worker's pod template. Apply it directly:

```bash
kubectl patch deployment media-processor -n hotel-reservation --type merge -p '{
  "spec": {
    "template": {
      "spec": {
        "volumes": [{"name": "dshm", "emptyDir": {"medium": "Memory", "sizeLimit": "256Mi"}}],
        "containers": [{"name": "worker", "volumeMounts": [{"name": "dshm", "mountPath": "/dev/shm"}]}]
      }
    }
  }
}'
```

Then verify recovery:

```bash
# Worker should now be Running 1/1
kubectl get pods -n hotel-reservation -l app=media-processor

# /dev/shm should now report 256M instead of 64M
kubectl exec -n hotel-reservation deploy/media-processor -- df -h /dev/shm
```

The mitigation oracle is implemented in:

```text
sregym/conductor/oracles/dev_shm_mitigation_oracle.py
```

The default `MitigationOracle` only checks pod health, which an agent could satisfy by simply deleting the crash-looping worker. The custom oracle passes only when all three hold:

```text
1. Worker Deployment still exists with >= 1 desired replica
   (rejects "delete or scale the worker to zero")

2. Pod template mounts a medium: Memory emptyDir at /dev/shm
   (rejects "shrink the write so it fits in 64 MiB" and disk-backed volumes)

3. All worker pods are Running and Ready
   (rejects "just restart it" and an under-sized shm that still overflows)
```

## 6. Validation

Local validation on a 4-node kind cluster using `cli.py` with `gemini-2.5-pro` as the judge (via LiteLLM):

| Run | Diagnosis | Mitigation | Notes |
|---|---|---|---|
| ClaudeCode | 100/100 | Pass | Correctly identified the /dev/shm tmpfs limit; D1/D2/D3 all 1.0. TTL ~821s, TTM ~1046s. |

Oracle rejection tests (manual):

```text
submit() without any fix:        success: false  (default 64 MiB shm still in effect)
delete the worker + submit():    success: false  (Deployment no longer exists)
disk-backed volume + submit():   success: false  (medium: Memory not present)
```

No new Python dependencies are added. No agent RBAC expansion is required.

Closes #823


Saadmrp1038:
@SuMayaBee Thanks for the PR! Might take 1-2 days for me to review.

SuMayaBee:
> @SuMayaBee Thanks for the PR! Might take 1-2 days for me to review.

Okay no worries, So quick reply @Saadmrp1038 , under 1 min. 😮 
I wonder, if you write any scripts to get the notifications instantly, my curious mind wanna know just! 🤔 👀

Saadmrp1038:
@SuMayaBee With so many new issues and PRs, my GitHub tab remains always open :)

Saadmrp1038 (approved):

Saadmrp1038:
@SuMayaBee Thanks for the effort! This fault works great and the code is also very clean. Didn't find any issues.

---

Although one small note. Some of the commands in the PR does not actually work.

The suggested kubectl patch command uses `--type merge` which is JSON Merge Patch. This replaces the containers array entirely, losing the `image` field:
```
$ kubectl patch deployment media-processor -n hotel-reservation --type merge -p '{
  "spec": {
    "template": {
      "spec": {
        "volumes": [{"name": "dshm", "emptyDir": {"medium": "Memory", "sizeLimit": "256Mi"}}],
        "containers": [{"name": "worker", "volumeMounts": [{"name": "dshm", "mountPath": "/dev/shm"}]}]
      }
    }
  }
}'
The Deployment "media-processor" is invalid: spec.template.spec.containers[0].image: Required value
```


For this one the container crashes too quickly, so there's never a window to exec in:
```
$  kubectl exec -n hotel-reservation deploy/media-processor -- df -h /dev/shm /
error: Internal error occurred: unable to upgrade connection: container not found ("worker")
```

---
No functional bugs though. Merging the PR!

Saadmrp1038:
@SuMayaBee Congrats on your first PR merged here 🎉
Keep up the good work!

SuMayaBee:
@Saadmrp1038 Thanks for the merging the PR, meanwhile I will edit the PR description. Happy to contribute something in SREGym! 😊

PR #832, opened 2026-05-31
Add file_descriptor_exhaustion problem

TejasShukla2007:
Adds a new problem simulating file descriptor exhaustion.

In Unix-like systems, a file descriptor (FD) is a handle to access files, sockets, pipes, or any I/O resource. Each process is allowed only a limited number of these. The default soft limit is 1024 in containers. When the limit is hit, the app does not crash but rather starts dropping connections or logging mysterious I/O errors. 

More details on this topic have been mentioned in the following blog - https://medium.com/@badrikmodi/the-silent-killer-file-descriptor-exhaustion-in-modern-services-9144bf795b6f

### 1. Real World Failure Story

1. [Moving the Entire Stack to K8s Within a Year – Lessons Learned - ThredUP - DevOpsStage 2018](https://youtu.be/tA8Sr3Nsx1I?t=1414)

In this case, every time new containers are run, new file descriptors are allocated without deleting the already existing file descriptors. Consequently, this causes a "Too many open files" error.  

2. [Millions of open file descriptors for /usr/bin causing downtime](https://gitlab.com/ongresinc/stackgres/-/work_items/3023)

It describes how the number of file descriptors grew to millions within a few days, causing containers within the pods to crash/restart continually.

### 2. How the failure is simulated on SREGym

In real world systems, this failure builds up gradually and it takes days for its symptoms to show. So, the new problem class at `sregym/conductor/problems/file_descriptor_exhaustion.py` simulates the problem by spinning up a background thread that opens `1100` connections to the `frontend` on port `5000`, holds them, sleeps for some time, releases current connections, and creates new ones. This exhausts the file descriptor limit fast. To open connections to port `5000` of the `frontend`, `kubectl port-forward` is used. 

The frontend's file descriptor limit is kept at the standard Linux default, i.e., a value of `1024`.

The `frontend` successfully accepts the connections till it hits the limit, and then begins rejecting the traffic with `too many open files` error.

A custom oracle `FDMitigationOracle` is used that inherits from the generic `MitigationOracle` to run the generic tests. Additionally, it checks the logs of the faulty service pod for the `too many open files` error. If it does not see this error for a sustained period, the mitigation passed verdict is given. The oracle checks the log for a sustained period so that the agent cannot cheat the test by simply restarting the pod and immediately submitting.

The problem is registered as `"file_descriptor_exhaustion"` under `registry.py`.

### 3. Problem Runtime Behaviour

Once the fault is injected, the cluster has the following observable state:

Pod health: 
```bash
```
On running the above command, the frontend pod shows 1/1 Running with 0 restarts. A naive agent might get misled by this observation assuming everything to be fine. 

Application logs:
```bash
kubectl logs frontend-7574f495f8-fm46j -n hotel-reservation
```
Only after checking the application logs would an agent be able to diagnose the problem. The above command results in an output of the following kind:

```bash
2026-05-31T07:10:13Z INF services/frontend/server.go:140 > srv-reservation.
2026-05-31T07:10:13Z INF services/frontend/server.go:65 > Successfull
2026-05-31T07:10:13Z INF services/frontend/server.go:87 > Serving http
2026/05/31 07:10:15 http: Accept error: accept tcp [::]:5000: accept4: too many open files; retrying in 5ms
2026/05/31 07:10:15 http: Accept error: accept tcp [::]:5000: accept4: too many open files; retrying in 10ms
```
The `too many open files` error helps in determining that this is a file descriptor exhaustion issue.

One way for the AI agent to successfully diagnose and mitigate the issue is to check the pod logs, identify the file descriptor exhaustion, and elevate the limit to a suitable value, preferably to the hard limit (`ulimit -n $(ulimit -Hn)`).

### 4. Fault recovery
The fault is recovered by stopping the background flooder thread and closing the port-forward tunnel. Moreover, the `ulimit` patch on the `frontend` deployment is reverted to cleanly restore its original entrypoint configuration.

### 5. Problem Validation
The problem passed `uv run python tests/integration/validate_problem.py --problem file_descriptor_exhaustion` on my system.

/validate-problem file_descriptor_exhaustion


Saadmrp1038:
@TejasShukla2007 Thanks for the PR!

The file descriptor value of 30 feels too unrealistic and could have unintended consequences. Is it possible to use a normal limit value like `1024/4096`, then do something that requires a higher limit (check the file descriptor limit and then intentionally put load that exceeds that value)? I actually faced this similar issue while testing another fault #768. Check the review comments. I ran into the error `OSError: [Errno 24] No file descriptors available` because the default file descriptor limit was too low for the amount of connections we were creating.

TejasShukla2007:
> @TejasShukla2007 Thanks for the PR!
> 
> The file descriptor value of 30 feels too unrealistic and could have unintended consequences. Is it possible to use a normal limit value like `1024/4096`, then do something that requires a higher limit (check the file descriptor limit and then intentionally put load that exceeds that value)? I actually faced this similar issue while testing another fault #768. Check the review comments. I ran into the error `OSError: [Errno 24] No file descriptors available` because the default file descriptor limit was too low for the amount of connections we were creating.

I have changed the fault injection method. Now, the problem is simulated by creating a pod that puts a load on the `frontend` beyond its file descriptor limit. The file descriptor limit for the `frontend` pod is set at `1024`. 

This is much closer to the real world case scenario of ThredUp where the error was caused due to an internal component creating too many file descriptors.

sregym/conductor/problems/file_descriptor_exhaustion.py
```diff
@@ -0,0 +1,104 @@
...
+            description=(
+                f"The {self.faulty_service} deployment is encountering file descriptor exhaustion. "
+                f"The current limit (ulimit -n {self.forced_ulimit}) is insufficient for the deployment, "
+                f"causing the 'Too many open files' error."
+            ),
+        )
+        self.diagnosis_oracle = LLMAsAJudgeOracle(problem=self, expected=self.root_cause)
+
+        self.app.create_workload()
+        self.mitigation_oracle = MitigationOracle(problem=self)
```
Saadmrp1038: `MitigationOracle` only checks that all pods are in `Running` phase with ready containers. After fault injection, the frontend pod stays `Running` with 1/1 `Ready`. It doesn't crash on FD exhaustion. It logs "too many open files" and retries. The `fd-flooder` pod is also `Running/Ready`.

As a result, the mitigation oracle will pass even if the agent does nothing.

sregym/conductor/problems/file_descriptor_exhaustion.py
```diff
@@ -0,0 +1,104 @@
...
+
+        container = {
+            "name": "flooder",
+            "image": "python:3.12-alpine",
+            "command": ["sh", "-c", 'ulimit -n $(ulimit -Hn); exec python -c "$SCRIPT"'],
+            "env": [{"name": "SCRIPT", "value": script}]
+        }
+
+        deployment = {
+            "metadata": {"name": "fd-flooder", "labels": {"app": "fd-flooder"}},
```
Saadmrp1038: nit: we should use a realistic/generic name here

sregym/conductor/problems/file_descriptor_exhaustion.py
Saadmrp1038, line 1: nit: We have ruff setup for this repo. You can check https://github.com/SREGym/SREGym/pull/484 for how to set it up locally. Will help catch simple lint errors and import ordering issues.

sregym/conductor/problems/file_descriptor_exhaustion.py
```diff
@@ -0,0 +1,104 @@
...
+from sregym.conductor.problems.base import Problem
+from sregym.service.kubectl import KubeCtl
+from sregym.utils.decorators import mark_fault_injected
+
+class FileDescriptorExhaustion(Problem):
+    def __init__(self):
+        self.app = HotelReservation()
+        self.namespace = self.app.namespace
+        self.faulty_service = "frontend"
+        self.forced_ulimit = 20
```
Saadmrp1038: Didn't we update it to 1024? This makes the root cause description wrong.

Saadmrp1038 (changes requested):
@TejasShukla2007 Thanks for the effort! I left some inline comments below. Please ping me when you've addressed them.

sregym/conductor/problems/file_descriptor_exhaustion.py
```diff
@@ -0,0 +1,104 @@
...
+
+        super().__init__(app=self.app, namespace=self.namespace)
+        self.kubectl = KubeCtl()
+        self.root_cause = self.build_structured_root_cause(
+            component=f"deployment/{self.faulty_service}",
+            namespace=f"{self.namespace}",
+            description=(
+                f"The {self.faulty_service} deployment is encountering file descriptor exhaustion. "
+                f"The current limit (ulimit -n {self.forced_ulimit}) is insufficient for the deployment, "
+                f"causing the 'Too many open files' error."
```
Saadmrp1038: Need to include what is eating up the connections. Otherwise agent could just say that the error is this pod is taking up 1000+ connections.

Saadmrp1038 (changes requested):

Saadmrp1038 (commented):
The agent tried to delete the flooder pod to fix the fault, which is not what we want. The thing with how it is currently set up is that the agent can see the TCP holding script and decide that it is not needed, and deletes it. Another approach to this could be a continuous flooding system. During injection, we would spin up a background thread that will continuously create a connections, sleep for some time, release the current connections, and create a new ones.

TejasShukla2007:
@Saadmrp1038 I have changed the fault injection method. Now, the fault is injected by spinning up a background thread. Moreover, I have updated the root cause description and added a custom oracle `FDMitigationOracle` to check the logs for `too many open files` error. Note that it checks the logs for some time before giving the verdict so that an agent cannot cheat by simply restarting the faulty service pod and immediately submitting.

Saadmrp1038 (approved):

Saadmrp1038:
@TejasShukla2007 The changes look good! I ran and tested the problem multiple times. Everything seems to be working correctly now.

| Run | Diagnosis | Mitigation | Agent Behaviour |
|-----|-----------|------------|------------------|
| 1 | 89% ✅ | ❌ Timeout | Wasted ~15 min on app noises before finding frontend logs. Found correct fix (`ulimit -n 65536`) but ran out of time. |
| 2 | 100% ✅ | ✅ | Faster diagnosis (~2 min). Hit same trap (removing ulimit still fails), but quickly realized default soft=1024 and applied explicit raise. |
| 3 | 100% ✅ | ❌ | Again perfect diagnosis. Removed ulimit wrapper -> saw errors briefly stop (flooder reconnecting) -> submitted too early. Flooder came back, oracle caught it. |

The injection sets `ulimit -n 1024` which is the same as the Docker default soft limit for containers (1024 soft / 524288 hard). So when agents "fix" it by just removing the ulimit wrapper, the container still starts with soft limit of 1024 and the flooder still exhausts it. The only real fix is explicitly raising the limit (`ulimit -n 65536`). This makes the problem harder than it looks. 2/3 agents fell into this trap.

On some machines the Docker default soft limit might be higher than 1024 (if manually changed). In that case, just removing the wrapper would actually fix it. I guess that's fine. Unless manually changed, Docker defaults to 1024 soft limit.

Saadmrp1038:
@TejasShukla2007 Congrats on your first PR merged here 🎉

PR #835, opened 2026-05-31
Add PodDisruptionBudget-blocks-node-drain problem

abdallahade1:
# New Problem: PodDisruptionBudget Blocks Node Drain (`pdb_blocks_drain_hotel_reservation`)
_Related RFC: #833_

## Summary

This PR adds a new problem simulating a misconfigured **PodDisruptionBudget (PDB)** that deadlocks voluntary node maintenance. A PDB with `minAvailable` equal to the target deployment's replica count permits **zero** voluntary disruptions, so draining the node that hosts the protected pod hangs forever, the node is stuck `SchedulingDisabled` and the drain never completes, even though the application itself keeps serving traffic.

To my knowledge no existing problem covers the **voluntary-disruption / PDB** failure class. It is distinct from the resource-based problems (`resource_request_too_large` → scheduling `Pending`; `resource_request_too_small` → self `OOMKilled`) and from the node assignment problems: the failure mechanism here is the **disruption controller's budget arithmetic**, not resource pressure or scheduling capacity. This makes it deterministic and reproducible on an emulated Kind cluster, where node-level resource pressure is not reliably reproducible.

---

A team adds a PodDisruptionBudget to protect a service from losing availability during cluster operations. The budget is set too aggressively, `minAvailable` equal to (or above) the deployment's replica count so the budget mathematically permits no voluntary pod evictions at all.

Everything looks healthy until a routine maintenance event: an operator cordons a node and runs `kubectl drain` to move workloads off it (for a kernel patch, an instance replacement, a cluster-autoscaler scale-down, etc.). The drain evicts everything it can, then blocks indefinitely on the protected pod, retrying every few seconds with `Cannot evict pod as it would violate the pod's disruption budget`. The node sits in `Ready,SchedulingDisabled` and the maintenance operation never finishes. The application is still up, so dashboards are green but node maintenance, rollouts that rely on eviction, and autoscaler scale downs all silently deadlock. This is the classic "a safety mechanism becomes a self-inflicted denial of service" incident.

This pattern is well documented as a common maintenance-time failure:

- **VMware / Broadcom KB**: a PDB with zero allowed disruptions blocks node draining, leaving the node `Ready,SchedulingDisabled` and getting cluster upgrades stuck: https://knowledge.broadcom.com/external/article/345904/
- **Red Hat OpenShift**: Having both a PodDisruptionBudget and replicas set to 1 leads to the drain being blocked, so an administrator cannot drain the node for OS updates: https://access.redhat.com/solutions/4736031
- **Statsig (real incident)**: Switched from `minAvailable` to `maxUnavailable` after discovering services with fewer than five pods were stuck at `disruptionsAllowed: 0` permanently: https://jorijn.com/en/knowledge-base/kubernetes/deployments/kubernetes-pod-disruption-budget/
- **Spectro Cloud**: blocking PDBs cause cloud-managed clusters to spin up more and more nodes during rolling upgrades because nodes never finish draining: https://www.spectrocloud.com/blog/how-to-handle-blocking-poddisruptionbudgets-on-kubernetes-with-distributed-storage
- **Practitioner writeup (OpenShift node-drain failure due to active PDB)**: https://medium.com/@tamber/solution-avoid-kubernetes-openshift-node-drain-failure-due-to-active-poddisruptionbudget-df68efed2c4f

---

## 2. How the Failure Is Simulated on SREGym

Target: the `frontend` deployment of the Hotel Reservation application (the problem is written generically and also accepts `social_network` / `astronomy_shop`).

**Injection** (`VirtualizationFaultInjector.inject_pdb_blocks_node_drain`):

1. Read the target deployment's current replica count and its label selector.
2. Create a PodDisruptionBudget named **`maintenance-guard`** that selects the deployment's pods with `minAvailable` equal to the replica count. This yields `ALLOWED DISRUPTIONS = 0`.
3. Find the node hosting the pod and `cordon` it (simulating the start of a maintenance operation). The node name is persisted to `/tmp/<svc>_pdb_drain_node.txt` for recovery.
4. Start a background `kubectl drain <node> --ignore-daemonsets --delete-emptydir-data --force --timeout=0`. The drain evicts everything it can, then deadlocks retrying on the protected pod.

**Recovery** (`recover_pdb_blocks_node_drain`): delete the PDB, uncordon the saved node, and wait for the namespace to stabilize.
The fault component is deliberately given a **generic, plausible name** (`maintenance-guard`) so it does not leak the benchmark identity or the fault category to the agent.

**Why this design fits an emulated cluster:** the PDB / disruption-controller mechanism is evaluated synchronously by the control plane and does not depend on node memory/CPU pressure, the kubelet eviction manager, or stats pipelines, all of which are unreliable or disabled on Kind. I verified empirically that Kind's kubelet has memory eviction disabled (`evictionHard` has no `memory.available` threshold), which is precisely why a resource-pressure based problem is *not* reproducible here but this PDB-based problem *is*.

### Files

- `sregym/generators/fault/inject_virtual.py`: added `inject_pdb_blocks_node_drain` / `recover_pdb_blocks_node_drain`.
- `sregym/conductor/problems/pdb_blocks_drain.py`: `PDBBlocksDrain` problem class.
- `sregym/conductor/oracles/pdb_blocks_drain_mitigation.py`: `PDBBlocksDrainMitigationOracle`.
- `sregym/conductor/problems/registry.py`: import + registry entry `pdb_blocks_drain_hotel_reservation`.

### Mitigation Oracle

The oracle checks three conditions, so it cannot be satisfied by a workaround that fails to address the root cause:

1. **No PDB blocks disruption**: every PDB in the namespace reports `status.disruptionsAllowed > 0` (the over-constrained PDB was corrected or removed).
2. **No node left cordoned**: no node is `unschedulable`.
3. **All pods Running and ready.**

Reading `disruptionsAllowed` directly is deterministic (the disruption controller computes it synchronously) and has no side effects, so the oracle never disturbs the cluster. An agent that merely uncordons the node without fixing the PDB still fails check 1; an agent that deletes the PDB but leaves the node cordoned fails check 2.

---

## 3. Problem Runtime Behavior

After `start pdb_blocks_drain_hotel_reservation`, injection logs:

```
Fault Injection
Created over-constrained PDB 'maintenance-guard' (minAvailable=1) for frontend 
Cordoned node kind-worker3 (now SchedulingDisabled)
Started stuck drain on kind-worker3; it will deadlock on frontend (blocked by PDB)
```

Observable cluster symptoms:

```
$ kubectl get pdb -n hotel-reservation
NAME                MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
maintenance-guard   1               N/A               0                     89s

$ kubectl get nodes
NAME                 STATUS                     ROLES           AGE   VERSION
kind-worker3         Ready,SchedulingDisabled   <none>          11h   v1.32.0
```

The background drain loops indefinitely:

```
error when evicting pods/"frontend-..." -n "hotel-reservation" (will retry after 5s):
Cannot evict pod as it would violate the pod's disruption budget.
```

The `frontend` pod itself stays `Running` throughout, the application keeps serving, which is what makes this incident subtle: only voluntary disruption operations are deadlocked.

The node selected for cordon/drain is discovered dynamically from where the pod is scheduled (observed on `kind-worker2` and `kind-worker3` across separate runs), so the problem is not tied to any specific node.

---

## 4. Mitigation Verification (Grader Results)

**Unmitigated → fail** (`submit()` with the fault still active):

```
Mitigation Evaluation
❌ PDB 'maintenance-guard' still blocks disruption (disruptionsAllowed=0)
❌ Node kind-worker2 is still cordoned (SchedulingDisabled)
Mitigation Result: ❌ Fail
"Mitigation": { "success": false }
```

**Mitigated → pass** (after `kubectl delete pdb maintenance-guard` + `kubectl uncordon <node>`, then `submit()`):

```
Mitigation Evaluation
Mitigation Result: ✅ Pass
[EVAL] Mitigation Succeed
"Mitigation": { "success": true }
```

Fault recovery and baseline reconciliation also run cleanly during teardown:

```
Fault Recovery
Deleted PDB 'maintenance-guard'
Uncordoned node kind-worker3
Recovered from PDB-blocked-drain fault for frontend
```

---

## 5. Agent Behavior

I was unable to complete a Stratus agent run for this submission because the LLM provider available to me (Gemini free tier via LiteLLM) was rate-limited during testing, the diagnosis LLM-as-a-judge call exhausted its retries with `Max retries exceeded. Unable to complete the request.`, and a full multi-step agent run was not feasible under the same limits. The **mitigation oracle is pure Kubernetes logic and does not depend on an LLM**, so it was validated directly through the grader (Section 4).

I intend to follow up with a Stratus run once a non-rate-limited model/key is available. I expect the relevant edge case for an agent to be the "partial fix" trap: uncordoning the node (which makes the cluster *look* healthier) without correcting the PDB. The oracle's `disruptionsAllowed > 0` check is specifically designed to reject that workaround.

---

## Notes / Open Questions

- The problem is registered for Hotel Reservation; the class accepts `social_network` and `astronomy_shop` as well if you'd like additional registry entries.
- Injection cordons a node and starts a real background kubectl drain, so it modifies node scheduling state (cleaned up on recovery). Per discussion on Slack, this is acceptable to run on the emulated cluster without gating behind non_emulated_cluster_problems.
- Naming: the PDB is `maintenance-guard` to avoid leaking the fault category. Open to a different generic name if you have a convention.


Saadmrp1038 (changes requested):
@abdallahade1 Thanks for the effort! Fault works great but some issues are present here:

1. The `kubectl drain --force` evicts ALL non-PDB pods from the node. But MongoDB pods with PV node affinity can't reschedule (FailedScheduling: "volume node affinity conflict"), and services depending on them go into CrashLoopBackOff.
   
   This results in app being completely broken and not "silently deadlocked" as described. This makes the diagnosis trivial (crashing pods + FailedScheduling events lead directly to the cordoned node) and defeats the purpose of the fault.
   
   This is non-deterministic since it  can only happen when the drained node hosts MongoDB. When it doesn't, the fault works as described.
   
   *after fault injection when target node has mongodb*
```
$ kubectl get pods -n hotel-reservation -o wide
NAME                                      READY   STATUS    RESTARTS      AGE   IP              NODE
consul-5c696cc4dc-xk4xt                   1/1     Running   0             33m   10.244.153.62   node0
frontend-6bbffdc996-cbn99                 1/1     Running   0             67s   10.244.27.32    node1
geo-57c7bd6d5c-r6m5b                      1/1     Running   4 (30m ago)   31m   10.244.153.0    node0
jaeger-5b9775bd94-qp6fw                   1/1     Running   0             31m   10.244.153.10   node0
memcached-profile-6849578b8-t9tb9         1/1     Running   0             13m   10.244.52.121   node2
memcached-rate-948594dfd-gq9pz            1/1     Running   0             13m   10.244.52.75    node2
memcached-reserve-68dbb57476-f2sdd        1/1     Running   0             13m   10.244.52.112   node2
mongodb-geo-8cb7d9f5b-9k8ff               0/1     Pending   0             29s   <none>          <none>
mongodb-profile-775dfd4bd7-5pb92          0/1     Pending   0             28s   <none>          <none>
mongodb-rate-bfbcf4587-nx6q4              0/1     Pending   0             29s   <none>          <none>
mongodb-recommendation-674dddbcdf-rhrnb   0/1     Pending   0             29s   <none>          <none>
mongodb-reservation-6cdf69dcc-h9p95       0/1     Pending   0             29s   <none>          <none>
mongodb-user-6dcf4877-9l6ch               0/1     Pending   0             29s   <none>          <none>
profile-69cb98cb67-qr799                  1/1     Running   3 (12m ago)   13m   10.244.52.79    node2
rate-6fff88f6f5-zxb6j                     1/1     Running   3 (12m ago)   13m   10.244.52.105   node2
recommendation-64754c9c4f-xx64n           1/1     Running   3 (12m ago)   13m   10.244.52.109   node2
reservation-7bb77b596b-n9qfn              1/1     Running   4 (30m ago)   31m   10.244.153.27   node0
search-5f668579c9-fgg8p                   1/1     Running   0             13m   10.244.52.106   node2
user-6548d868cd-vmlkc                     1/1     Running   3 (12m ago)   13m   10.244.52.83    node2
```

*warning event on mongo pods*
```
Warning  FailedScheduling  33s   default-scheduler  0/3 nodes are available: 1 node(s) were unschedulable, 2 node(s) had volume node affinity conflict.
```

2. If `recover_fault` fails or is skipped for any reason, the node stays cordoned across subsequent problem runs. We need to add a proper cleanup step to `fix_kubernetes()` in `conductor.py`.

abdallahade1:
@Saadmrp1038 
Thanks for the detailed review. I've pushed fixes:
1. Drain breaking the app / non-determinism. You were right that kubectl drain --force evicting every pod (including PV-pinned MongoDB that can't reschedule) broke the app and made the symptom non-deterministic. I removed the full drain entirely. Injection now cordons the node and makes a single eviction attempt against only the target (frontend) pod, which the PDB rejects. No other pods are ever touched, so MongoDB and the dependent services stay healthy regardless of which node hosts them, the "looks healthy but maintenance is blocked" symptom is now deterministic by construction. The observable signals (node SchedulingDisabled, PDB disruptionsAllowed: 0, the rejected eviction) are unchanged.
2. Leftover cleanup. Added a cleanup step to fix_kubernetes() that removes any leftover maintenance-guard PDB and uncordons any node left SchedulingDisabled, so a skipped/failed recovery can't pollute later runs. I confirmed this hook runs during the startup fix phase ([FIX] PDBBlocksDrain leftover if any).
One note: I wasn't able to capture a fresh full end-to-end CLI run as final confirmation, my local Kind cluster kept losing worker nodes to WSL sleep/instability between runs. The drain removal makes the MongoDB-eviction issue structurally impossible (visible in the diff), and I verified the cleanup hook executes in the startup logs. If you'd like any other changes please tell me.

Saadmrp1038 (changes requested):
@abdallahade1 Can we keep the kubectl drain? It feels more natural. Without it there's too little for the agent to go on. A maintenance operation normally includes both cordon and drain.

Maybe you could set up the pods so that the frontend and MongoDB pods are not on the same node. That way, when we cordon and drain the node frontend is on, it won't break anything (the stateless pods just reschedule, and the drain hangs on the PDB-protected frontend).

abdallahade1:
@Saadmrp1038 
I agreed this is better than removing the drain, so I've implemented your suggestion.
Injection: before draining, I now detect which nodes host MongoDB pods, add soft pod anti-affinity so frontend reschedules onto a MongoDB-free node, then explicitly verify frontend's node has no MongoDB pod before cordoning and running the real kubectl drain. So the full cordon+drain is back, but it can never hit a PV-pinned MongoDB pod, the separation is guaranteed by the explicit check, not just the soft preference, so the fault stays deterministic and the app stays healthy. The drain still deadlocks on the PDB-protected frontend, which is the intended symptom.
Recovery: deletes the PDB, uncordons the node, and removes the injected anti-affinity so scheduling returns to normal.
I'm still completing a full clean grader run, my local WSL/Kind setup struggles to keep the whole stack stable (the worker kubelets give out under the full observability + app load), so I'm sorting out a larger environment to confirm the live run.

sregym/conductor/problems/registry.py

Saadmrp1038 (changes requested):
@abdallahade1 Works great now! Just found a small lint error.

On a larger note, after doing some agent runs there's really no visible fault symptom for the agent to diagnose. The agent doesn't even know that a drain command was run and failed to complete. So even when the agent initially noticed the PDB, it didn't think there was anything wrong with it.

The agent's exact reasoning was:
```
The `maintenance-guard` PDB has `MIN AVAILABLE: 1` for the frontend service with `ALLOWED DISRUPTIONS: 0`. This means the frontend is locked down—no disruptions allowed. The PDB is currently satisfied since the frontend is running at 1/1, so it's not causing an active issue right now.
```

Funnily, after some more running around chasing other leads, it did find the exact issue:
```
Now I see a critical issue:
  1. node1 is cordoned (Unschedulable: true)
  2. The frontend pod is running on node1 (the only cordoned node)
  3. The PDB maintenance-guard prevents the frontend from being disrupted (allowedDisruptions: 0)

This is the fault! ... The PDB creates a deadlock: if an admin tries to drain the cordoned node to take it offline safely, the PDB blocks the eviction because it requires at least one frontend pod to remain available. With only one frontend pod running, there's no room for disruption, so the node can't be drained and the pod can't be moved.
```

But it doubted itself and went in another direction:
```
But stepping back, I'm wondering if I'm looking at the right fault. Let me reconsider what was actually injected—maybe it's not the cordoning or the PDB, but something about the geo service itself.
```

Ultimately the agent timed out before it could submit anything. Based on the provided incident reports, what do you think? Is it fine as is, or should there be some sort of signal for the agent to see?

---

~I think it could be fine as it is now. The agent did notice the issue (after quite some time) but did not act on it. That means a more capable SRE agent should be able to fix it.~

On second thought, asking the agent to fix something that is not even broken right now might not be fair. The agent does not even know that the `drain` command is failing. Without any visible symptoms or degradation, the agent has to chase random stuff. On the other hand, if the agent can see the error message for the `drain` command, the whole fault becomes quite trivial. The error message directly says `Cannot evict pod as it would violate the pod's disruption budget`. Seeing this, the agent can easily fix the fault. So, both not seeing the error and seeing the error pose a problem for the difficulty of this problem.

tianyin:
Hi @abdallahade1 

Your PR triggered a lot of discussion in our internal group.

It's a great PR and I personally love the problem. In fact, the problem describes the kinds of issues that I myself built my PhD thesis upon :) See my thesis work back in 2016: https://tianyin.github.io/pub/pcheck.pdf which is about how to find latent misconfigurations.

Unfortunately, currently, this kind of problems does not fit in SREGym's model, as SREGym asks the agent to resolve failures instead of discovering all latent issues. In your problem, it seems to us that there is no observable symptom, so we conclude that it is not fair to benchmark the SRE agents using this problem.

IMO this is a limitation, in terms of scope, of SREGym.

One solution is to re-construct the problem in the way that you introduce some observable failures and then the misconfiguration issues is a part of your story. But that would be a different problem construction.

I will close this issue and, as an evaluation task, I would count it done. Thanks for all the work!

If you could think about the new construction, I would be really appreciative!  

~t

abdallahade1:
@tianyin 
Thank you so much for your feedback, I'm really glad the problem sparked discussion. I'm reading your thesis now, the framing of latent misconfigurations as a distinct class to detect before they manifest is fascinating and it clarifies for me precisely why this problem sits awkwardly in SREGym's model, it's a latent fault with no active symptom until a disruption operation triggers it.
I'd love to think about the re-construction. My initial intuition is to keep the over-strict PDB as the root cause, but build the scenario around an operation that stays visibly stuck rather than failing once and going quiet, a rollout or scale-down that a controller keeps actively retrying, so there's an ongoing, observable "pods not progressing / node won't drain" symptom (with surfacing events) that an agent can detect and act on, while the PDB misconfiguration remains the underlying cause to fix. That would turn the latent issue into part of a story with a live, degrading symptom.
I'll think it through more carefully and follow up with a concrete proposal. Thank you again for the thoughtful feedback and for counting the task complete. I really enjoyed the work.

tianyin:
> so there's an ongoing, observable "pods not progressing / node won't drain" symptom (with surfacing events) that an agent can detect and act on, while the PDB misconfiguration remains the underlying cause to fix

This will be great. The only tricky part is that you will need to ensure the agent to use the drain command :)

PR #836, opened 2026-05-31
Add low_memory_limit_hotel_reservation: OOMKilled CrashLoop from a too-low container memory limit

TahseenSust:
```
/validate-problem low_memory_limit_hotel_reservation
```

This is scenario #8 ("What happens if pod resource limits are set too low?") from [Munduri Sai Kumar's *10 Real-World Kubernetes Failure Scenarios*](https://medium.com/@munduris939/10-real-world-kubernetes-failure-scenarios-every-devops-engineer-should-know-with-examples-7282d1a445c9). When a container's `resources.limits.memory` is set below what the process actually needs, the kernel OOM-kills it (exit 137), Kubernetes restarts the pod, it OOMs again, and the workload ends up in CrashLoopBackOff. The article's example is a Node.js app pinned to 128Mi that keeps crashing and taking the service down with it.

This failure mode isn't covered yet: `resource_request_*` faults at *scheduling* (pods stay Pending) and `valkey_memory_disruption` is an app-internal OOM. Here the pod schedules fine and is killed at *runtime*.

## How it's simulated

New problem `MemoryLimitOOMKill` on `hotel_reservation`, faulty service `mongodb-geo`.

- **Inject:** capture the container's original memory resources + replica count, then patch `limits.memory` down to `12Mi`. `mongod` needs far more than that, so it's OOM-killed on startup, deterministically.
- **Recover:** patch the original memory limit back.

Both use a strategic merge patch (`KubeCtl.patch_deployment`), so only `resources` changes. The container is discovered by name (the geo deployment's container is `hotel-reserv-geo-mongo`, not `mongodb-geo`), so inject/recover always hit the right container.

Diagnosis uses the standard `LLMAsAJudgeOracle`. Mitigation uses a custom oracle instead of the generic one: it requires the original deployment to still exist at its original replica count with all pods Running and Ready. This way a real fix (raise the limit) passes, but deleting or scaling the workload to zero doesn't.

## Runtime behavior

On `start`, the geo-mongo pod enters the OOM loop while everything else stays healthy:

```
NAME                           READY   STATUS      RESTARTS      AGE
mongodb-geo-75c787db66-9pcfh   0/1     CrashLoopBackOff   4 (64s ago)   101s
```

`kubectl describe` confirms the signature (killed the instant it starts):

```
Last State:  Terminated
  Reason:    OOMKilled
  Exit Code: 137
```

After raising the limit and letting it roll out, the mitigation oracle passes:

```
== Mitigation Evaluation (memory-limit OOMKill) ==
[OK] Target workload exists, is at full scale, and is no longer OOMKilling.
[EVAL] Mitigation Succeed
```

## Agent behavior

Ran Stratus (gpt-5) against the problem in the isolated agent container. The diagnosis agent scored 100/100: it listed pods, saw mongodb-geo OOMKilled, described the pod and deployment, and identified the 12Mi memory limit on container hotel-reserv-geo-mongo as the root cause. The mitigation agent applied the intended fix — patching the container's memory limit up to 512Mi/256Mi — and the pod returned to 1/1 Running, Ready. Across retries it never attempted a shortcut (no delete or scale-to-zero); it always fixed the root cause. The run did not register a passing mitigation due to a Stratus-side ClusterStateOracle connectivity flake (the 127.0.0.1:16443 filtering proxy refused connections at validation time), unrelated to the injected fault; the conductor-side mitigation oracle passes on the same fix when validated via the CLI.


HacksonClark:
This isn't really a "real failure". What we mean by real failure is that its sourced from a postmortem. We also are looking for difficult problems that agents can't solve yet -- but the agent solves this agent very easily! I also think it's a duplicate of another problem. Please feel free to try again soon! You can open an issue if you're unsure and want us to comment on your proposed problem and thought process.

tianyin:
Thank you for trying @TahseenSust !

As @HacksonClark mentioned, we are hoping that the failure scenario could reflect real-world failures, not synthetic ones. Otherwise, it is too simple :))

Give another try!

We encourage you to discuss the problem in https://github.com/SREGym/SREGym/issues first before you implement it (to ensure the problem is legit). In fact, the memory limit problem is already implemented in SREGym.

TahseenSust:
@tianyin Thank you for the feedback!

I’m trying to come up with a more realistic failure scenario. I initially looked into the Reddit PI-Day outage, but it seems someone has already opened a PR for that.

I’ll explore some other real-world incidents and discuss them in the issues section first before implementing anything.

If you have any suggestions for interesting real-world failure cases that would fit SREGym well, I’d really appreciate it if you could point me toward them.

Issue #842, opened 2026-06-01
[Problem RFC] A regenerated config artifact outgrows the consumer's compiled capacity


Neilblaze:
Hello! 👋🏼 

I am submitting/proposing this RFC as part of the UIUC++ SRSE evaluation task for a new SREGym benchmark problem. This is infact my second RFC (followed by #834), & just like last time, I would like to propose a problem that reproduces what I consider the most consequential instance of this class in recent memory -- a regenerated artifact doubling past a capacity that was baked into the consumer.

---

## Motivation

The real-world example that came to mind for me is the **Cloudflare outage of 2025-11-18**. A permissions change in a database caused a query to start returning duplicate rows, which doubled the size of the Bot Management "feature file" that Cloudflare's proxy loads. The proxy had preallocated for roughly 200 features. When the regenerated file exceeded that, the load path hit the limit and panicked (a Rust `.unwrap()` on the error result), and a large share of the network started returning 5xx. The file itself was valid. It was simply larger than the consumer was built to accept.

> [!NOTE]
> What makes this _failure mode_ interesting to me is that the failing component is downstream of the actual change, and the artifact is not corrupt. The team initially suspected a hyper-scale DDoS before tracing it to the feature file. I think that misdirection is exactly what makes it a good agent task, where the crashing worker is a symptom, the oversized artifact is the cause, and restarting the worker does nothing because the same artifact is loaded again.
‎ 
### I find this scenario valuable for a benchmark for two reasons:

1. The diagnosis is not "read the value, it is wrong." Every entry in the artifact is well-formed. The agent has to notice that the _count_ exceeds a limit the consumer enforces, which is a different signal from a parse error or a missing key.
2. My expectation is that the first reflex on a `CrashLoopBackOff` worker is to restart or recreate it. I specifically want a setup where that looks reasonable and fails, because the regenerated artifact is still mounted and the worker rejects it again on the next start. The fix has to act on the artifact (bring it back within capacity) or on the capacity itself, not on the pod.

### Adjacent hazards of the same class that I came across:

- Kubernetes caps a single ConfigMap or Secret at roughly 1 MiB, so a generated artifact that grows unbounded eventually fails to apply at all. The compiled-capacity limit I model here trips earlier and inside the consumer, which is closer to the Cloudflare mechanism.
- The same shape shows up with any preallocated buffer or fixed-size table that a config is loaded into, which is common in request-classification and policy-evaluation code paths.

---

## Proposal

[...]

PR #845, opened 2026-06-01
Add: oversized_config_artifact_hotel_reservation problem

Neilblaze:
This PR adds a new SREGym problem, `oversized_config_artifact_hotel_reservation`, as part of my evaluation task for UIUC++ SRSE that models a _configuration-pipeline failure_ on the Hotel Reservation app. 

A config-driven worker loads a generated feature list on startup and is built for a fixed number of entries. When an upstream step regenerates that artifact past the worker's compiled capacity, the worker aborts while loading it and enters `CrashLoopBackOff`. The artifact is well-formed, so this is not a malformed value or a missing key, where, the defect is the entry count exceeding a limit baked into the consumer. The rest of the namespace and the node stay healthy, and restarting the worker does not help, because the same oversized artifact is loaded again on the next start.

Related RFC: #842    /    Ref: https://github.com/SREGym/SREGym/issues/842#issuecomment-4592250727

/validate-problem oversized_config_artifact_hotel_reservation

---

A configuration consumer that preallocates for a fixed number of entries will reject an artifact that is valid but larger than that. The artifact passes every per-value check, so the failure shows up downstream of the change that produced it, in the code that loads it, not in the data itself.

As previously mentioned in #842, a notable real-world incident that immediately came to mind for me was the the **_Cloudflare outage of 2025-11-18_**. A database permissions change caused a query to return duplicate rows, which doubled the size of the Bot Management feature file the proxy loads. The proxy had preallocated for roughly 200 features. When the regenerated file exceeded that, the load path hit the limit and panicked (a Rust `.unwrap()` on the error result), and a large fraction of the network returned 5xx. The file was valid. It was simply larger than the consumer was built to accept, and regenerating or restarting did not help because the oversized file was still being served.

#### Related sources:

- Cloudflare, ["Cloudflare outage on November 18, 2025"](https://blog.cloudflare.com/18-november-2025-outage/) (official post-mortem). The duplicate-rows query doubled the feature file past the preallocated capacity, and the parser's `.unwrap()` on the size-limit error panicked the load path.
- [Hacker News discussion of the post-mortem](https://news.ycombinator.com/item?id=45973709), corroborating the hard-coded file-size-limit detail.
- [Kubernetes ConfigMap documentation](https://kubernetes.io/docs/concepts/configuration/configmap/), the adjacent platform-level limit (a ConfigMap or Secret is capped at roughly 1 MiB), which is the same class of failure one layer up.
‎ 

- **App:** Hotel Reservation
- **Injection layer:** Direct K8s API. No Helm changes, no node access, no privileged pods.

A small config-driven worker, `Deployment/traffic-classifier` (`busybox:1.36`), loads a feature list (`features.list` in `ConfigMap/traffic-classifier-config`) on startup. The worker is "compiled" for a fixed capacity of 200 entries, where it counts the non-empty lines in the artifact, and if the count exceeds 200 it logs the overflow and exits non-zero. This mirrors a real consumer preallocating a fixed-size table for the artifact. It writes a readiness marker only after a successful load, and a readiness probe gates on that marker, so the worker is never reported Ready while it is failing to load (it does not flicker Ready during the brief window each crash spends Running before it exits).
‎ 

1. Write the artifact regenerated far past capacity (400 entries) into the ConfigMap.
2. Deploy the worker. On startup it counts 400 entries against its compiled capacity of 200, logs `feature set 400 exceeds compiled capacity 200, aborting load`, and exits 1.
3. Block until the worker is actually in `CrashLoopBackOff`, so the fault is live before the diagnosis stage opens.

The `recover_fault` deletes the worker and its ConfigMap and waits for the deletion, which is idempotent and returns the namespace to its pre-injection state.

> [!NOTE]
> The worker is `traffic-classifier` and the artifact is a `features.list`, which read like an ordinary request-scoring component. The startup log states that the feature set exceeds the compiled capacity, which is the message a real consumer would emit rather than a benchmark giveaway.

### Oracle policy

The default `MitigationOracle` only checks that the namespace's pods are Running, which is wrong here in two directions: it would accept deleting the worker (the namespace then has only healthy pods left) and it would accept emptying the artifact (the worker then loads zero entries and stays Running while doing nothing). The new `OversizedConfigMitigationOracle` checks the invariant that actually defines the fix, where, the worker Deployment must exist and be Ready, and the artifact it loads must still carry entries.

| Fix | Accepted | Why |
|---|---|---|
| Bring the artifact back within capacity (reduce entries) and let the worker reload | Yes | This is Cloudflare's actual remediation: deploy a known-good, smaller artifact. Worker becomes Ready, artifact still populated. |
| Raise the consumer's compiled capacity | Yes | Equivalent to making the consumer tolerate the larger artifact. Worker becomes Ready, artifact still populated. |
| Delete the worker Deployment | No | Removes the workload instead of restoring it; oracle requires the Deployment to exist. |
| Scale the worker to zero | No | Stops the crash loop without restoring function; oracle requires at least one Ready replica. |
| Empty the artifact | No | Worker goes Ready but does nothing useful; oracle requires the artifact to remain non-empty. |
| Restart or recreate the worker only | No | The oversized artifact is still mounted, so the worker crash-loops again. |

‎ 

After injection, the Hotel Reservation pods stay healthy and the nodes have no `DiskPressure`. The only anomaly is `traffic-classifier` in `CrashLoopBackOff`.

```bash
# 1. Spot the crash-looping worker (restart count climbing)
kubectl get pods -n hotel-reservation -l app=traffic-classifier

# 2. Read the logs -> the artifact is too large, not malformed
kubectl logs -n hotel-reservation -l app=traffic-classifier --previous

# 3. Confirm exit code 1 / Error, not OOMKilled or Evicted
kubectl describe pod -n hotel-reservation -l app=traffic-classifier

# 4. Inspect the artifact the worker loads and count its entries
kubectl get configmap traffic-classifier-config -n hotel-reservation -o jsonpath='{.data.features\.list}' | grep -c .
```

```text
worker pod:       traffic-classifier-... 0/1 CrashLoopBackOff, restart count climbing
container logs:   loaded 400 features against compiled capacity 200
                  feature set 400 exceeds compiled capacity 200, aborting load
describe pod:     Last State Terminated, exit code 1 (not OOMKilled, not Evicted)
artifact size:    400 entries  <-- the worker accepts at most 200
node/namespace:   every other hotel-reservation pod Running, all nodes DiskPressure=False
```

The distinction from existing config problems is that the artifact is valid (no missing key, no bad value). Instead, the count exceeds a limit the consumer enforces, and the limit is the thing to find.

<br/>

## 4. The Agent behavior

A correct trajectory reads the worker's log, sees that it loaded more entries than its compiled capacity and aborted, inspects the artifact, and resolves the size mismatch, either by bringing the artifact back within capacity (the Cloudflare remediation) or by raising the consumer's capacity. The oracle then sees a Ready worker whose artifact still carries entries.

The oracle is built to reject the reflexes an agent reaches for first:
- Restart or recreate the worker. The oversized artifact is still mounted, so it crash-loops again.
- Delete the worker or scale it to zero. This stops the crash loop but removes the workload.
- Empty or delete the artifact. The worker goes Ready but no longer does anything.

## Validation

I ran the actual problem and oracle code against a live cluster (`inject_fault`, `recover_fault`, and `OversizedConfigMitigationOracle.evaluate()` through `uv run`), and also hardened one rough edge the run surfaced.

Mechanism, verified live:

| Claim | Observed |
|---|---|
| Worker aborts when the artifact exceeds capacity | `loaded 400 features against compiled capacity 200` then `feature set 400 exceeds compiled capacity 200, aborting load`; pod CrashLoopBackOff, exit code 1 |
| The crash-looping worker is never reported Ready | with the readiness probe, `readyReplicas` stayed 0 across an 8-poll, 40s window (0 Ready flickers) |
| Rest of the namespace and the node stay healthy | the worker is the only anomaly; no DiskPressure, no eviction, no OOM |
| `recover_fault` teardown is clean | worker Deployment and ConfigMap deleted |

‎ 
### Mitigation oracle matrix

| Scenario | Cluster action | `evaluate()` |
|---|---|---|
| Not mitigated | none | **FAIL**: worker not Ready (crash-looping on the oversized artifact) |
| Naive: restart the worker, artifact unchanged | rollout restart | **FAIL**: the same oversized artifact is loaded again |
| Correct fix: bring the artifact within capacity and reload | shrink to 60 entries and roll | **PASS** (worker Ready, artifact still carries 60 entries) |
| Reject: delete the worker | delete the Deployment | **FAIL**: the Deployment must exist |

> A config-only fix (shrink the artifact without restarting the worker) was also measured directly, where the crash-looping pod picked up the smaller artifact and became Ready in 5.1s, well within the oracle window. The oracle timeout was raised to 180s so that a correct fix is never graded a failure while the consumer waits out a longer CrashLoopBackOff before it reloads.

**UPDATE [03-06-26]:** The most recent commit/update contains a hardening applied on top of the previous commit. The consumer originally had no readiness probe, so the crash-looping pod was briefly counted Ready during the short window each restart spends Running before it exits with the overflow error. The oracle's consecutive-healthy-poll guard already absorbed this, but to remove the ambiguity the consumer now writes a readiness marker only after a successful load and a readiness probe gates on it. After the change the crash-looping pod never reports Ready (0 flickers, measured above), while a corrected artifact still drives it Ready.

### Diagnosis-stage grading (LLM judge)

The diagnosis oracle (`LLMAsAJudgeOracle` -> `DiagnosisJudge`) grades the submitted root cause against the expected one across three dimensions (Fault Localization, Fault Characterization, Scope Precision), with a 0.70 composite threshold. Run with `JUDGE_MODEL_ID=gpt-5.1`:

|---|---|---|
| Correct: CrashLoop from the artifact entry count exceeding the consumer's compiled capacity; node and namespace healthy; fix is to bring the artifact within capacity or raise it and reload | 1.00 | **True** (100/100, all nine checklist items) |
| Incorrect, near miss: a malformed/invalid feature value causing a parse error | 0.67 | **False** (localization 1.00, scope 1.00, characterization 0.00) |
| Incorrect, clear miss: the node is out of disk space | 0.00 | **False** (rejected on all three dimensions) |

Do note that, the near miss is the most informative result as it names the right component and the right scope, so it is rejected purely on characterization (oversized-but-valid versus malformed), which is exactly the distinction the problem turns on. The clear miss (a node resource problem) is rejected on every dimension.


Neilblaze:

Saadmrp1038:
@Neilblaze Thanks for the PR! Might take a week or two for me to review this :)

Neilblaze:
@Saadmrp1038 Cool, take your time, no hurries :)

Also, in reference to https://github.com/SREGym/SREGym/issues/842#issuecomment-4592187163, 
> If you want to create multiple you are more than welcome! More good faults mean better for us :3

I was just curious if it'd be fine to raise another issue as well? :3
I feel it's worth working on. But yeah, you can always take your time and review things at your own pace.

Lmk if that sounds good?

Saadmrp1038:
@Neilblaze Of course! Feel free to raise more if you want.

Neilblaze:
Hi @Saadmrp1038!, 

Hopefully this is still in the queue, and you haven't started the review yet. I just pushed a new commit ([`0955d27`](https://github.com/SREGym/SREGym/pull/845/commits/0955d2746bf3d9749e49164a2b9886dd53ad37a2)), which adds additional case hardening on top of [`0e1e0d3`](https://github.com/SREGym/SREGym/pull/845/commits/0e1e0d399169e22b82213e8adb64df78983a9ca7).

Also, followed by yesterday's incident (https://github.com/SREGym/SREGym/pull/838#issuecomment-4607862603), I've thoroughly audited the changes for #845, followed by #848 on my end. Everything has been rechecked carefully, and I don't expect a recurrence of the issues we encountered previously.

Still, if anything else is required from my side, please do a tag. Thanks!

tianyin:

Neilblaze:
@tianyin, this PR was ready for review, but it looks like it got lost in the pile. I had already finished working on it, so it should be ready to review/merge if everything looks good.

HacksonClark:
/evaluate-problem oversized_config_artifact_hotel_reservation

HacksonClark:
/evaluate-problem oversized_config_artifact_hotel_reservation

HacksonClark:
/evaluate-problem oversized_config_artifact_hotel_reservation

HacksonClark:
/evaluate-problem oversized_config_artifact_hotel_reservation

HacksonClark:
/evaluate-problem oversized_config_artifact_hotel_reservation

HacksonClark:
Closing due to saturation.

PR #847, opened 2026-06-01
Add mutating_webhook_resource_limits problem

keremakcali:
The motivating failure case for this implementation involves a mutating admission webhook silently rewriting pod memory limits at admission time, applying a value far higher than what was expected and breaking deployments via namespace quota rejection. The Deployment spec was untouched and showed the configured limit; the wrong value only existed on the mutated pod and matched nothing in the cluster's declared configuration.

The same mechanism — a webhook silently rewriting pod resource limits at admission time — can fail in either direction. The incident above showed it with a value too high; this problem simulates the inverse, where the injected value is too low and the pod OOMKills on startup. A low limit was chosen because reliably reproducing the quota-rejection symptom depends on the cluster having tight quotas, which isn't guaranteed across environments, whereas an aggressively low limit deterministically triggers OOMKill on any reasonable workload.

**Reference:** 
https://github.com/kubernetes/autoscaler/issues/8401

## How the failure is simulated in SREGym

**Problem class:** `sregym/conductor/problems/mutating_webhook_resource_limits.py`
**Registry key:** `mutating_webhook_resource_limits_social_network`

The fault builds a fully self-contained webhook backend in `platform-ops` — a `python:3.12-alpine` container running a small HTTPS server whose code lives in a ConfigMap, so no custom image or registry is needed. Fresh TLS material is generated on every injection: one CA, one server cert signed by that CA, and the correct `caBundle` wired into the `MutatingWebhookConfiguration` so TLS validates and admission always goes through.

Once the backend is up, a `MutatingWebhookConfiguration` named `pod-policy.platform.k8s.io` is created scoped to `social-network` via `namespaceSelector`, with `failurePolicy: Fail` and a `/mutate` handler. From that point every pod CREATE in the namespace gets a JSON patch injecting `16Mi` as the first container's memory limit. One `nginx-thrift` pod is then deleted to trigger an immediate mutated replacement and make the symptom visible right away.

The patch logic covers three cases depending on what the incoming pod spec has — no resources field, resources but no limits, or limits already present — using the `add` op throughout to avoid JSON Patch path-not-found errors.

`INJECTED_MEMORY_LIMIT = "16Mi"` is a class constant so the injected value is defined once and flows through the webhook server code, the root cause description, and all print statements automatically.

**Credits:**
The overall structure of the problem class and the fault injection pattern follow the approach established by @mohamedharake in `admission_webhook_tls_mismatch` (#777).

## Runtime behaviour

Tested via the SREGym CLI on a local Kind cluster.

| Phase | Observation |
|-------|-------------|
| App deploy | `social-network` deploys normally before fault injection |
| Backend setup | `platform-ops/platform-policy-controller` is running as a reachable HTTPS service |
| Fault injection | One `nginx-thrift` pod is deleted; replacement is created and immediately mutated |
| Symptom | `nginx-thrift` pod repeatedly enters `OOMKilled` / `CrashLoopBackOff` |
| Diagnostic trap | `kubectl get deployment nginx-thrift -o yaml` shows no memory limit set |
| Diagnostic signal | `kubectl describe pod <nginx-thrift-xxx>` shows `Limits: memory: 16Mi` |
| Blast radius | Scoped to `social-network` via `namespaceSelector`; rest of cluster unaffected |
| Recovery | Deleting the `MutatingWebhookConfiguration` and restarting the deployment restores normal behaviour |

Checking the Deployment spec returns nothing for resources, but inspecting the actual running pod shows `16Mi` injected under limits. The spec and the pod disagree — which only happens if something modified the pod between spec and runtime.

## Expected agent behaviour

Tested end-to-end with the `claudecode` agent (`claude-sonnet-4-6`). Diagnosis passed (composite 0.78), mitigation passed. TTL 97.6s, TTM 167.4s.

Expected successful agent path:

1. Notice `nginx-thrift` is repeatedly OOMKilling.
2. Check `kubectl describe pod` on the crashing pod and find `Limits: memory: 16Mi`.
3. Check the Deployment spec and find no memory limit set — notice the discrepancy.
4. Suspect admission-time mutation as the only explanation for spec and pod disagreeing.
5. Run `kubectl get mutatingwebhookconfigurations` and find `pod-policy.platform.k8s.io`.
6. Inspect the webhook configuration and confirm it patches memory limits in `social-network`.
7. Mitigate by deleting the webhook configuration and restarting the deployment.

The intended challenge is that the visible symptom looks like an application problem — OOMKill, CrashLoopBackOff — but the root cause is a cluster-scoped admission webhook that is invisible from those angles and only discoverable by comparing what the Deployment spec declares against what the pod actually runs with.

/validate-problem mutating_webhook_resource_limits_social_network


Saadmrp1038:
@keremakcali Can you please link some real-world incident stories that inspired this fault? Not generic troubleshooting guides. Preferably post-mortems. Without an incident to ground it in, it's harder to judge the relevance of a fault. Although I do love that the diagnosis for this fault is not quite straight forward!

keremakcali:
Oh, I'm so sorry. I pasted the wrong reference link. I unfortunately don't have an official post-mortem because I had a hard time finding one that wasn't already implemented or mentioned in posted RFCs, and that I could understand well enough to believe I could implement something reliably (or i am bad at googling). So I'm using this GitHub issue as the basis of this implementation instead. Also, I just updated the PR text to better reflect how i decided to go through with this implementation. I hope it's alright @Saadmrp1038 .

Saadmrp1038:
@keremakcali That's alright. I will need to review to see how the agent performs on this fault over multiple runs. It seems good at first glance. Maybe we can tweak some stuff to make it harder.

keremakcali:
> Maybe we can tweak some stuff to make it harder.

@Saadmrp1038  Sounds great! Thanks a lot for the reply.

Saadmrp1038 (changes requested):
@keremakcali Thanks for the effort! The implementation looks good. I tested it end-to-end and everything works correctly.

One thing is as it is now the problem is too easy. There are a couple of tweaks that can be done to make it harder for the agent to solve:

1. Consider patching the `nginx-thrift` deployment with an explicit memory limit (e.g. `256Mi` limit / `128Mi` request) BEFORE activating the webhook, so the deployment spec shows a real value. Right now `resources: {}` in the spec immediately signals something injected this field when the pod shows `16Mi`. The webhook currently only patches `limits.memory`. If you set both requests and limits in the spec, you'll need the webhook to patch both fields (otherwise Kubernetes rejects the pod with `requests > limits`). This also feels like a more realistic patch.

2.  Consider installing 3-4 decoy MutatingWebhookConfigurations (e.g. cert-manager, istio, kyverno names). Although make sure they don't target/affect anything.

Please ping me when you've addressed them.

keremakcali:
Hey @Saadmrp1038, I implemented these changes. Tried it out with the agent and it seems like it still can solve it. I tried to address some minor details it had used in the runs as well with not much success. The main thing that makes it easy for the agent seems to be how easy it is to notice a single OOMKilling pod that has a different memory request/limit from its spec. All the other complexities just increase the runtime until it finally finds the correct webhook that mutates it from decoys.  I have an idea to make this aspect harder, so I wanted to ask if it makes sense:

I was planning to add more pods that get affected by this mutating webhook that injects a very low memory limit. In this case some pods will OOMKill but some will not because they don't require as much memory. I did one such run and the agent did fix the OOMKilling pods but didn't notice the non-OOMKilling pods. Maybe the mitigation oracle should check if it addressed all affected pods. This part is where I am not sure if it makes sense though, because it feels a little bit like an artificially created trap. But at the same time, if those limits persist they might eventually crash due to things like traffic or other stuff that increase their memory usage, which makes them worth checking in the oracle. I would appreciate your feedback on this.

Saadmrp1038:
@keremakcali The mitigation is easy that I know. Does it get diagnosis right too? My thought was the agent would see another plausible hypothesis and submit that as root cause without exploring all the options.

keremakcali:
@Saadmrp1038  Yeah, actually in diagnosis a lot of the time it frames the nginx-thrift as the faulty component even though it is the victim. I think in one run it thought a decoy MWC was the culprit. Those are the problems that i saw about diagnosis. It almost always figures out the right MWC in the mitigation phase eventually tho. I am not super sure what to get of all these but yeah diagnosis is usually problematic. I was mostly focusing on the mitigation...

Saadmrp1038:
@keremakcali Diagnosis failing but mitigation succeeding is also an interesting thing imo. The agent, when diagnosing, initially finds a plausible reason and stops. It doesn't try to make sure there isn't any other possible cause for the issue. But when mitigating, when the mitigation strategy based on the diagnosis fails, the agent then expands its search and finds out that the root cause is somewhere else.

---
That's why previous suggestions were mostly about adding distractors and not about making the mitigation harder. Because I don't think we can make the mitigaiton for this fault that hard anyway.

keremakcali:
@Saadmrp1038  Hmm.. That makes a lot of sense. In that case i think i will be pushing the changes and we see what happens.

keremakcali:
Hi @Saadmrp1038 just pushed the implementation addressing the tweaks.

Saadmrp1038 (approved):

Saadmrp1038:
@keremakcali Looks good to me. Claude Code most of the time fails diagnosis by a few points due to not targeting the webhook configuration itself as the faulty component but rather blaming the backend service (platform-policy-controller). Stratus fails more drastically as it never checks mutating webhooks at all and submits a wrong diagnosis blaming the pod/deployment configuration.

Issue #846, opened 2026-06-01
[Problem RFC] A database connection cap set too low starves a service's connection pool


Neilblaze:
Hello! 👋🏼 

I am submitting/proposing this RFC as part of the UIUC++ SRSE evaluation task for a new SREGym benchmark problem.  

I noticed that the MongoDB-related problems we already have cover authentication (`revoke_auth`, `auth_miss_mongodb`) and data or version issues (`update_incompatible_correlated`), but none of them cover the database refusing to accept *connections* because its connection limit is set too low for the client that depends on it. That is a different failure from "the database is down" or "auth is wrong", while the database process is perfectly healthy, & it is simply turning away new connections.

> I may be over-separating this from the existing resource problems, but I think the diagnosis path is genuinely distinct. The database pod and the service pod are both Running and Ready, the node is fine, and yet the service's queries fail. The signal is not a crash or an eviction, it is a connection being refused before it is even authenticated.

Hence, I would like to propose a problem that reproduces the classic database connection-pool-exhaustion failure.

---

## Motivation

This is one of the most common ways a healthy database takes down the services in front of it. A fleet of application replicas, each holding a client-side connection pool, collectively opens more connections than the server is configured to accept. Once the server hits its limit it refuses new connections with an error like "too many open connections", and every request that needs a fresh connection fails, even though the database itself is idle and healthy.

The class is significant enough at scale that the industry built managed infrastructure specifically to absorb it. AWS RDS Proxy exists because serverless and containerized fleets routinely exhaust a database's connection limit, and PgBouncer occupies the same role for PostgreSQL. The MongoDB server has a direct knob for this limit (`net.maxIncomingConnections`, exposed on the command line as `--maxConns`), and setting it too low for the workload is a documented way to cause exactly this outage.

> [!NOTE]
> What makes this a good agent task, to me, is the misdirection. The crashing or erroring component is the *service*, but the service is not the problem. The database is "up" by every pod-level signal. The agent has to look at why the database is refusing connections and find a connection cap that is too small, which is not where a pod-centric investigation naturally leads.

I find this scenario valuable for the following reasons:

1. Both pods stay Running and Ready. In the Hotel Reservation app neither `mongodb-geo` nor `geo` carries a health probe, so `kubectl get pods` shows everything green while the geo service cannot get the database connections it needs.
[...]

PR #848, opened 2026-06-01
Add: connection_pool_exhaustion_hotel_reservation problem

Neilblaze:
This PR adds a new SREGym problem, `connection_pool_exhaustion_hotel_reservation`, as part of my evaluation task for UIUC++ SRSE that models a database connection-exhaustion failure on the Hotel Reservation app. 

The `mongodb-geo` database is started with an incoming-connection cap far below the connection pool that the `geo` service maintains against it. Once the service needs more than a handful of connections, the database refuses new ones with "too many open connections" and the service's queries fail. Both the database pod and the service pod stay Running and Ready, because neither carries a health probe, so `kubectl get pods` shows nothing wrong. Restarting either pod does not help, because the cap is part of the database's startup configuration and is still in effect after the restart.

Related RFC: #846

/validate-problem connection_pool_exhaustion_hotel_reservation

---

A healthy database can take down the services in front of it by running out of connection slots. Each application replica holds a client-side connection pool, and the fleet collectively opens more connections than the server is configured to accept. Once the server hits its limit, it refuses new connections, and every request that needs a fresh connection fails, even though the database is idle and up.

The class is significant enough at scale that the industry built managed infrastructure specifically to absorb it, and in fact, AWS RDS Proxy exists because serverless and containerized fleets routinely exhaust a database's connection limit, and PgBouncer plays the same role for PostgreSQL. MongoDB exposes the limit directly as `net.maxIncomingConnections` (`--maxConns` on the command line), and setting it too low for the workload reproduces the outage exactly.

### Related sources:

- [AWS RDS Proxy](https://aws.amazon.com/rds/proxy/), managed connection pooling built to absorb connection exhaustion from large client fleets.
- [MongoDB manual: `net.maxIncomingConnections` / `--maxConns`](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-net.maxIncomingConnections).
- [PostgreSQL connection pool exhaustion, lessons from a production outage](https://www.c-sharpcorner.com/article/postgresql-connection-pool-exhaustion-lessons-from-a-production-outage/).
‎ 

- **App:** Hotel Reservation
- **Injection layer:** Direct K8s API. The problem patches the `mongodb-geo` container's startup arguments. No Helm changes, no node access.

1. Snapshot the current `mongodb-geo` container arguments (`["--auth"]`).
2. Append `--maxConns 5` and roll the database out. 5 is the smallest value mongod accepts (it refuses to start below that), and far below the driver pool the `geo` service opens, so the cap is pathological while the database still runs.
3. Confirm the cap is active by opening a burst of connections and observing the database refuse them after a few.

The `recover_fault` restores the original arguments from the snapshot and rolls the database back out, which is idempotent.

> [!NOTE]
> In relation to the said context, the `mongodb-geo` and `geo` are the app's own components, and `--maxConns` is a real mongod flag, so the agent simply sees a database configured with a low connection limit, which is what the real incident looks like.

### Oracle policy

The default `MitigationOracle` only checks that pods are Running, which is useless here, given that both pods are Ready throughout. The mongod `serverStatus` command is also gated behind authentication on this deployment, so the oracle cannot read the configured limit directly. Instead, the `ConnectionPoolMitigationOracle` verifies the cap functionally, as it opens a burst of 30 connections to the database and requires at least 25 to succeed. A connection counts against the cap at establishment time, before authentication, so the probe needs no credentials.

| Fix | Accepted | Why |
|---|---|---|
| Raise `--maxConns` to a sane value and roll out | Yes | The database can serve the service's pool again; the probe opens the full burst. |
| Remove the `--maxConns` cap and roll out | Yes | Restores the default limit; the probe opens the full burst. |
| Restart the service or the database without changing the cap | No | The cap is part of startup config and persists; the probe is still refused early. |
| Scale `mongodb-geo` to zero | No | Stops the errors by removing the database; the oracle requires it Ready. |
| Delete `mongodb-geo` | No | Removes capacity rather than restoring it; the oracle requires the Deployment to exist. |

‎ 

After injection, the application pods stay healthy and the nodes are fine. The geo service's database calls fail under load while every pod reports Ready.

```bash
# 1. Pods look healthy: mongodb-geo and geo are both Running/Ready
kubectl get pods -n hotel-reservation -l io.kompose.service=mongodb-geo
kubectl get pods -n hotel-reservation -l io.kompose.service=geo

# 2. The database is refusing connections beyond its cap
kubectl logs deploy/mongodb-geo -n hotel-reservation | grep -i "too many open connections"

# 3. The cap is in the database's startup args
kubectl get deploy mongodb-geo -n hotel-reservation -o jsonpath='{.spec.template.spec.containers[0].args}'

# 4. Confirm functionally: only a few connections open before refusal
kubectl exec deploy/mongodb-geo -n hotel-reservation -- mongo --quiet --host 127.0.0.1 \
  --eval 'var c=[];var ok=0;for(var i=0;i<30;i++){try{var m=new Mongo("127.0.0.1:27017");m.getDB("admin").runCommand({ping:1});c.push(m);ok++;}catch(e){break;}}print("opened="+ok)'
```

```text
pods:          mongodb-geo 1/1 Running, geo 1/1 Running  (both Ready, no probes)
mongod log:    "Connection refused because there are too many open connections","attr":{"connectionCount":6}
args:          ["--auth","--maxConns","5"]
probe:         opened=4   (a healthy server opens all 30)
```

The key distinction from the existing MongoDB problems is that this is neither an auth failure nor a data problem, rather, the database is healthy and simply turning away connections because its limit is too small for its client.

‎ 

## 4. The Agent behavior

A correct trajectory notices that the geo service's database operations fail while both pods are Ready, looks at the database rather than the service, finds the database refusing connections (in its log and its low connection burst), traces it to the `--maxConns` cap, and raises or removes it so the database rolls out with room for the service's pool.

The oracle rejects the reflexes an agent reaches for first:
- Restart the service or the database. The cap is part of startup config, so it is still in effect after the restart and the probe is refused again.
- Scale the database to zero or delete it. This stops the errors but removes the database, while, the oracle requires it Ready and present.

## Validation

I ran the actual problem and oracle code against a live cluster (`inject_fault`, `recover_fault`, and `ConnectionPoolMitigationOracle.evaluate()` through `uv run`).

Mechanism, verified live:

| Claim | Observed |
|---|---|
| Healthy database accepts a full connection burst | probe opens 30/30 before injection |
| `--maxConns` below 5 is rejected by mongod | `--maxConns 2` -> `BadValue: maxConns has to be at least 5`, so the injector uses 5 |
| Cap at 5 keeps both pods Running and Ready | `mongodb-geo` and `geo` both Ready, 0 restarts; geo's startup opens a single connection, within the cap, so it registers in consul and serves rather than crashing |
| Cap refuses connections beyond the limit | injection confirmed the cap active: the probe opens only 3 before mongod refuses |
| `recover_fault` restores the original args | args back to `["--auth"]`, probe opens 30/30 |

‎ 
Mitigation oracle, verified live on live cluster:

| Scenario | Cluster action | mongodb-geo / geo | probe | `evaluate()` |
|---|---|---|---|---|
| Not mitigated | inject | Ready / Ready | 3/30 | **FAIL** |
| Naive: restart the database, cap unchanged | rollout restart | Ready / Ready | refused | **FAIL** |
| Correct fix: remove the cap and roll out | strip `--maxConns` and roll | Ready / Ready | 30/30 | **PASS** |
| Reject: scale the database to zero | scale to 0 | scaled to 0 / Ready | n/a | **FAIL** |

### Diagnosis-stage grading (LLM judge)

The diagnosis oracle (`LLMAsAJudgeOracle` -> `DiagnosisJudge`) grades the submitted root cause against the expected one with a three-dimension checklist (Fault Localization, Fault Characterization, Scope Precision) and a 0.70 composite threshold. Run with `JUDGE_MODEL_ID=gpt-5.1`:

|---|---|---|
| Correct: geo's queries fail because mongodb-geo is capped at `--maxConns 5`, far below geo's pool, so mongod refuses connections; both pods Ready; fix is to raise or remove the cap and roll out | 1.00 | **True** (100/100, all nine checklist items pass) |
| Incorrect: attributes the failure to a CNI / NetworkPolicy dropping traffic to mongodb-geo | 0.11 | **False** (11/100; localization 0.00, characterization 0.00, scope 0.33) |

The correct diagnosis clears every checklist item. The network-blaming diagnosis earns partial localization credit for naming mongodb-geo but is rejected on characterization and scope, landing well below the threshold, which confirms the expected root cause separates a connection-cap fault from a plausible network misdiagnosis.


Neilblaze:

tianyin:

Neilblaze:
@tianyin, this PR was ready for review, but it looks like it got lost in the pile. I had already finished working on it, so it should be ready to review/merge if everything looks good.

munimthahmid:
@Neilblaze No worries. Will review it.

HacksonClark:
/evaluate-problem connection_pool_exhaustion_hotel_reservation

HacksonClark:
/evaluate-problem connection_pool_exhaustion_hotel_reservation

HacksonClark:
From the workflow -- it seems that this problem is already solved by frontier models. I believe it passes the quality bar for the REU task, but I will have to close the PR as we can only include problems that challenge frontier agents.

Neilblaze:
Understandable, I guess the result could’ve been different if it had been attempted ~3 months ago 🥀 … but thanks for the run, @HacksonClark :)

HacksonClark:
> Understandable, I guess the result could’ve been different if it had been attempted ~3 months ago 🥀 … but thanks for the run, @HacksonClark :)

The frontier moves fast!

Issue #213, opened 2025-08-10
Remove wandb


HacksonClark:
wandb doesn't really make sense to have anymore now that the agent is completely decoupled from the benchmark.

PR #851, opened 2026-06-02
Add: secret_rotation_stale_env_credentials_astronomy_shop problem

### LINKED ISSUE #822: [Problem RFC] Stale Secret Environment Variable After Database Credential Rotation
Hi! I am working on the evaluation task, and I have an idea for a new benchmark problem.

My idea for the issue is a credential-rotation failure, when a backend password and Secret that stores this password are changed, but the pod keeps using the old credentials which were stored in the Secret before the rotation. Similar issue of stale credentials caused outage of multiple Cloudflare services ([Link](https://blog.cloudflare.com/cloudflare-incident-march-21-2025/)). Only nuance is that this particular incident is not described well, so to simulate the issue when pod uses stale credentials I plan to use fact that if the container uses a Secret stored as an environment variable, the Secret update will not be seen by the container unless it is restarted ([Link](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/)).

How I plan to simulate this failure on SREGym:

1. Use the `astronomy-shop` application, probably `product-catalog` with PostgreSQL.
2. Patch `product-catalog` so `DB_CONNECTION_STRING` comes from a Kubernetes Secret, then start it once with the old working password.
3. Rotate the PostgreSQL password and update the Secret to the new connection string, but do not restart `product-catalog`.
4. Terminate existing PostgreSQL sessions so `product-catalog` has to reconnect using its stale old credential.

To diagnose this issue, the agent is expected to go beyond the simple check of pods since with this issue they would be all running, and find that the DB-backed product behavior is failing. It could then trace the failure to `product-catalog`, inspect how `DB_CONNECTION_STRING` is provided, check that the Kubernetes Secret and PostgreSQL now use the new password, and infer that the running pod still has the old Secret value because it was not restarted after the credential rotation. To mitigate the problem, agent should just restart the `product-catalog` pod.

Let me know what you think about this problem and if it is a good idea for a benchmark problem!

Ipetr14:
Closes #822 

## Problem

This PR adds `secret_rotation_stale_env_credentials_astronomy_shop` issue. The problem targets `deployment/product-catalog` in namespace `astronomy-shop`. Injection rotates the PostgreSQL password and updates the Secret-backed `DB_CONNECTION_STRING`, but the already-running `product-catalog` pod keeps the old environment value loaded at container startup. The Secret and PostgreSQL backend use the new credential; the active `product-catalog` process still uses the old one.

## Real-world failure story

This models a mechanism of some real issues: credential/configuration material is updated, but a long-running process does not use the new value until it is restarted. A close mechanism appears in Laravel Horizon issue #213, where database credentials changed in `.env` were not picked up until the processes were manually restarted:

https://github.com/laravel/horizon/issues/213

In our case, mechanism causes one of the pods to use wrong credentials to access the database. This may cause outages of various services of the application like in this case (the mechanism of the issue is different, but similar consequnces):

https://blog.cloudflare.com/cloudflare-incident-march-21-2025/

## Fault injection

The injection first restores the original, correct state of the application:

1. Restore PostgreSQL user `otelu` to the original password `otelp`.
2. Set `product-catalog` back to a literal baseline `DB_CONNECTION_STRING`.
3. Delete `secret/product-catalog-db-conn` if it exists.
4. Restart `deployment/product-catalog`.

Then it injects the fault:

| Step | Action                                                                              | Result                                                           |
| ---- | ----------------------------------------------------------------------------------- | ---------------------------------------------------------------- |
| 1    | Create `secret/product-catalog-db-conn` with the original connection string.        | The Secret initially matches the old PostgreSQL password.        |
| 2    | Patch `deployment/product-catalog` so `DB_CONNECTION_STRING` comes from the Secret. | New pods read DB credentials through `valueFrom.secretKeyRef`.   |
| 3    | Restart `product-catalog` and record the current pod UID.                           | The active pod has the old connection string in its process env. |
| 4    | Rotate PostgreSQL from `otelp` to `otelp_7k9m2q4x`.                                 | PostgreSQL now accepts only the new password.                    |
| 5    | Update `secret/product-catalog-db-conn` with the new connection string.             | The Kubernetes Secret contains new password.                     |
| 6    | Terminate existing PostgreSQL sessions for `otelu`.                                 | `product-catalog` must reconnect.                                |
| 7    | Wait until the expected stale credential state is confirmed.                                               | Secret/PostgreSQL are new; running `product-catalog` env is old. |

Expected active fault state:

```text
Secret:          postgres://otelu:otelp_7k9m2q4x@postgresql/otel?sslmode=disable
PostgreSQL:      accepts otelp_7k9m2q4x, rejects otelp
product-catalog: still uses postgres://otelu:otelp@postgresql/otel?sslmode=disable
```

## Observable symptoms

| Symptom                                                                            | Command                                                                                                                                    | Expected evidence                                                                   |
| ---------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------- |
| Pods health looks normal  | `kubectl get pods -A`                                                                                                       | `product-catalog` and the other app pods are `Running`.                   |
| `product-catalog` is configured to read DB credentials from the Secret.            | `kubectl get deploy product-catalog -n astronomy-shop -o yaml \| grep -A8 -B4 DB_CONNECTION_STRING`                                        | `DB_CONNECTION_STRING` uses `valueFrom.secretKeyRef.name: product-catalog-db-conn`. |
| The Secret contains the rotated DB credential.                                     | `kubectl get secret product-catalog-db-conn -n astronomy-shop -o jsonpath='{.data.DB_CONNECTION_STRING}' \| base64 -d; echo`               | `postgres://otelu:otelp_7k9m2q4x@postgresql/otel?sslmode=disable`                   |
| The running `product-catalog` process still has the stale credential.              | `kubectl debug -n astronomy-shop -it "$(kubectl get pod -n astronomy-shop -l app.kubernetes.io/name=product-catalog -o jsonpath='{.items[0].metadata.name}')" --image=busybox:1.36 --target=product-catalog --profile=sysadmin -- sh -c 'tr "\0" "\n" < /proc/1/environ \| grep "^DB_CONNECTION_STRING="'`                                                   | `DB_CONNECTION_STRING=postgres://otelu:otelp@postgresql/otel?sslmode=disable`                            |
| PostgreSQL rejects the stale password.                                             | `kubectl exec -n astronomy-shop deploy/postgresql -- sh -lc 'PGPASSWORD=otelp psql -h postgresql -U otelu -d otel -c "select 1"'`          | Authentication fails.                                                               |
| PostgreSQL accepts the rotated password.                                           | `kubectl exec -n astronomy-shop deploy/postgresql -- sh -lc 'PGPASSWORD=otelp_7k9m2q4x psql -h postgresql -U otelu -d otel -c "select 1"'` | Query succeeds.                                                                     |

## Expected mitigation

Valid mitigation would be to restart the `product-catalog` pod so it would pick the correct credentials from env Secert:

```bash
kubectl rollout restart deployment/product-catalog -n astronomy-shop
kubectl rollout status deployment/product-catalog -n astronomy-shop --timeout=180s
```

## Mitigation oracle

The oracle checks:

| Check                                       | Required passing state                                         |
| ------------------------------------------- | -------------------------------------------------------------- |
| `deployment/product-catalog` exists         | Deployment can be read and parsed.                             |
| Product-catalog pods are healthy            | Current pods are `Running`.                        |
| Deployment still references the Secret      | `DB_CONNECTION_STRING` uses `product-catalog-db-conn`.         |
| Secret value is rotated                     | Secret contains `otelp_7k9m2q4x`.                              |
| Running product-catalog env is rotated      | Effective runtime connection string contains `otelp_7k9m2q4x`. |
| PostgreSQL accepts product-catalog password | Password parsed from product-catalog env succeeds.             |
| PostgreSQL rejects old password             | `otelp` no longer authenticates.                               |
| PostgreSQL accepts new password             | `otelp_7k9m2q4x` authenticates.                                |

## Recovery

`recover_fault()` restores the original application state:

1. Restore PostgreSQL user `otelu` to password `otelp`.
2. Set `deployment/product-catalog` back to the literal original connection string.
3. Delete `secret/product-catalog-db-conn`.
4. Restart `deployment/product-catalog`.
5. Clear the stored stale pod UID.

This returns the app to its original, correct state.

## Testing and validation

Implementation of the issue passed automatic validation. Submission during the mitigation stage without any fixes fails, and passes with the implemented solution during mitigation stage. All expected symptoms present after the fault injection.

## Agent run results

I ran one Stratus attempt with `gpt-5` on my local deployment. It completely failed the diagnosis (0.0/100.0) and mitigation blaming the telemetry issues and trying to fix them. The nuance is that for some reasons my local deployment actually has telemetry issues in the `astronomy-shop` that show up when I check the logs of `deployment/product_catalog` (for multiple other problems that use `astronomy-shop` as well), so this noise is affecting the work of the agent. For this reason, I am not sure how agent would actually score on this problem without this additional noise.

## Code changes

| File                                                                 | Change                                                                                                                 |
| -------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- |
| `sregym/conductor/problems/secret_rotation_stale_env_credentials.py` | Adds the problem class, injection, recovery, and diagnosis oracle wiring.                                    |
| `sregym/conductor/oracles/secret_rotation_stale_env_mitigation.py`   | Adds a mitigation oracle to check if the issue was fixed |
| `sregym/conductor/problems/registry.py`                              | Registers `secret_rotation_stale_env_credentials_astronomy_shop`.                                                      |

/validate-problem secret_rotation_stale_env_credentials_astronomy_shop


Saadmrp1038 (changes requested):
@Ipetr14 Thanks for the effort!

I had some concerns:

1. The `postgresql-init` ConfigMap contains `CREATE USER otelu WITH PASSWORD 'otelp';` in plain text. An agent can read this, compare it to the Secret, and diagnose the issue as a simple password mismatch without ever reasoning about stale process environments or runtime vs. declared state.

2. PostgreSQL has no persistent storage in this deployment. If the PostgreSQL pod restarts, `init.sql` re-runs and resets the password to `otelp`. The rotation is non-durable. The forward-fix (restart product-catalog to pick up `otelp_7k9m2q4x`) breaks as soon as PostgreSQL restarts because PG will only accept `otelp` again.

3. The oracle does not verify that the rotated password (`otelp_7k9m2q4x`) still works in PostgreSQL after mitigation. An agent can `ALTER USER otelu WITH PASSWORD 'otelp'`, update the Secret to match, restart the pod, and pass the oracle. This is dangerous because `otelu` is shared infrastructure (`accounting` also uses it), and modifying database credentials risks cascading failures. The oracle should reject any mitigation that changes the PostgreSQL password.

Ipetr14:
Thank you for the review! I have made some updates to address all three concerns:

- Now during the injection the ConfigMap is changed so `init.sql` declares `otelu` with the rotated password to remove the shortcut
- This change of ConfigMap also makes the rotation durable, and if the PostgreSQL pod is restarted the new password remains the correct one
- The oracle now checks that after the submission of mitigation the PostgreSQL accepts the rotated password and not the old one

During testing I also noticed that pods `accounting` and `product-reviews` have old credentials after the rotation during the injection, so to fix this issue and avoid leaking the old password to the agent through these pods, during the injection the credentials of these pods are also rotated.

sregym/conductor/problems/secret_rotation_stale_env_credentials.py
```diff
@@ -0,0 +1,355 @@
...
+                break
+            time.sleep(self._POSTGRES_PASSWORD_CHECK_INTERVAL_SECONDS)
+        logger.info("[secret-rotation] Verification: %s", state)
+        if not all(state.values()):
+            raise RuntimeError(f"Fault verification failed; stale env state was not created: {state}")
+
+        print(f"Service: {self.faulty_service} | Namespace: {self.namespace}\n")
+
+    @mark_fault_injected
+    def recover_fault(self):
```
Saadmrp1038, line 351: The mitigation oracle assumes the injected state (rotation) is present when it runs. After `recover_fault()` the oracle fails because PG no longer accepts `otelp_7k9m2q4x`. This causes issues when doing automated testing where I try to check if mitigation oracle will pass or not after calling `inject_fault` and `recover_fault` simultaneously. The `recover_fault` should work in such a way that afterwards mitigation oracle passes.

Saadmrp1038 (changes requested):
@Ipetr14 The changes look good. There's still one issue with how `recover_fault` works. I commented the details below.

sregym/conductor/problems/registry.py

Saadmrp1038 (changes requested):

Ipetr14:
Thanks for the clarification on how `recovery_fault()` should work. I updated it so it leaves the rotated credentials in place and restarts the `product_catalog` pod so it would have rotated credentials as well. I ran problem validation, and now after fault recovery oracle passes. I also fixed the ordering of imports in `registry.py`.

Saadmrp1038 (approved):

Saadmrp1038:
@Ipetr14 Thanks for addressing my comments! Works correctly now.

I did a few runs and the results are interesting. My initial hypothesis from the RFC kind of holds:

> Even if the diagnosis is totally wrong, there's a high likelihood the agent would randomly mitigate the problem by restarting the faulty deployment.

In the third run, the diagnosis score was quite low and the mitigation attempt was also not quite accurate, but the restart _did_ fix the fault.

| Run | Diagnosis Score | Diagnosis Success | Mitigation Success | Diagnosis Summary | Mitigation Summary |
|-----|----------------|-------------------|-------------------|-------------------|-------------------|
| 1 | — | Timed out | — | Found auth failure and credential mismatch but got distracted investigating unrelated issues, never submitted | — |
| 2 | 89/100 | ✅ | ✅ | Correctly identified that the pod has stale credentials from before the Secret rotation, and env vars don't auto-refresh in running pods | Restarted product-catalog deployment |
| 3 | 34/100 | ❌ | ✅ | Submitted that PG password was never updated to match the Secret (opposite of what actually happened) | Set PG password to `otelp_7k9m2q4x` (already the current value) then restarted product-catalog. Passed because the restart picked up the correct Secret value, and the ALTER USER didn't change anything since PG already had that password. |

Congrats on the first PR merged here 🎉

PR #858, opened 2026-06-03
Fix/finalizer deadlock mitigation

ermias19:
<html><head></head><body><h1>[Problem] Add FinalizerDeadlock — Stuck Finalizer Deadlock for Hotel Reservation</h1>
<h2>Summary</h2>
<p>This PR introduces a new SREGym problem: <code>FinalizerDeadlock</code>, simulating a real-world Kubernetes failure pattern where a resource becomes permanently stuck in <code>Terminating</code> state due to an orphaned finalizer whose responsible controller no longer exists.</p>
<p>This is a high-value problem because the failure is <strong>non-obvious</strong> — the cluster and most pods appear healthy, but the affected namespace is silently degraded. Diagnosing it requires understanding Kubernetes object lifecycle and finalizer mechanics, going well beyond standard pod health inspection.</p>
<hr>
<h2>Motivation</h2>
<p>Kubernetes finalizers are widely used by operators and controllers to ensure cleanup logic runs before a resource is deleted. In production environments, this mechanism can silently break when:</p>
<ul>
<li>The controller responsible for removing the finalizer is unavailable or has been removed during an upgrade</li>
<li>The finalizer references a non-existent or misconfigured controller</li>
<li>The cleanup operation repeatedly fails with no retry limit</li>
</ul>
<p>When this happens, the resource enters a permanent <code>Terminating</code> state. The deletion timestamp is set, but Kubernetes will never complete the deletion. This is a real-world failure pattern observed across Kubernetes operators, CRDs, namespaces, storage resources, and custom controllers.</p>
<p>The resulting incident is particularly dangerous because:</p>
<ul>
<li>Standard health checks pass — pods in other namespaces appear fine</li>
<li>The affected namespace becomes degraded — new pods cannot be scheduled or restarted</li>
<li>Services depending on that namespace start failing — in this case, the Hotel Reservation booking service returns errors</li>
<li>Root cause investigation requires inspecting resource metadata directly, not just pod logs or status</li>
</ul>
<hr>
<h2>Problem Design</h2>

| Field | Value |
|---|---|
| **Problem key** | `finalizer_deadlock_hotel_reservation` |
| **Target application** | Hotel Reservation |
| **Faulty resource** | `ConfigMap/reservation-cleanup-token` |
| **Namespace** | `hotel-reservation` |
| **Finalizer** | `cleanup.reservations.io/pending-cleanup` |
| **Fault type** | Orphaned finalizer — no controller to remove it |
 
### Failure Chain
 
```
ConfigMap created with finalizer
        ↓
kubectl delete triggered (--wait=false)
        ↓
Deletion timestamp set by Kubernetes
        ↓
Kubernetes waits for controller to remove finalizer
        ↓
Controller does not exist → finalizer never removed
        ↓
ConfigMap stuck in Terminating indefinitely
        ↓
Namespace degraded → pod rescheduling blocked
        ↓
Booking service unavailable (pods cannot restart)
        ↓
Cluster appears partially healthy — root cause non-obvious
```
 
### Fault Injection (`inject_fault`)
 
1. Creates `ConfigMap/reservation-cleanup-token` with finalizer `cleanup.reservations.io/pending-cleanup`
2. Triggers deletion with `--wait=false` so Kubernetes sets the deletion timestamp but does not wait
3. The responsible controller does not exist — the finalizer is never removed
4. The ConfigMap remains permanently stuck in `Terminating`
### Fault Recovery (`recover_fault`)
 
1. Patches the ConfigMap to remove the finalizer using a JSON patch (`op: remove`)
2. Verifies the finalizer is gone by re-reading the resource
3. Deletes the ConfigMap with `--ignore-not-found` to clean up
4. Confirms the resource is fully gone and the application returns to healthy state
---
 
### Diagnosis Oracle — `LLMAsAJudgeOracle`
 
Evaluates whether the agent correctly identified the root cause across three dimensions:
 
| Dimension | What it checks |
|---|---|
| **D1 — Fault Localization** | Did the agent identify the correct resource (`ConfigMap/reservation-cleanup-token`)? |
| **D2 — Fault Characterization** | Did the agent identify the orphaned finalizer and absent controller as the cause? |
| **D3 — Scope Precision** | Did the agent avoid blaming unrelated components? |
 
### Mitigation Oracle — `FinalizerDeadlockMitigationOracle`
 
Verifies that the agent correctly resolved the incident:
 
- **Pass:** ConfigMap is fully deleted (404) AND application health check passes
- **Fail — still terminating with finalizer:** Deletion timestamp set, finalizer still present
- **Fail — still terminating:** Deletion timestamp set, finalizer removed but resource not gone
- **Fail — still exists:** Resource was never deleted at all
### Reward Hacking Resistance
 
The following actions are explicitly rejected by the oracle:
 
- Restarting unrelated pods
- Scaling deployments up or down
- Deleting unrelated resources
- Temporary workarounds that leave the finalizer in place
---
 
## Files Changed
 
| File | Change | Purpose |
|---|---|---|
| `sregym/conductor/problems/finalizer_deadlock.py` | New | Core problem — fault injection, recovery, and oracle wiring |
| `sregym/conductor/oracles/finalizer_deadlock_mitigation.py` | New | Custom mitigation oracle — checks finalizer removal and app health |
| `sregym/conductor/problems/registry.py` | Modified | Registers `finalizer_deadlock_hotel_reservation` |
 
---
 
### `FinalizerDeadlock.__init__`
 
- Uses a local `app` variable before `super().__init__()` to avoid redundant `self.app` assignment before the base class sets it
- `self.namespace` is not reassigned after `super().__init__()` since `Problem.__init__` already sets it
- `self.kubectl = KubeCtl()` is initialized separately as it is not part of the base `Problem` interface
### `_clear_finalizer_if_present`
 
- Reads the ConfigMap first to check if finalizers exist before patching — avoids unnecessary `kubectl patch` calls
- Uses JSON patch (`op: remove`) instead of merge patch for precise finalizer removal
- Verifies the finalizer is actually gone after patching — raises `RuntimeError` if it persists
- Handles `404` gracefully at both the read and post-patch verification steps
### `_wait_until_configmap_deleted`
 
- Polls with a 30-second timeout
- Calls `_clear_finalizer_if_present` on each iteration to unblock any stuck termination during setup or teardown
- Raises `TimeoutError` if the resource does not disappear within the deadline
### `_create_finalized_configmap`
 
- Split into `_delete_existing_configmap` and `_create_configmap_with_finalizer` for clarity
- Handles `409 Conflict` on creation by clearing any existing finalizer, waiting for full deletion, then retrying
---
 

##  Problem Validation — `finalizer_deadlock_hotel_reservation`
 
**Result:** **PASSED**
 
The problem completed the full lifecycle: the app deployed, the mitigation oracle detected the injected fault, and `recover_fault()` restored the app to a healthy state. Human review is still required.
 
| Stage | Status |
|-------|:------:|
| Resolve problem in registry | FinalizerDeadlock · app `Hotel Reservation` |
| Deploy application  | `Hotel Reservation` deployed to namespace `hotel-reservation` |
| Inject fault  | `inject_fault()` completed without error |
| Oracle fails after fault injection | Oracle reported failure after 1 check(s) |
| Recover fault  | `recover_fault()` completed without error |
| Oracle passes after recovery  | Oracle reported success after 2 check(s) |
 
_Lifecycle: deploy app → inject fault → oracle fails → recover fault → oracle passes._

</body></html>


HacksonClark:
Problem is too easy (agent already solves it).

PR #860, opened 2026-06-03
Fix orphaned child processes from shell-launched agent cleanup (#857)

### LINKED ISSUE #857: Agent timeout cleanup can leave orphaned child processes for shell-launched agents
SREGym launches non-container agents with subprocess.Popen(..., shell=True). The tracked Popen is the shell wrapper, which on timeout calls cleanup at https://github.com/SREGym/SREGym/blob/241d40a94493ff9b52b2f41524a0a26a0451126c/sregym/agent_launcher.py#L148 but it only kills the shell wrapper not the child agent process leaving orphan processes running. 

https://github.com/SREGym/SREGym/blob/241d40a94493ff9b52b2f41524a0a26a0451126c/sregym/agent_launcher.py#L17

- child processes can keep running in the background calling mcp 
-  an orphaned driver can submit an old diagnosis/mitigation while the current problem is active

Saadmrp1038:
Fixes #857

Non-container agents are launched with `subprocess.Popen(..., shell=True)`, so the tracked process is the shell wrapper. On timeout, `cleanup_agent` called `terminate()`/`kill()` on that wrapper PID only, leaving the real driver and its children orphaned.

Given fix:
- Launch the shell wrapper in its own session (`start_new_session=True`) so it leads a process group (`pgid == pid`), tracked on `AgentProcess`.
- Cleanup now signals the whole group: `SIGTERM` -> wait for the group to drain -> `SIGKILL` escalation, then reaps the wrapper.
- Branch cleanup on the per-process `container_name` instead of the global container runner, so a shell agent is never silently left running in a mixed run.
- Fall back to single-process termination when no pgid was captured.

### DIFF
--- a/sregym/agent_launcher.py
+++ b/sregym/agent_launcher.py
@@ -1,8 +1,11 @@
+import contextlib
 import logging
 import os
+import signal
 import subprocess
 import sys
 import threading
+import time
 from datetime import UTC, datetime
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
@hydrogenbond007 Can you please take a look at this and see if it works?

tianyin (approved):
The code looks good. Feel free to merge if @hydrogenbond007 confirms.

hydrogenbond007:
Hey appreciate the quick escalation! 

pr looks good to be merged @Saadmrp1038 @tianyin

tianyin:
Merged. @hydrogenbond007 Let us know if you still experience the same problem.

Issue #811, opened 2026-05-28
[Problem RFC] CoreDNS saturation via `ndots:5` DNS query amplification


Mr-NOBY:
Hello all, I am working on the evaluation task and would like to propose a new problem for the benchmark.

What came to my mind was the Kubernetes `ndots:5` DNS amplification issue. By default, every pod gets `ndots:5` in its `/etc/resolv.conf`, which means any domain with fewer than 5 dots (i.e., virtually every real domain like `api.stripe.com`) is treated as relative. The resolver then tries appending each search domain (`default.svc.cluster.local`, `svc.cluster.local`, `cluster.local`) before finally trying the actual name. A single external lookup can generate up to 10 DNS queries to CoreDNS. At scale, this silently saturates CoreDNS and causes intermittent latency spikes and timeouts across the entire cluster, while every pod still shows `Running`/`Ready`. This has bitten multiple companies in production, including Zalando (whose CoreDNS pods were OOMKilled by the DNS storm), Preply, and Airbnb.

References:
- [Kubernetes Failure Stories - Zalando - KubeCon Barcelona 2019](https://www.youtube.com/watch?v=6sDTB4eV4F8) Zalando's CoreDNS OOMKill incident caused by `ndots:5` query amplification
- [10 Ways to Shoot Yourself in the Foot with Kubernetes, #9 Will Surprise You - Datadog - KubeCon Barcelona 2019](https://www.youtube.com/watch?v=QKI-JRs2RIE) covers DNS `ndots:5` as a major production pitfall
- [Did Kubernetes Make My p95s Worse? - Airbnb - KubeCon NA 2019](https://www.youtube.com/watch?v=QXApVwRBeys) latency investigation involving CPU throttling and DNS issues ([k8s.af entry](https://k8s.af))
- [DNS issues in Kubernetes. Public postmortem #1 - Preply - 2020](https://medium.com/preply-engineering/dns-postmortem-e169efd45afd) public postmortem involving conntrack, DNS, and CoreDNS-autoscaler ([k8s.af entry](https://k8s.af))
- [Intermittent delays in Kubernetes - MindTickle - 2019](https://medium.com/techmindtickle/intermittent-delays-in-kubernetes-e9de8239e2fa) 5-second DNS lookup delays due to conntrack DNAT/SNAT race conditions ([k8s.af entry](https://k8s.af))

## How I plan to simulate the failure on SREGym
I am considering using the **astronomy-shop** application for this.
[...]

PR #865, opened 2026-06-05
add: coredns_saturation_ndots problem

Mr-NOBY:
### Problem description #811 

Simulates the real-world Kubernetes `ndots:5` DNS amplification issue. By default every pod gets `ndots:5` in its `/etc/resolv.conf`, meaning any domain with fewer than 5 dots (i.e., virtually every real domain like `api.stripe.com`) is treated as relative. The resolver appends each search domain (`default.svc.cluster.local`, `svc.cluster.local`, `cluster.local`) before trying the actual name, generating up to 10 DNS queries per lookup. At scale this silently saturates CoreDNS and causes intermittent latency spikes and timeouts across the entire cluster, while all pods remain Running/Ready.

This has bitten multiple companies in production, including Zalando (whose CoreDNS pods were OOMKilled by the DNS storm), Preply, and Airbnb.

References:
- [Kubernetes Failure Stories - Zalando - KubeCon Barcelona 2019](https://www.youtube.com/watch?v=6sDTB4eV4F8)
- [10 Ways to Shoot Yourself in the Foot with Kubernetes, #9 Will Surprise You - Datadog - KubeCon Barcelona 2019](https://www.youtube.com/watch?v=QKI-JRs2RIE)
- [Did Kubernetes Make My p95s Worse? - Airbnb - KubeCon NA 2019](https://www.youtube.com/watch?v=QXApVwRBeys)
- [DNS issues in Kubernetes. Public postmortem #1 - Preply - 2020](https://medium.com/preply-engineering/dns-postmortem-e169efd45afd)

### Fault injection mechanism

1. **Constrain CoreDNS**: Patch the `kube-system/coredns` deployment to 1 replica with resource limits (`cpu: 150m`, `memory: 128Mi`). Original config is saved as an annotation on the deployment itself for crash-safe recovery.
2. **Deploy DNS stress workload**: An `analytics-collector` deployment (2 replicas, 10 threads each) generates high-volume fire-and-forget raw UDP DNS queries while explicitly simulating the pod's configured ndots search-path expansion (ndots:5 by default), constructing all expanded queries per search domain, and sending A and AAAA packets independently, producing up to 8 queries per domain lookup and amplifying load on CoreDNS.
3. **Result**: CoreDNS becomes saturated, legitimate DNS queries time out, and the application experiences latency spikes or timeouts, while all pods remain in Running/Ready state or enter a CrashLoopBackOff state depending on the number of requests,.

### Recovery

- Delete the `analytics-collector` stress deployment
- Restore CoreDNS to pre-injection configuration from the saved annotation
- Rollout restart CoreDNS to apply restored limits

### Mitigation oracle

Custom `CoreDNSMitigationOracle` with tiered evaluation:

| Check | Type | Score |
|-------|------|-------|
| DNS health (in-pod timing, 10 samples, median < 50ms, max < 100ms) | **Gate** | Required |
| ndots ≤ 2 on analytics-collector or cluster-wide | **Full pass** | ✅ |
| Stress workload deleted/scaled to 0 + DNS healthy | **Partial pass** | ⚠️ |
| CoreDNS healthy (replicas ready, no CrashLoopBackOff) | Supporting | Informational |

The agent must look beyond `kubectl get pods` (all pods show Running/Ready) and trace the issue to DNS .

### Changes

- **New**: `sregym/conductor/problems/coredns_saturation_ndots.py` - Problem class
- **New**: `sregym/conductor/oracles/coredns_saturation_mitigation.py` - Mitigation oracle
- **Modified**: `sregym/conductor/conductor.py` - Added CoreDNS cleanup in `fix_kubernetes()`
- **Modified**: `sregym/conductor/problems/registry.py` -  Registered `coredns_saturation_ndots_hotel_reservation`

### Testing

- Validated fault injection: CoreDNS becomes resource-constrained and DNS queries time out
- Validated recovery: CoreDNS restored to original config from annotation
- Validated oracle: passes after ndots reduction, partial pass after stress removal, fails during active fault
- Validated conductor cleanup: `fix_kubernetes()` correctly restores CoreDNS on re-runs

/validate-problem coredns_saturation_ndots_hotel_reservation

### DIFF
--- a/sregym/conductor/problems/coredns_saturation_ndots.py
+++ b/sregym/conductor/problems/coredns_saturation_ndots.py
@@ -30,7 +30,7 @@ _STRESS_DEPLOYMENT = "analytics-collector"
 # Annotation key used to persist the original CoreDNS configuration on the
 # deployment itself.  This survives process crashes and avoids fragile temp
 # files or hardcoded defaults.
-_ORIGINAL_CONFIG_ANNOTATION = "sregym.io/coredns-original-config"
+_ORIGINAL_CONFIG_ANNOTATION = "original-config"
 
 
 class CoreDNSSaturationNdots(Problem):
@@ -279,6 +279,15 @@ def _restore_coredns_from_annotation(apps_v1, kubectl):
         return
 
     original = json.loads(raw)
+
+    resources = original.get("resources", {})
+    if "limits" not in resources:
+        resources["limits"] = {}
+
+    # Set to None if 'cpu' wasn't part of the original limits config
+    if "cpu" not in resources["limits"]:
+        resources["limits"]["cpu"] = None
+
     patch_body = {
         "metadata": {"annotations": {_ORIGINAL_CONFIG_ANNOTATION: None}},
         "spec": {
@@ -288,7 +297,7 @@ def _restore_coredns_from_annotation(apps_v1, kubectl):
                     "containers": [
                         {
                             "name": _COREDNS_DEPLOYMENT,
-                            "resources": original.get("resources", {}),
+                            "resources": resources,
                         }
                     ]
                 }
--- a/sregym/conductor/problems/coredns_saturation_ndots.py
+++ b/sregym/conductor/problems/coredns_saturation_ndots.py
@@ -1,0 +1,306 @@
+"""CoreDNS saturation caused by ndots:5 DNS query amplification.
+
+A stress workload (analytics-collector) generates high-volume DNS lookups whose
+names contain fewer dots than the Kubernetes default ndots:5.  This triggers
+search-path expansion, multiplying every query into several suffixed lookups and
+overwhelming the resource-constrained, single-replica CoreDNS pod.  The result
+is cluster-wide DNS latency and timeouts.
+
+Canonical fix: set ``dnsConfig.options: [{name: ndots, value: "2"}]`` on the
+offending deployment so search-path amplification stops.
+"""
+
+import json
+
+import yaml
+from kubernetes import client
+from kubernetes.client.rest import ApiException
+
+from sregym.conductor.oracles.coredns_saturation_mitigation import CoreDNSMitigationOracle
+from sregym.conductor.oracles.llm_as_a_judge.llm_as_a_judge_oracle import LLMAsAJudgeOracle
+from sregym.conductor.problems.base import Problem
+from sregym.service.apps.hotel_reservation import HotelReservation
+from sregym.service.kubectl import KubeCtl
+from sregym.utils.decorators import mark_fault_injected
+
+_COREDNS_DEPLOYMENT = "coredns"
+_COREDNS_NAMESPACE = "kube-system"
+_STRESS_DEPLOYMENT = "analytics-collector"
+
+# Annotation key used to persist the original CoreDNS configuration on the
+# deployment itself.  This survives process crashes and avoids fragile temp
+# files or hardcoded defaults.
+_ORIGINAL_CONFIG_ANNOTATION = "sregym.io/coredns-original-config"
+
+
+class CoreDNSSaturationNdots(Problem):
+    """Inject CoreDNS saturation via ndots:5 query amplification."""
+
+    def __init__(self):
+        self.app = HotelReservation()
+        super().__init__(app=self.app, namespace=self.app.namespace)
+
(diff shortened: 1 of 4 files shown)


sregym/conductor/problems/coredns_saturation_ndots.py
```diff
@@ -0,0 +1,306 @@
...
+from sregym.utils.decorators import mark_fault_injected
+
+_COREDNS_DEPLOYMENT = "coredns"
+_COREDNS_NAMESPACE = "kube-system"
+_STRESS_DEPLOYMENT = "analytics-collector"
+
+# Annotation key used to persist the original CoreDNS configuration on the
+# deployment itself.  This survives process crashes and avoids fragile temp
+# files or hardcoded defaults.
+_ORIGINAL_CONFIG_ANNOTATION = "sregym.io/coredns-original-config"
```
Saadmrp1038: nit: Let's use a realistic/generic name here

sregym/conductor/problems/coredns_saturation_ndots.py
```diff
@@ -0,0 +1,306 @@
...
+            self.apps_v1.delete_namespaced_deployment(_STRESS_DEPLOYMENT, self.namespace, grace_period_seconds=0)
+            print("Deleted DNS stress deployment.")
+        except ApiException as e:
+            if e.status == 404:
+                print("DNS stress deployment already absent.")
+            else:
+                print(f"Error deleting stress deployment: {e}")
+
+        # 2. Restore CoreDNS to its pre-injection state using the saved annotation.
+        _restore_coredns_from_annotation(self.apps_v1, self.kubectl)
```
Saadmrp1038, line 218: I don't think this recover mechanism is undoing the CPU limit. Kubernetes strategic merge patch doesn't remove keys that are absent in the patch body. The restore needs to explicitly null the CPU limit ("cpu": None) or use a JSON merge patch.

Saadmrp1038 (commented):
@Mr-NOBY I tested this end-to-end and found fundamental issues with the fault design.

This isn't really a DNS configuration fault rather a DDoS problem. The ndots amplification logic in the Python code doesn't matter because the base flood rate alone (without amplification) still OOMKills CoreDNS. Setting `ndots:2` on `analytics-collector` does reduce the query count (the Python code reads resolv.conf and skips amplification), but the base flood rate without amplification still OOMKills CoreDNS. The oracle's "full pass" path is unreachable. Scaling CoreDNS restores DNS fully but the oracle rejects it. The only passing mitigation is deleting the stress pod. The amplification itself is also faked. In the real world the resolver handles ndots expansion transparently (blocking on each query), but here the code manually constructs the expanded queries and blasts them all via raw UDP without waiting for responses.

> Validated oracle: passes after ndots reduction

In my case it didn't pass. Around 500+ queries per second even after `ndots:2` fix OOMkilled the CoreDNS. I am not sure why it would work for you.

I ran this a couple of times with an agent. The RFC envisions this careful path: notice latency, trace to DNS, check CoreDNS, examine `resolv.conf`, understand `ndots:5` amplification. The agent just did `kubectl get all`, noticed `analytics-collector`, ran `kubectl describe pod`, read the Python DDoS code in the pod spec, called it a "DNS flood attack," and deleted it. It's reasoning was correct but still got low score due to not mentioning ndots. There was no reason to dig deeper when the source code is right there in the spec. 

`ndots:5` isn't really a bug itself. It's a deliberate Kubernetes default for service discovery convenience. It only becomes a problem at scale when many services make external calls and the amplification silently accumulates across the cluster. But here there's nothing subtle to discover. No agent would reason about `resolv.conf` ndots settings when the pod spec shows the query calls in a tight loop.

Mr-NOBY:
@Saadmrp1038 Thanks for pointing that out. I will take a closer look upon these issues.

>  In my case it didn't pass. Around 500+ queries per second even after `ndots:2` fix OOMkilled the CoreDNS. I am not sure why it would work for you.

I am going to investigate this more as for me setting `ndots:2` returned the latencies back to normal (the same latencies with `analytics-collector` scaled to 0) which passed the oracle.

I will readjust the approach by abandoning the explicit tight loop causing the "DNS flood attack" and simulate the load in a different proper way that triggers the amplification without DDoS-ing the cluster. I'll give you an update on the PR once I finish that.

tianyin:

Issue #817, opened 2026-05-29
[Problem RFC] Co-located backup scan evicts OS page cache and slows hotel search


alpblkba:
# [Problem RFC] Co-located backup scan evicts OS page cache and slows hotel search

## Motivation

A latency-sensitive service can look healthy in Kubernetes while becoming slow because another pod on the same node evicted its hot file-backed working set from the OS page cache.
Elasticsearch/Lucene workloads rely heavily on the OS page cache. Elastic documents that when cached data is evicted sooner, OS page faults increase and query latency goes up. A real Elastic operator report says page-cache eviction was one of their biggest Elasticsearch performance problems, commonly triggered by logging, backups, and HDFS tasks.

Kubernetes has the same abstraction leak. Kubernetes issue #125720, “Pod Memory Request=Limit Triggers Pagecache Reclamation and Service Jitter,” gives a minimal repro where scheduling a new pod triggers pagecache reclamation and service jitter in existing Kafka/Elasticsearch-like pods, even when the node is not apparently overcommitted. The important point is simple: Kubernetes can say the pod is fine, while the kernel has silently made the service slow.

## Proposal

In `hotel_reservation`, the hotel search path uses a local file-backed ranking/search artifact, similar to a small Lucene-style index. When the artifact is warm in memory, search is fast. The injected fault schedules a backup scanner or index verifier on the same node. The scanner reads enough data to evict the hot search pages from the OS page cache.

For implementation, this can be done either by adding a lightweight file-backed lookup to the hotel search path, or by placing a small search-index gateway in front of the existing search service. The gateway keeps the application semantics unchanged, but gives the benchmark a deterministic file-backed working set to warm and evict.

The visible symptom is slow hotel search, higher p95/p99 latency, or occasional timeouts, while the pod may still be Running and Ready with clean logs, normal-looking CPU, and healthy-looking database/network signals. This is the intended trap for the agent: the symptom looks like database slowness, frontend regression, autoscaling failure, or CPU underprovisioning, but the root cause is node-local cache pressure caused by harmful co-location.

## Target app

Initial target: `hotel_reservation`, because travel systems inherently have search, ranking, availability, and pricing paths. The same idea could later fit `train_ticket` route/seat search, `astronomy_shop` catalog search, or `social_network` feed ranking, but Hotel Reservation is the cleanest first implementation.

## Repro and diagnosis

[...]

PR #866, opened 2026-06-05
Add hotel reservation page-cache thrash problem

alpblkba:
## Summary

I started from the page-cache noisy-neighbor idea discussed in #817 and implemented the smallest version that I could validate end-to-end.

The motivation is a failure mode where Kubernetes objects look mostly healthy, but a latency-sensitive service is colocated with a batch workload that is aggressively scanning a file-backed working set. The root cause is not a bad image, missing service, DNS issue, or CrashLoop. It is bad workload placement and node-local cache pressure.

In this version, I focused on the frontend path of Hotel Reservation. The fault places a synthetic `cache-thrasher` workload on the same node as the `frontend` pod. The thrasher uses a bounded `emptyDir` working set and repeatedly scans it, so the problem stays contained and does not consume host disk unboundedly.

## What I implemented

The injected fault currently does this:

1. Finds the node where the Hotel Reservation `frontend` pod is running.
2. Creates a `cache-thrasher` deployment on that same node.
3. Uses a bounded file-backed working set under an `emptyDir`.
4. Keeps the thrasher pod Running, so the obvious Kubernetes health checks are not enough.
5. Recovers by deleting the synthetic thrasher deployment.

The mitigation side checks for the condition that should actually matter: the cache-thrashing workload should no longer be running on the frontend node, and the frontend should still respond after mitigation.

That gives us a first validated version of the noisy-neighbor scenario without making the oracle depend on an unstable latency metric yet.

## Validation

I validated the full lifecycle locally on an ARM kind cluster:

```text
  --problem page_cache_thrash_hotel_reservation \
  --inject-timeout 300 \
  --recover-timeout 600 \
  --poll-interval 15
```

The validation passed:

```text
Resolve problem in registry: passed
Deploy application: passed
Inject fault: passed
Oracle fails after fault injection: passed
Recover fault: passed
Oracle passes after recovery: passed
```

Requested CI validation:

```text
/validate-problem page_cache_thrash_hotel_reservation
```

## Notes from trying the latency path

I also tried to connect this to the existing Prometheus p95/p99 path, but I did not include that in the oracle yet.

For regular `HotelReservation`, my manual Prometheus checks did not produce usable `traces_span_metrics_*` latency samples for `hotel-reservation`. I watched the problem from a second terminal while the validation/manual run was active and queried Prometheus directly, including the existing p95-style histogram query and active alerts.

For example:

```text
kubectl exec -n observe deploy/prometheus-server -c prometheus-server -- \
  wget -qO- 'http://localhost:9090/api/v1/query?query=histogram_quantile(0.95,sum%20by%20(le,service_name,namespace)%20(rate(traces_span_metrics_duration_milliseconds_bucket%7Bnamespace%3D%22hotel-reservation%22%7D%5B2m%5D)))'
```

and:

```text
kubectl exec -n observe deploy/prometheus-server -c prometheus-server -- \
  wget -qO- "http://localhost:9090/api/v1/alerts"
```

I did see spanmetrics for `blueprint-hotel-reservation`, so Blueprint may be the better path if we want a Prometheus-backed latency oracle. However, on my ARM kind setup, the Blueprint app did not become healthy as a baseline before any injected fault. Several service deployments entered `CrashLoopBackOff` or `Completed`. Because of that, I kept this first version on regular `HotelReservation`, which validated cleanly.

## Follow-up directions

I think there are a few natural ways to harden or extend this:

- add a Prometheus p95/p99 oracle once the telemetry path is stable for the target app;
- add a workload-log-based latency oracle if that is more reliable than Prometheus in local validation;
- build a BlueprintHotelReservation version if the Blueprint baseline is expected to work in CI;
- split the RFC idea into separate noisy-neighbor problems:
  - frontend colocated with cache thrasher;
  - search/backend path colocated with cache thrasher, for example `search`, `geo`, or `rate`;
  - database/cache-heavy pod colocated with cache thrasher.

For now, I kept the PR narrow: one reproducible placement fault, one bounded synthetic workload, one lifecycle-validating oracle.

## Files changed

- `sregym/conductor/problems/page_cache_thrash_hotel_reservation.py`
- `sregym/conductor/oracles/page_cache_thrash_mitigation_oracle.py`


Saadmrp1038 (commented):
@alpblkba Thanks for the effort!

I tested this on a CloudLab cluster (40 CPU cores, 196 GB RAM nodes) and saw no observable degradation. In fact unless I missed something there's no notable latency increase anywhere.

Without any of kind symptoms or degradation there's nothing for the agent to diagnose and fix.

alpblkba:
Thanks for taking a look @Saadmrp1038 

I agree that the current version needs a clearer observable degradation signal.  I initially tried to connect this to p95/p99 latency, but the regular HotelReservation path did not expose a straightforward Prometheus metric in my local setup. My current thought is to avoid depending on Prometheus and Blueprint for now and instead add a small workload-side latency probe against the HotelReservation user path: measure a baseline before injection, inject the co-located cache-pressure pod, and check whether p95 latency increases relative to the baseline while the pods remain Running/Ready.

As a secondary validation signal, we could alos look at node-level memory/cache pressure indicators such as pgmajfault, workingset_refault, or related /proc/vmstat counters, but I would probably keep the user-path latency as the main oracle.

Would this be a reasonable direction for this fault, or would you prefer another observable signal for noisy-neighbor/cache-pressure scenarios?
If you have a preferred observable signal for this kind of fault I’d be happy to try that direction.

Also, if this PR is not the best fit for the benchmark as-is, I can look for another issue/fault candidate and come back with a cleaner follow-up PR.

Saadmrp1038:
@alpblkba I don't have any preferences on the oracle signal. Although I have concerns that this might boil down to a simple noisy-neighbor problem where the agent sees a pod consuming ~1 CPU core and 800Mi (from my testing with `kubectl top`), deletes it, and moves on without ever reasoning about page-cache eviction or the underlying mechanism. What should the ideal mitigation be?

---

Also, the deployment name `cache-thrasher` already exposes the fault to the agent. We should use a more generic name here.

alpblkba:
That concern makes sense. I think the current version can still collapse into “find the expensive pod and delete it”, which is not the behavior we want.

I am going to rework the direction a bit before pushing again: rename cache-thrasher, make the background workload look like a required service, fail delete/restart/reboot-style mitigations, and add more logging so we can see what diagnostics the agent actually runs.
one thing I’m still unsure about is the signal for the agent because without Prometheus or p95/p99 labels (or any other observability solution), what would be the preferred way for the agent to observe that the frontend is degraded? A simple curl-based timing check could work, but I’m not sure if that is the right SREGym signal.

secondly, do you think seeing the worker as resource-heavy in kubectl top is already a poor signal? or is it acceptable as an initial suspicion signal, as long as the worker is necessary and cannot be deleted/restarted, forcing the agent to continue into lower-level checks like /proc/vmstat, PSI, cgroup io.stat, or iostat before deciding on a mitigation?
alternatively, if we make CPU/memory less suspicious and move most of the signal into kernel-filesystem level behavior, do you think the agent would still reliably notice that something is wrong? it would make the task more faithful to the page-cache mechanism, but we may need some user-visible degradation signal so the agent has a reason to investigate.

The mitigation I’m leaning toward is "changing/moving the contention surface", not removing/rebooting the workload. for example, moving the worker’s artifact path away from the frontend-sensitive path, or changing the worker mode so it remains active but no longer creates the same filesystem/page-cache interference. or rate-limiting the contention causing resource.

alpblkba:
quick update about the rework I am planning @Saadmrp1038 

The main change is to make the background workload a required service, not something the agent can simply delete or restart. I’ll remove the page-cache-specific naming, add a worker heartbeat/report artifact, and reject shortcut operations such as deleting, restarting, or disabling the worker.

Then I'll check the reworked version locally with both SREGym cli.py command tracing and terminal-bench/tmux output. My goal now is to inspect what the agent actually does whether it only runs basic Kubernetes checks, or whether it reaches to relevant (and certain) filesystem/kernel/cgroup-level diagnostics before applying a mitigation. agent's deep reasoning and diagnostic path clearer by showing that the issue is not network, throttling, database, or functional correctness.

(terminal-bench would only be used for local agent-behavior analysis, and not added as a dependency to the PR.)

I’ll update the PR once the rework is cleaner and I have local agent behavior to report.

Saadmrp1038:
@alpblkba The plan sounds great!

> My goal now is to inspect what the agent actually does whether it only runs basic Kubernetes checks, or whether it reaches to relevant

This is a good approach. It is often hard to say what an agent will do without actually running one.

tianyin:

PR #869, opened 2026-06-05
(fix): modify ready check to account for Suceeded

0xprames:
we hit the 600s timeout without this change - the runs seem to come with a set of specific assumptions about the cluster

im using `k3s` right now - my kubeconfig happens to be a `k3s` config i.e NOT `kubeadm`  etc, but the prometheus pods failed to schedule bc `sregym/observer/prometheus/prometheus/values.yaml` assumes:
```
"node-role.kubernetes.io/control-plane": ""
```

`k3s` happens to label 
```
node-role.kubernetes.io/control-plane=true
``` 

i am moving to use my DO or EKS clusters - but, if all my nodes have any value set there, the cluster will auto fail due to scheduler stuck trying to find a valid node

### DIFF
--- a/sregym/service/kubectl.py
+++ b/sregym/service/kubectl.py
@@ -137,6 +137,19 @@ class KubeCtl:
         """Fetch the service configuration."""
         return client.CoreV1Api().read_namespaced_service(name=name, namespace=namespace)
 
+    @staticmethod
+    def _is_completed_job_pod(pod) -> bool:
+        """True if the pod is a successfully-completed Job pod.
+
+        Such pods (e.g. k3s's helm-install-* pods in kube-system) sit in phase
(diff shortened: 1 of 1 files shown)


Saadmrp1038:
@0xprames I see. We didn't really test on k3s. Our suggested local setup is kind.

> but the prometheus pods failed to schedule bc sregym/observer/prometheus/prometheus/values.yaml assumes

I will take a look into it.

Saadmrp1038 (approved):

Saadmrp1038:
The changes LGTM

0xprames:
right but even if i use DOK8s for example

we can't run OS level faults bc no ssh access (or even ansible, inventory access) for those managed k8s workers

PR #871, opened 2026-06-05
Kill kubectl process group on timeout to prevent orphaned commands

Saadmrp1038:
## Problem
- `subprocess.run(shell=True, timeout=...)` only kills the `/bin/sh` wrapper on timeout, leaving the real `kubectl` (and piped/background children) running as orphans.
- Reproduced with `kubectl get pods --watch`: 2 live orphans left behind after the call "timed out."

## Fix
- Run commands in a new session (`start_new_session=True`) and SIGKILL the whole process group on timeout, so the entire command tree is cleaned up.
- Bound the post-kill pipe drain so a process that escaped the group can't re-introduce a hang during cleanup.
- Applied to both `exec_command` and `dry_run_json_output`; preserves existing `TimeoutExpired` / `CalledProcessError` semantics so callers are unchanged.

## Testing
- `kubectl get pods --watch`: 2 orphans pre-fix, 0 after. Verified locally and inside the live MCP pod.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,10 @@ uv run prek install
 Choose either a) or b) to set up your cluster and then proceed to the next steps.
 
 ### a) Kubernetes Cluster (Recommended)
-SREGym supports any kubernetes cluster that your `kubectl` context is set to, whether it's a cluster from a cloud provider or one you build yourself.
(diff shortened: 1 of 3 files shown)


Saadmrp1038:
@HacksonClark Can you check the readme changes please?

tianyin (approved):
The README looks good to me.

Issue #870, opened 2026-06-05
Fix OpenEBS setup idempotency for current operator manifest


hydrogenbond007:
The OpenEBS operator manifest already creates openebs-device. SREGym then re-applies a different StorageClass definition, causing Kubernetes to reject the deploy because StorageClass parameters are immutable. 

Lead to a lot of setup fails since pvc forbade updates

https://github.com/SREGym/SREGym/blob/main/sregym/conductor/conductor.py#L653

tianyin:
Fixed by #872 

Thanks again @hydrogenbond007 !

PR #872, opened 2026-06-05
Fix OpenEBS setup idempotency

hydrogenbond007:
The current setup applies the upstream OpenEBS operator manifest, which already creates `StorageClass/openebs-device`. SREGym then applies a different inline `openebs-device` definition. Kubernetes rejects that because `StorageClass.parameters` are immutable after creation, causing benchmark runs to fail during deploy/setup before the agent can diagnose the injected fault.

https://github.com/SREGym/SREGym/issues/870 

- Checks if Sotrageclass already exists before creating it
- Verify the operator-provided version matches the setup

### DIFF
--- a/sregym/conductor/conductor.py
+++ b/sregym/conductor/conductor.py
@@ -1,6 +1,7 @@
 import asyncio
 import concurrent.futures
 import logging
+import shlex
 import shutil
 import time
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


tianyin:
Thank you for the contribution! @hydrogenbond007 

Does this only affect khaos related problem? 

I'm confused why we didn't catch it earlier.

hydrogenbond007:
Not limited to khaos, any deployment that hits deploy_app() also hits OpenEbs path>

Likely you guys slipped since you are pulling the upstream yaml 

Lmk if this looks good!

tianyin:
> Likely you guys slipped since you are pulling the upstream yaml

so the upstream yaml changed in this case, right?

The PR looks great to me! 

I let @Saadmrp1038 merge it.

Saadmrp1038:
I know why this slipped. `kubectl.exec_command` was silently swallowing the stderr. #780 added a `logger.error(...)` to that except block last week, which surfaced it. Interestingly, we never got benchmark run failures for this. It was a no-op.

Saadmrp1038:
LGTM

Saadmrp1038 (approved):

PR #873, opened 2026-06-05
Finalizer deadlock controller clean

ermias19:
# [Problem] Add FinalizerDeadlock  Stuck Finalizer Deadlock for Hotel Reservation
 
## Summary
 
This replaces the earlier finalizer PR (`#858`) which got closed because agents could already solve it.
 
The old version was simple ConfigMap gets stuck, agent removes the finalizer, done. That worked as a Kubernetes exercise but not as a benchmark because the fix was too obvious.
 
This version keeps the same failure class but makes the root cause harder to reach. There is now a real controller running that owns the finalizer. The controller is broken  its ClusterRole lost `patch` and `update` on `configmaps/finalizers` so it keeps hitting 403 and logging the error, but it cannot finish cleanup. The ConfigMap stays stuck.
 
The agent cannot just remove the finalizer directly. The oracle rejects that path. The agent has to find the controller logs, trace the 403 back to the ClusterRole, fix the missing rules, and let the controller do the cleanup itself.
 
---
 
## What the failure looks like
 
```
ConfigMap/reservation-cleanup-token → stuck in Terminating
    → cleanup-controller is running but getting 403
    → ClusterRole is missing patch/update on configmaps/finalizers
    → controller cannot finish reconcile
    → ConfigMap never deleted
```
 
The cluster looks mostly healthy. The controller pod is Running. The only signal is in the controller logs — repeated 403 Forbidden on `configmaps/finalizers`. The agent has to find that.
 
---
 
## Problem design
 
| Field | Value |
|---|---|
| Problem key | `finalizer_deadlock_controller_hotel_reservation` |
| Target app | Hotel Reservation |
| Stuck resource | `ConfigMap/reservation-cleanup-token` |
| Namespace | `hotel-reservation` |
| Controller | `Deployment/cleanup-controller` |
| ServiceAccount | `cleanup-controller` |
| ClusterRole | `configmap-cleanup-controller` |
| Finalizer | `cleanup.reservations.io/pending-cleanup` |
 
---
 
## What inject_fault does
 
1. Creates `ServiceAccount/cleanup-controller`
2. Creates `ClusterRole/configmap-cleanup-controller` with full patch/update rules
3. Creates `ClusterRoleBinding` binding the SA to the role
4. Creates `ConfigMap/reservation-cleanup-token` with the finalizer
5. Starts `Deployment/cleanup-controller` — at this point it is healthy and could reconcile
6. Triggers `kubectl delete configmap ... --wait=false` — ConfigMap enters Terminating
7. Waits 3 seconds so the controller sees the deletionTimestamp
8. Strips patch/update from the ClusterRole — controller now gets 403 on every retry
The 3-second gap matters. The controller needs to register the deletion before RBAC is broken, otherwise it never starts trying and the logs are empty.
 
---
 
## What the correct agent path looks like
 
1. See `ConfigMap/reservation-cleanup-token` stuck in Terminating
2. Check metadata — see `cleanup.reservations.io/pending-cleanup` finalizer
3. Find `Deployment/cleanup-controller`
4. Read controller logs — see repeated `FORBIDDEN: ... configmaps/finalizers`
5. Inspect the ClusterRoleBinding → find `configmap-cleanup-controller`
6. Inspect the ClusterRole → see only `get/list/watch`, no `patch/update`
7. Restore the missing rules
8. Controller reconciles on next cycle and removes the finalizer
9. ConfigMap deletion completes automatically
---
 
## Oracle
 
`FinalizerDeadlockControllerMitigationOracle` checks all of these:
 
- `ClusterRole/configmap-cleanup-controller` exists and has `patch` + `update` on `configmaps` and `configmaps/finalizers`
- `ConfigMap/reservation-cleanup-token` is fully deleted (404)
- `Deployment/cleanup-controller` still exists and is healthy
- All Hotel Reservation application pods are healthy
**Shortcut rejection** — if the ConfigMap is gone but the ClusterRole is still broken, the oracle detects this and fails the run. That means the agent patched the finalizer directly instead of fixing the root cause. The oracle distinguishes the two paths by checking ClusterRole state against ConfigMap state together.
 
These are also rejected:
- Deleting the controller Deployment
- Deleting the ServiceAccount or RBAC objects instead of repairing them
- Restarting unrelated pods
---
 
## MCP RBAC change
 
The agent needs to be able to patch the ClusterRole to fix it. Added to `mcp_server/k8s/clusterrole.yaml`:
 
```yaml
- apiGroups: ["rbac.authorization.k8s.io"]
  resources:
    - clusterroles
  verbs: ["get", "list", "watch", "patch", "update"]
```
 
Bindings stay read-only — the agent can read them to trace the ownership chain but cannot modify them:
 
```yaml
- apiGroups: ["rbac.authorization.k8s.io"]
  resources:
    - clusterrolebindings
    - roles
    - rolebindings
  verbs: ["get", "list", "watch"]
```
 
This follows the same pattern as the admission webhook PR.
 
---
 
## Files changed
 
| File | Purpose |
|---|---|
| `sregym/conductor/problems/finalizer_deadlock_controller.py` | New problem |
| `sregym/conductor/oracles/finalizer_deadlock_controller_mitigation.py` | New oracle |
| `sregym/conductor/problems/registry.py` | Registers the problem |
| `mcp_server/k8s/clusterrole.yaml` | Agent RBAC expansion |
| `cleanup_controller_loop.py` | Reconcile loop script running inside the controller pod |
| `Dockerfile.cleanup_controller` | Builds the controller image |
 
---
 
## Validation
 
```bash
  --problem finalizer_deadlock_controller_hotel_reservation \
  --inject-timeout 180 \
  --recover-timeout 180
```
 
```
Resolve problem in registry:     PASSED
Deploy application:              PASSED
Inject fault:                    PASSED
Oracle fails after injection:    PASSED
Recover fault:                   PASSED
Oracle passes after recovery:    PASSED
```
 
---
 
## Real-world reference
 
This failure class is documented in kubernetes/kubernetes issue #60807 — resources stuck in Terminating because a controller lost its ability to remove finalizers. That thread has hundreds of production reports going back to 2018 across Rook, Fission, Rancher, and others.
 
---
 
## Note on agent trials
 
I will run multiple agent trials before treating this as final. The goal is to see the agent solve it inconsistently — not 0% and not 100%. Will drop the traces in a follow-up comment once I have them.


sregym/conductor/problems/registry.py

sregym/conductor/problems/registry.py
```diff
@@ -13,17 +13,19 @@
...
 )
+from sregym.conductor.problems.dev_shm_exhaustion_hotel_reservation import DevShmExhaustionHotelReservation
 from sregym.conductor.problems.duplicate_pvc_mounts import DuplicatePVCMounts
 from sregym.conductor.problems.edge_request_filter_cpu_saturation import EdgeRequestFilterCPUSaturation
 from sregym.conductor.problems.env_variable_shadowing import EnvVariableShadowing
 from sregym.conductor.problems.ephemeral_port_range_hotel_reservation import EphemeralPortRangeHotelReservation
 from sregym.conductor.problems.expired_tls_hotel_reservation import ExpiredTlsHotelReservation
 from sregym.conductor.problems.failed_readiness_probe import FailedReadinessProbe
 from sregym.conductor.problems.faulty_image_correlated import FaultyImageCorrelated
+from sregym.conductor.problems.finalizer_deadlock import FinalizerDeadlock
```
Saadmrp1038: I am guessing this is from your previous try. This import does not exist.

sregym/conductor/problems/registry.py
```diff
@@ -129,6 +131,8 @@ def __init__(self):
             "assign_to_non_existent_node": AssignNonExistentNode,
             "auth_miss_mongodb": MongoDBAuthMissing,
             "configmap_drift_hotel_reservation": lambda: ConfigMapDrift(faulty_service="geo"),
+            "finalizer_deadlock_hotel_reservation": FinalizerDeadlock,
```
Saadmrp1038: I am guessing this is from your previous try. This  fault does not exist.

sregym/conductor/oracles/finalizer_deadlock_controller_mitigation.py
```diff
@@ -0,0 +1,200 @@
...
+
+from kubernetes import client
+from kubernetes.client.rest import ApiException
+
+from sregym.conductor.oracles.base import Oracle
+
+_REQUIRED_VERBS = {"patch", "update"}
+_ROLLOUT_SETTLE_SECONDS = 60
+_ROLLOUT_POLL_INTERVAL = 5
+_CONTROLLER_COMPLETION_ANNOTATION = "sregym.io/finalizer-cleanup-completed"
```
Saadmrp1038: nit: We should use more realistic/generic name here and not expose the `sregym` name.

sregym/conductor/problems/finalizer_deadlock_controller.py
```diff
@@ -0,0 +1,571 @@
...
+
+_CONTROLLER_NAME = "cleanup-controller"
+_CONTROLLER_SCRIPT_CONFIGMAP = "cleanup-controller-script"
+_SA_NAME = "cleanup-controller"
+_CLUSTERROLE_NAME = "configmap-cleanup-controller"
+_CLUSTERROLEBINDING_NAME = "configmap-cleanup-controller"
+_CONFIGMAP_NAME = "reservation-cleanup-token"
+_FINALIZER = "cleanup.reservations.io/pending-cleanup"
+_CONTROLLER_IMAGE = "mongo:4.4.6"
+_CONTROLLER_COMPLETION_ANNOTATION = "sregym.io/finalizer-cleanup-completed"
```
Saadmrp1038: nit: We should use more realistic/generic name here and not expose the `sregym` name.

sregym/conductor/problems/finalizer_deadlock_controller.py
```diff
@@ -0,0 +1,571 @@
...
+    def _create_controller_script(self):
+        script = textwrap.dedent(
+            f"""\
+            #!/bin/sh
+            set -eu
+
+            echo "cleanup-controller starting namespace={self.namespace} configmap={self.configmap_name} finalizer={self.finalizer}"
+            while true; do
+              echo "ConfigMap {self.configmap_name} is Terminating with finalizer {self.finalizer}; attempting controller-owned cleanup"
+              echo "FORBIDDEN: ServiceAccount lacks permission to patch configmaps/finalizers in namespace {self.namespace}. Hint: check ClusterRole {self.clusterrole_name}; it may be missing patch/update rules for configmaps and configmaps/finalizers. Error: HTTP 403 Forbidden"
```
Saadmrp1038: Is this what an actual Kubernetes 403 message looks like? This seems to expose a lot of data.

Saadmrp1038 (changes requested):
@ermias19 Did a light review and dropped some nit comments.

> Will drop the traces in a follow-up comment once I have them.

You mentioned a follow-up after doing agent runs. Please ping me when it's done.

sregym/conductor/problems/finalizer_deadlock_controller.py
ermias19: yes that's actually correct. came up with `cleanup-controller.platform.io` its more generic and hides any sregym related calling.

sregym/conductor/problems/finalizer_deadlock_controller.py
ermias19: yes it is @Saadmrp1038 , shall I provide only generic and higher level information, by removing infos right like 

-ServiceAccount name
-ClusterRole name
-missing permission hints
-ServiceAccount name

WDYT ??

Saadmrp1038: @ermias19 I think we should try to follow whatever the norm is for these things.

sregym/conductor/problems/registry.py
ermias19: @Saadmrp1038, I have set up Ruff, used it to format the code and enforce the linter rules, and committed the affected modules.

ermias19:
@Saadmrp1038,  here are the the agent run reports. i have attached these two report.
[validation-summary.md](https://github.com/user-attachments/files/28829210/validation-summary.md)
[stratus_ALL_results.csv](https://github.com/user-attachments/files/28829214/stratus_ALL_results.csv)

sregym/service/container_runner.py
Saadmrp1038, line 1: This seems like a logic change(?). Is there a reason for this?

sregym/conductor/problems/finalizer_deadlock_controller.py
Saadmrp1038, line 1: nit: We should use a realistic/generic label instead of `sregym.io/injected`

sregym/conductor/oracles/finalizer_deadlock_controller_mitigation.py
Saadmrp1038, line 1: If an agent patches the ClusterRole with `verbs: ["*"]`, the oracle rejects it because it only checks  for `patch` and `update`.

Saadmrp1038 (changes requested):
@ermias19 The fault works great! Left some minor comments. I will merge the PR when they are addressed.

It seems like this is quite a difficult fault for the agents, as from my testing too, the agent kept timing out.

---

On a side note, I appreciate the repo-wide `ruff check .`, but this might have been more suited to a separate PR. Makes reviewing a bit harder. :)

sregym/service/container_runner.py
ermias19: yes @Saadmrp1038 ,  This fix is done, cause of `Ruff rule SIM114`. the SIM114 rule  Combines `if` branches using logical `or` operator. 

under the statement the value, which is :- 

```
env_vars["API_HOSTNAME"] = "host.docker.internal"
            mcp_port = env_vars.get("MCP_SERVER_PORT", os.environ.get("MCP_SERVER_PORT", "9954"))
            env_vars["MCP_SERVER_URL"] = f"http://host.docker.internal:{mcp_port}"
        elif self.config.network_mode == "host"   ```

is the same for both linux and mac cases. that's why.

ermias19: Since ruff is actually affecting alot of files, may i revert the linting and formating changes made by ruff ?? so we could raise the change on a separate PR as you said. WDYT?

Saadmrp1038: @ermias19 nah it's fine for this one. I reviewed the changes already.

Saadmrp1038 (approved):

Saadmrp1038:
@ermias19 Thanks for addressing all my comments! Merging this now.

PR #874, opened 2026-06-05
Updated default mitigation oracle

### LINKED ISSUE #853: Reward Hacking with Default Mitigation Oracle
The default mitigation oracle simply checks if all pods are in a running state and if we downscale the problematic deployment to 0 replicas or delete it altogether, an agent could technically still pass the benchmark. 

Example reproduction:

```bash
uv run python cli.py
start service_port_conflict_hotel_reservation
submit() # the diagnosis is irrelevant here
kubectl delete deployment recommendation -n hotel-reservation # or alternatively kubectl scale deployment recommendation -n hotel-reservation --replicas=0
submit()
```

The mitigation oracle will accept this fix which should not be the case since the problem hasn't really been solved. A potential fix for this problem could be to simply check that all of the deployments specific to the application are up and running as expected.

vsmart-06:
Fixes [#853](https://github.com/SREGym/SREGym/issues/853)

Updated the default mitigation oracle to also check that all the deployments in the namespace before the fault was injected are still maintained. This is slightly stricter than `DeploymentReadinessOracle` but that can potentially be replaced with this default oracle if such strictness is required.


Saadmrp1038:
Reviewing this currently. Might take some time to see that there are no unwanted side effects.

Saadmrp1038 (approved):

Saadmrp1038:
@vsmart-06 Thanks for the PR! There were some edge cases. I addressed those.
- Only verify that original deployments still exist. Don't fail on extra ones added by faults (handles `node_conntrack_exhaustion`)
- Check health against current `spec.replicas` instead of snapshotted values. Faults can change replica counts during injection (handles `pod_anti_affinity_deadlock` and `duplicate_pvc_mounts`)
- Added explicit `desired == 0` check since the above change would otherwise let scale-to-zero pass (original PR didn't need this)
- Added `super().__init__()` call and diagnostic logging on failure paths

PR #875, opened 2026-06-05
fix: bind MCP port-forward to 0.0.0.0 for Docker container reachability

Saadmrp1038:
On native Linux, `host-gateway` resolves to `172.17.0.1` (Docker bridge), not `127.0.0.1`. The agent container connects via `host.docker.internal` which resolves to the bridge IP, but `kubectl port-forward` defaults to binding on `127.0.0.1` only.

This caused `MCP client entered context but session is not connected` errors for stratus agent since PR #789.

Fix: add `--address 0.0.0.0` so port-forward listens on all interfaces, reachable from Docker containers on any platform.

### DIFF
--- a/sregym/service/mcp_server.py
+++ b/sregym/service/mcp_server.py
@@ -104,7 +104,9 @@ class MCPServer:
                 time.sleep(3)
                 continue
 
-            command = f"kubectl port-forward svc/{self.service_name} {self.port}:9954 -n {self.namespace}"
+            command = (
+                f"kubectl port-forward svc/{self.service_name} {self.port}:9954 -n {self.namespace} --address 0.0.0.0"
+            )
             self.port_forward_process = subprocess.Popen(
                 command,
                 shell=True,


HacksonClark:
This doesn't cause the same issue we saw with 0.0.0.0 on (WSL I think), correct? I feel we should use the docker DNS names like docker.internal where we can. If this use of 0.0.0.0 doesn't cause the same issue than I will approve this.

Saadmrp1038:
@HacksonClark 
> This doesn't cause the same issue we saw with 0.0.0.0 on (WSL I think), correct?

Nah. I just added `--address 0.0.0.0`. It's an additive change. Just widens the listening surface (listen from all ports).

HacksonClark (approved):
Additive change as per https://github.com/SREGym/SREGym/pull/875#issuecomment-4636167471

Saadmrp1038:
> I feel we should use the docker DNS names like docker.internal where we can

Is that even possible here?

HacksonClark:
> > I feel we should use the docker DNS names like docker.internal where we can
> 
> Is that even possible here?

No, I don't think so.

PR #876, opened 2026-06-06
Derive the default problem namespace from the application

ermias19:
# [Refactor] Simplify Problem constructor initialization
 
## Summary
 
This PR simplifies problem constructor initialization by making `Problem.__init__()` derive the default namespace directly from the application object.
 
Previously, most problem classes repeated the same pattern:
 
```python
self.app = SomeApp()
super().__init__(app=self.app, namespace=self.app.namespace)
self.namespace = self.app.namespace
```
 
This PR centralizes that behavior in the base Problem class:
 
```python
def __init__(self, app, namespace: str | None = None):
    self.app = app
    self.namespace = app.namespace if namespace is None else namespace
```
 
Problem classes can now use the simpler form:
 
```python
super().__init__(app=SomeApp())
```
 
For parameterized multi-app problems, the local app-selection logic stays explicit:
 
```python
if app_name == "hotel_reservation":
    app = HotelReservation()
elif app_name == "social_network":
    app = SocialNetwork()
else:
    raise ValueError(...)
 
super().__init__(app=app)
```
 
---
 
## What changed
 
- Updated `Problem.__init__()` so `namespace` is optional and defaults to `app.namespace`
- Removed redundant `self.app = ...` and `self.namespace = self.app.namespace` assignments from problem constructors
- Simplified fixed-app problems to call `super().__init__(app=...)` directly
- Kept multi-app problem constructors explicit where `app_name` controls the selected application
- Removed unnecessary one-app `app_name` branches in fixed-app problems
- Preserved existing problem behavior — this is a constructor cleanup only
---
 
## Why
 
This reduces repeated boilerplate across problem classes and makes constructor behavior consistent. The `app` and `namespace` ownership now live in one place: the base `Problem` class.
 
---
 
## Validation
 
Compile check:
 
```bash
python3 -m compileall -q sregym/conductor/problems sregym/conductor/oracles
```
 
Unit tests:
 
```bash
uv run pytest tests/problems tests/oracles
```
 
```
32 passed
```

 
Constructor smoke test across the full problem registry:
 
```
checked=109   failures=0
```
 
Full lifecycle validation on an existing problem:
 
```bash
  --problem misconfig_app_hotel_res \
  --inject-timeout 180 \
  --recover-timeout 300 \
  --poll-interval 15
```
 
```
PASSED   deploy -> inject -> oracle fails -> recover -> oracle passes
```
 
---
 
## Notes
 
`tests/integration/smoke_test.py` currently fails because it expects the conductor to start at mitigation, while current conductor behavior defaults to diagnosis when no `tasklist.yml` exists. This is unrelated to the constructor cleanup.


Saadmrp1038:
LGTM!

Saadmrp1038 (approved):

Saadmrp1038:
@ermias19 can you check out the codeql check failure please?

---
/validate-problem skip

Saadmrp1038:
Fixed

PR #881, opened 2026-06-09
Add - native_sidecar_crashloop_hotel_reservation`

labid-al-nahiyan-avro:
Adds a new SREGym problem that simulates a Kubernetes 1.32 native sidecar container
(`initContainer` with `restartPolicy: Always`) whose `startupProbe` always fails for misconfigaration.
The sidecar enters `CrashLoopBackOff`. Because Kubernetes requires all native sidecars
to reach the `started` state before application containers begin, the main container
is permanently blocked in the `Init` phase. Old ReplicaSet pods continue serving
traffic, making the failure operationally quiet until the rollout is noticed as stuck.

## Real-World Failure Story

 A native sidecar is declared as an entry in `spec.initContainers`
with `restartPolicy: Always`, a fundamentally different lifecycle model from both
regular init containers (which must complete) and legacy sidecar patterns (plain
`spec.containers`).

**The failure this problem simulates:** a team adds an OTel collector native sidecar
with a `startupProbe` that reads `/tmp/otel-ready`. Due to a misconfiguration in the
collector's exporter pipeline, `/tmp/otel-ready` is never written. The `startupProbe`
always exits 1. After `failureThreshold` failures the kubelet restarts the sidecar,
which enters `CrashLoopBackOff`.

**The critical consequence:** Kubernetes requires all
`initContainers` (including native sidecars) to reach the `started` state before app
containers begin. A sidecar whose `startupProbe` never succeeds never becomes
`started`. The main application container is permanently blocked in the `Init` phase.

Old ReplicaSet pods continue serving traffic (RollingUpdate keeps them alive), so the
failure is operationally quiet. `kubectl get pods` shows `Init:0/1` on new pods
alongside `Running` old pods.

### Fault injection (`inject_fault`)

1. Save `frontend` deployment JSON to `/tmp/frontend_native_sidecar_original.json`.
2. Strategic-merge patch injects this `initContainer`:

```yaml
initContainers:
- name: otel-collector-sidecar
  image: busybox:1.28          # pinned; busybox ≥1.30 changed exit-code semantics
  restartPolicy: Always         # ← native sidecar declaration (K8s 1.32 GA)
  command:
    - /bin/sh
    - -c
    - echo otel-collector starting...; while true; do sleep 30; done
  startupProbe:
    exec:
      command: ["/bin/sh", "-c", "cat /tmp/otel-ready"]   # file never created
    initialDelaySeconds: 5
    periodSeconds: 5
    failureThreshold: 3
```

3. `kubectl rollout restart deployment/frontend` triggers a new ReplicaSet.
4. New pods enter `Init:0/1`. The sidecar starts, runs, but its `startupProbe` fails
   every 5 seconds. After 3 failures the kubelet kills and restarts the sidecar
   (exit code 137, SIGKILL). It enters `CrashLoopBackOff`. The main container
   remains permanently blocked at `PodInitializing`.

### Fault recovery (`recover_fault`)

1. Apply snapshot from `/tmp/frontend_native_sidecar_original.json`.
2. Fallback (if snapshot missing): JSON-patch removes `initContainers[0]` by index.
3. `rollout restart` + `rollout status --timeout=120s`.

### Files changed

| File | Change |
|---|---|
| `sregym/generators/fault/inject_virtual.py` | `inject_native_sidecar_crashloop` / `recover_native_sidecar_crashloop` |
| `sregym/conductor/problems/native_sidecar_crashloop_hotel_reservation.py` | NEW |
| `sregym/conductor/oracles/native_sidecar_mitigation.py` | NEW |
| `sregym/conductor/problems/registry.py` | 1 import + 1 entry |
| `Problem List.md` | 1 row under Misconfiguration |

No changes to base classes, no new Python dependencies.

---

## 3. Problem Runtime Behavior

```
$ kubectl get pods -n hotel-reservation
NAME                                    READY   STATUS                  RESTARTS   AGE
frontend-755dc48b8d-t8glr               0/2     Init:CrashLoopBackOff   5          4m58s
frontend-7fc7ddc9db-9n2l2               1/1     Running                 0          5m23s
consul-5c767d7459-snj8v                 1/1     Running                 0          5m24s
...
```

Two `frontend` pods coexist: the old ReplicaSet pod is `Running` and serving traffic;
the new ReplicaSet pod is stuck indefinitely.

```
$ kubectl describe pod frontend-755dc48b8d-t8glr -n hotel-reservation

Init Containers:
  otel-collector-sidecar:
    State:      Waiting
      Reason:   CrashLoopBackOff
    Last State: Terminated
      Reason:   Error
      Exit Code: 137              # kubelet SIGKILL after probe failure
    Ready:      False
    Restart Count: 5
    Startup: exec [/bin/sh -c cat /tmp/otel-ready] delay=5s period=5s #failure=3

Containers:
  hotel-reserv-frontend:
    State:   Waiting
      Reason: PodInitializing    # blocked; never starts

Conditions:
  Initialized: False             # native sidecar never reached 'started'
  Ready:       False

Events:
  Warning  Unhealthy  ...  Startup probe failed: cat: can't open '/tmp/otel-ready': No such file or directory
  Normal   Killing    ...  Init container otel-collector-sidecar failed startup probe
  Warning  BackOff    ...  Back-off restarting failed container otel-collector-sidecar
```

```
$ kubectl logs frontend-755dc48b8d-t8glr -n hotel-reservation -c otel-collector-sidecar
otel-collector starting...
```

The sidecar process is alive and healthy. The `startupProbe` of sidecar is the sole kill
mechanism. This distinguishes it clearly from a container that crashes on startup.

**Diagnosis :**

```bash
# Step 1 — spot the asymmetry: one Running (old RS), one Init:CrashLoopBackOff (new RS)

# Step 2 — drill in: sidecar CrashLoops, main container PodInitializing
kubectl describe pod <new-frontend-pod> -n hotel-reservation

# Step 3 — confirm native sidecar declaration in deployment spec
kubectl get deployment frontend -n hotel-reservation -o yaml | grep -A 10 initContainers
# restartPolicy: Always  ← the K8s 1.32 native sidecar field

# Step 4 — check probe and log
kubectl logs <new-frontend-pod> -n hotel-reservation -c otel-collector-sidecar
# shows: otel-collector starting...  (process is alive; probe is what kills it)
```

**Mitigation:**

```bash
kubectl patch deployment frontend -n hotel-reservation \
  --type=json \
  -p='[{"op":"remove","path":"/spec/template/spec/initContainers/0"}]'

kubectl rollout status deployment/frontend -n hotel-reservation
# deployment "frontend" successfully rolled out
```

**Mitigation oracle checks two conditions** to prevent reward-hacking:
1. All pods in the namespace are `Running`.
2. The `frontend` deployment spec no longer contains an `initContainer` named
   `otel-collector-sidecar`so that the broken sidecar would not remain in the spec and does not reappear on the next rollout.


Saadmrp1038:
@labid-al-nahiyan-avro Thanks for the PR!

The plan sounds fine, but I think it's too close to [`init_container_dependency_hang_hotel_reservation`](https://github.com/SREGym/SREGym/blob/main/sregym/conductor/problems/init_container_dependency_hang.py). The steps to fix it look the same: new pods stuck in `Init` -> describe pod -> remove the bad init container. The native sidecar + startup probe is just another way to break an init container, but the agent doesn't have to understand any of that to fix it, so I don't think it tests anything new.

Unless it's possible to do something different here it's too close to what we already have. Also, the generic [`MitigationOracle`](https://github.com/SREGym/SREGym/blob/main/sregym/conductor/oracles/mitigation.py) already fails when pods are stuck in `Init`, so you likely don't need a new oracle.

Since it overlaps an existing problem, I'll close the PR.

---

Also, from initial overview the PR has some issues:

1. You uncommented `containerdConfigPatches` in `kind/kind-config-arm.yaml` and pointed the `docker.io` mirror at `http://128.105.144.30:5000`. That's probably the private registry on your own test environment. With this format containerd still falls back to Docker Hub when the mirror is unreachable, so pulls won't outright break. But everyone else's cluster will hit your dead mirror first and eat a connection timeout on every image pull.

2. The registry key (`native_sidecar_crashloop_hotelreservation_misconfig`) doesn't match the name in the Problem List.

3. And the description has odd newlines that look copy-pasted straight from Claude.

Please scope the PR to just the fault (unless required for the fault) and double-check what you're committing.

Issue #852, opened 2026-06-02
[Problem RFC] Missing imagePullSecret Dependency Fault


W-OK-E:
Hello everyone,

I am submitting this RFC as part of the UIUC++ SRSE evaluation task for a new SREGym benchmark problem.

## Proposal: Injecting Missing `imagePullSecret` Dependency Fault

I propose adding a new standalone object-layer authentication fault: **Missing or Invalid `imagePullSecret`**. The inspiration for this issue comes from **SREGym Issue #201** and is modeled after the cluster troubleshooting signals introduced in **Kubernetes PR [#117927](https://github.com/kubernetes/kubernetes/pull/117927)**. This PR updated the Kubelet to emit an explicit warning event (`FailedToRetrieveImagePullSecret`) when a referenced secret cannot be found—providing a clear object-layer signature that agents can leverage to diagnose underlying registry authentication issues.

---

### Why This is a Great Fit for SREGym

* **Addresses an Explicit Gap:** Gap analysis actively tracks `imagePullSecret` missing as an open fault class with zero current registry coverage.
* **Distinct Fault DNA:** The benchmark already includes image rollout faults like `incorrect_image`. However, this new fault features a completely valid image tag. The failure happens exclusively at the cluster object dependency layer, forcing an agent to debug the Kubernetes secret infrastructure rather than the deployment manifest.
* **Evaluates Real-World Correlation (PR #117927 Update):** Following Kubernetes PR #117927, Kubelet now emits a warning when a secret is missing. However, because this event is notoriously noisy and can trigger on perfectly healthy pods (e.g., if the image is public or alternate credentials succeed), this task forces the agent to contextually evaluate whether the log is an actual root cause or just noise based on the pod's true runtime state (`ImagePullBackOff`).

---

### Target Application

**HotelReservation** (`hotel-reservation` namespace).

Rationale from the existing problem inventory:

* HotelReservation hosts all of the benchmark's raw-KubeCtl problems — `network_policy_block`, `admission_webhook_outage`, `admission_webhook_tls_mismatch`, `expired_tls_hotel_reservation`, `node_conntrack_exhaustion`, `pod_cidr_exhaustion`. An imagePullSecret fault is a cluster-credential object dependency, placing it naturally in this family.
* The app uses a single shared image (`yinfangchen/hotelreservation:latest` on Docker Hub) for all its microservices, which simplifies targeting a single Deployment without tracking multiple image references.
[...]

PR #883, opened 2026-06-09
[Problem RFC] Missing Image Pull Secrets in hotel-reservation app

W-OK-E:
# Add Problem: Missing ImagePullSecret (Hotel Reservation)

This PR is based on the discussion in #852. Grateful to @Saadmrp1038 for all the suggestions and answering my questions. Happy to make any other changes if required. Description follows.
 
## Summary

- Adds `MissingImagePullSecretHotelReservation` — a new SREGym fault-injection problem
  where an agent must diagnose and fix a pod stuck in `ImagePullBackOff` due to a
  deleted imagePullSecret.
- Adds `ImagePullSecretMitigationOracle` to evaluate whether the agent's fix is correct.
- Registers the problem in the conductor registry.
- Ensures Cleanup: if the problem does not complete, appropriate measures in place to restore the cluster to its original state.

## Changed files

| File | Change |
|------|--------|
| `sregym/conductor/problems/missing_image_pull_secret_hotel_reservation.py` | New problem |
| `sregym/conductor/oracles/image_pull_secret_mitigation.py` | New mitigation oracle |
| `sregym/conductor/problems/registry.py` | Register the new problem |
| `sregym/conductor/conductor.py` | Wire cleanup into `fix_kubernetes()` |

### Infrastructure setup (`inject_fault`)

The problem creates a realistic private-registry scenario before injecting the fault:

1. **Private registry** — starts `sregym-private-registry` (Docker `registry:2` with
   htpasswd/bcrypt auth) on the Kind Docker bridge network. Configures containerd on
   every Kind node to accept HTTP pulls from it (`/etc/containerd/certs.d/<ip>:5000/`).

2. **Image relocation** — pulls the existing `hotel-reservation` image from the
   in-cluster containerd store, re-tags and pushes it into the private registry via
   the host-mapped port (`localhost:5001`). Repoints the `recommendation` Deployment
   to pull from `<registry-ip>:5000/hotel-reservation:latest` with
   `imagePullPolicy: Always`.

3. **Valid secret first** — creates `hotel-registry-creds` (a `kubernetes.io/dockerconfigjson`
   secret) and verifies the pod comes up healthy. This confirms the full plumbing works
   before any fault is injected.

4. **Discovery breadcrumbs** — creates a separate `infra-registry` namespace containing:
   - `private-registry-creds-master` secret (a copy of the valid credentials)
   - `registry-access-runbook` ConfigMap linking to the master secret

   An agent following events/logs naturally finds `FailedToRetrieveImagePullSecret`,
   then the runbook, then the master secret, then recreates the missing per-namespace
   secret. This mirrors the real-world on-call pattern.

5. **Docker Hub block** — starts `sregym-dockerhub-block` (nginx returning 403) on the
   Kind bridge network with DNS alias `dockerhub-mirror.internal`. Configures containerd
   on every Kind node to use this as a mirror for `docker.io`
   (`/etc/containerd/certs.d/docker.io/hosts.toml`). This prevents an agent from
   bypassing the secret by simply repointing the image to a public registry.

6. **The fault** — deletes `hotel-registry-creds`. Forces a pod restart. The pod
   immediately enters `ImagePullBackOff`.

7. **Decoy pods** — creates two pods that emit `FailedToRetrieveImagePullSecret` warning
   events but stay `Running` (they reference the decoy secret `nonexistent-pull-creds`,
   which doesn't exist, but their image is cached so the pull succeeds). An agent must
   correlate the warning event with actual pod phase to identify the one genuinely broken
   workload.

### What the agent is expected to do

The agent observes `recommendation` in `ImagePullBackOff`, investigates the events
(`FailedToRetrieveImagePullSecret` for `hotel-registry-creds`), navigates the
discovery path (`registry-access-runbook` ConfigMap → `infra-registry` namespace →
`private-registry-creds-master`), and recreates the missing secret in the application
namespace with the correct credentials. The pod should return to `Running`.

The decoy pods are a distractor: the agent must not be misled by other pods emitting
the same warning event into patching the wrong workload or concluding the problem is
already resolved.

### Recovery (`recover_fault`)

Restores the Deployment to its original public image and `imagePullPolicy`, removes
the private registry, removes the containerd mirror configs from all Kind nodes,
removes the `infra-registry` namespace, and deletes the decoy pods.

## Mitigation oracle (`ImagePullSecretMitigationOracle`)

Evaluates whether the agent's fix is complete. Checks in order:

1. `hotel-registry-creds` secret exists in the application namespace.
2. Secret is well-formed: valid base64 `.dockerconfigjson` with a non-empty `auths` key.
3. `recommendation` Deployment references the secret in `spec.template.spec.imagePullSecrets`.
4. All `recommendation` pods are `Running` and every container is `Ready`.

## Additional changes: containerd docker.io redirect survives Ctrl+C

The Docker Hub block writes `/etc/containerd/certs.d/docker.io/hosts.toml` on each
Kind node — a node-filesystem file with no corresponding Kubernetes object. Namespace
teardown and app undeploy do not touch it. If the conductor is killed (Ctrl+C) before
`recover_fault` runs, the redirect persists on the next startup and breaks ALL public
image pulls cluster-wide (e.g. `wrk2-job` enters `ImagePullBackOff` immediately,
completely unrelated to the intended fault).

**Fix**: four teardown methods (`_stop_registry`, `_get_kind_nodes`,
`_remove_containerd_dockerhub_block`, `_stop_dockerhub_block`) were converted to
`@staticmethod`s. A new `cleanup_leftovers()` staticmethod composes them. This is
called from `Conductor.fix_kubernetes()` at every startup, unconditionally, following
the existing pattern used by `PodCIDRExhaustionHotelReservation`,
`KubeletEvictionThresholdMisconfig`, and others. All operations are idempotent.

## Testing

- **Normal path**: `inject_fault` → pod enters `ImagePullBackOff` → agent recreates
  secret → `mitigation_oracle` passes → `recover_fault` → cluster restored cleanly.
- **Re-injection**: after `recover_fault`, `inject_fault` succeeds again without
  residual state.
- **Ctrl+C regression (manually verified)**: inject fault → exit without `recover_fault`
  → restart conductor → `fix_kubernetes()` logs
  `[FIX] MissingImagePullSecret containerd dockerhub-block leftover if any` →
  containerd `docker.io` config removed from all nodes → `wrk2-job` and all other pods
  come up `Running` → only the target pod enters `ImagePullBackOff` after re-injection.
- **Decoy isolation**: three pods emit `FailedToRetrieveImagePullSecret`; only
  `recommendation` is actually stuck. The two decoys are `Running`/`Ready`.


Saadmrp1038:
@W-OK-E Thanks for the PR! I will review it tomorrow.

Saadmrp1038 (changes requested):
@W-OK-E This currently only works on kind since it uses `docker exec` for node-level changes and the Docker bridge for networking. You will need to implement logic for real-cluster path. If possible can you explore an implementation where the kind vs real cluster difference is minimal? Most of the logic should be shareable, just the node access method differs.

W-OK-E:
Sure! On it.

W-OK-E:
Hello @Saadmrp1038 !

For the actual cluster implementation, I went on with aws EC2 instances and kept running into a issue regardless of what problem is launched at hand:

During the application deployment (`hotel-reservation` in this case) , it kept surfacing as `deploy_app()` → `wait_for_ready("kube-system")` timing out after 600s
because `calico-kube-controllers` is stuck in CrashLoopBackOff (the `Tier(default)` /
`Tiers.crd.projectcalico.org` FATAL), with
`kubectl get crd tiers.crd.projectcalico.org adminnetworkpolicies.policy.networking.k8s.io`
returning `NotFound`.

For which, I added `_repair_calico_crds()` to `conductor.py` as follows:

```python
 def _repair_calico_crds(self):
        """Re-apply the Calico manifest if calico-kube-controllers' required CRDs are
        missing. No-op on clusters not running manifest-based Calico (e.g. kind /
        kindnet), and idempotent when the CRDs are already present."""
        deploy = self.kubectl.exec_command(
            "kubectl get deploy -n kube-system calico-kube-controllers --ignore-not-found -o name"
        )
        if "calico-kube-controllers" not in (deploy or ""):
            return  # not a Calico cluster -- nothing to repair

        missing = [
            crd
            for crd in self._CALICO_REQUIRED_CRDS
            if crd
            not in (self.kubectl.exec_command(f"kubectl get crd {crd} --ignore-not-found -o name") or "")
        ]
        if not missing:
            return

        self.logger.warning(
            f"Calico CRDs missing {missing}; re-applying Calico manifest to recover calico-kube-controllers"
        )
        self.kubectl.exec_command(f"kubectl apply -f {self._CALICO_MANIFEST_URL}")
        # Force a fresh pod so it picks up the recreated CRDs immediately instead of
        # waiting out its CrashLoopBackOff backoff (which can be minutes).
        self.kubectl.exec_command(
            "kubectl delete pod -n kube-system -l k8s-app=calico-kube-controllers --ignore-not-found"
        )
```


Have you faced this issue before, or could this be due to my specific choice of aws as a cluster?  

PS - The code has not been committed yet to this PR so if you think I should report it elsewhere, let me know!

Saadmrp1038:
@W-OK-E Hmmm. I didn't face it before. Might be somehow related to using EC2s. When you ran the ansible setup script was there any error during Calico installation?

W-OK-E:
Hey @Saadmrp1038 , apologies for the delayed response, since I did not recall there being an explicit issue with calico installation, I went on to reproduce the error on 5 different sets of AWS clusters and managed to get some sort of an explanation. 

A few differences between my AWS and your cloudlab setup maybe worth highlighting here:


| Aspect | CloudLab | My AWS port |
| --- | --- | --- |
| Home dir convention | `/users/$USER` (CloudLab) | `/home/ubuntu` |
| Kernel | playbook *asserts* Ubuntu generic `6.8.0-generic` | assert disabled (`need_kernel_switch: false`, kernel assertion commented out — the kernel switch broke the AWS instances), so nodes run **`6.17.0-1017-aws`** |
| Disk under etcd | CloudLab **local SSD/NVMe** | **AWS EBS** (network-attached block storage) |
| Runtime | cri-dockerd / Docker (`--cri-socket /var/run/cri-dockerd.sock`, `kubeadm init` line ~536) | same |


On every new cluster, when ` ansible-playbook -i inventory.yml setup_cluster.yml` is run, Calico installs fine, and all its CRDs are present and its controller is healthy. The issue I found was that **two of Calico's CRDs get deleted from etcd later**, every time SREGym sets up a problem, because the deploy step **re-runs `kubectl apply` for the metrics-server and OpenEBS add-ons on every run**. 

My hypothesis currently is that on my AWS clusters, that re-apply burst pushes etcd hard enough that two specific CRDs (`tiers.crd.projectcalico.org` and `adminnetworkpolicies.policy.networking.k8s.io`) get knocked out of etcd. Once they're gone, `calico-kube-controllers` crash-loops and the `kube-system` readiness check hangs for 10 minutes and fails. 

 I think the root-cause is the environment my ported playbook runs in. CloudLab nodes use **local SSD/NVMe disk**; the AWS nodes run etcd on **EBS (network-attached storage)**, which has much higher and
spikier write-flush latency. etcd is extremely sensitive to that. I also had to **turn off the forced kernel** in `setup_cluster.yml` (it broke the AWS instances), so my nodes run a different kernel (`6.17-aws`) than the one you validate on (`6.8-generic`).
 
 **What I have tested / ruled out on 5 separate AWS Clusters:**
 - Confirmed via direct `etcdctl` that the two CRDs are genuinely **deleted from etcd** (not just an apiserver cache glitch), while the other ~18 Calico CRDs survive.
 - Confirmed via the apiserver log that the deletion happens **in the same 8-second  window** as the deploy step's metrics-server/OpenEBS `kubectl apply`.
 - Confirmed via `kubectl top nodes` that it is **not memory/CPU pressure** (<25% used).
 - Confirmed the Calico install itself is correct (clean on first run).
 - On the storage part hypothesis: **etcd fsync latency on EBS** causes apiserver timeouts mid-CRD-establishment during the apply burst, dropping those two CRDs. (Confirmable via   etcd `slow fdatasync` log lines / `etcd_disk_wal_fsync_duration_seconds`  causing the next cluster to timeout.)

 **Current Patch in my fork:** 
1. Don't re-apply metrics-server/OpenEBS if they're already installed which removes the burst that triggers it. 
2. Auto-recreate the two CRDs right before the `kube-system` health check so that it self-heals if they ever go missing. Both are independent, so the symptom no longer blocks runs even on the EBS/kernel combination.


Please let me know if you have any questions/suggestions on my analysis.

Saadmrp1038:
@W-OK-E Thanks for the detailed testing! This scenerio is really unexpected :)
For now let's keep the fixes you made.

W-OK-E:
No worries @Saadmrp1038 !! 
I am learning a ton in the process, I will update the PR once a few more tests are complete.

W-OK-E:
Hey @Saadmrp1038 !

I tested the problem on kind as well as AWS clusters and have pushed the changes, feel free to comment or suggest changes. Summary of what was done:

1. Tested on multiple separate AWS Clusters from end to end. Tested on kind too post the cluster port to ensure that the kind setup did not break.
2. Fixed a `kube-system` deply timeout bug **only** visible on EC2 instances - possibly because of removal of critical CRDs which was fixed with a simple addition to `conductor.py` that re-applies the Calico manifest to ensure the CRDs are present.
3. The metrics and the openebs manifests are not reapplied if they are already present. The reapplication could possibly be a cause behind the race condition and removal of CRDs.


However, I sort of feel that the explanation for *why* the two Calico CRDs lose the apiserver discovery-resync race on AWS but not CloudLab is flaky; the swapped kernel is another possible clue. I will spin up a few instances from other cloud providers maybe and see if it's the same there.

P.S.  - I tried testing Straus on this problem but hit rate limit after the first few API calls :( Please let me know if and when you do test an agent on this and how it goes.

sregym/conductor/problems/missing_image_pull_secret_hotel_reservation.py
```diff
@@ -0,0 +1,1001 @@
...
+    def _create_decoy_pods(self) -> list[str]:
+        """Create bare pods that emit FailedToRetrieveImagePullSecret but stay Running."""
+        names = []
+        for i in range(2):
+            name = f"hotel-res-logger-{i}"
+            pod = client.V1Pod(
+                metadata=client.V1ObjectMeta(
+                    name=name,
+                    namespace=self.namespace,
+                    labels={"sregym-decoy": "true"},
```
Saadmrp1038: nit: We should use a realistic/generic name here and not expose the `sregym` label.

sregym/conductor/problems/missing_image_pull_secret_hotel_reservation.py
```diff
@@ -0,0 +1,1001 @@
...
+from sregym.conductor.problems.base import Problem
+from sregym.service.apps.hotel_reservation import HotelReservation
+from sregym.utils.decorators import mark_fault_injected
+
+logger = logging.getLogger(__name__)
+
+_REGISTRY_DEPLOYMENT = "sregym-private-registry"
+_REGISTRY_HTPASSWD_SECRET = "sregym-private-registry-htpasswd"
+_REGISTRY_USER = "sregym"
+_REGISTRY_PASS = "sregympass"
```
Saadmrp1038: nit: We should use a realistic/generic name here and not expose the `sregym` label. Also probably a real password.

sregym/conductor/oracles/image_pull_secret_mitigation.py
```diff
@@ -0,0 +1,101 @@
...
+import logging
+
+from kubernetes import client
+
+from sregym.conductor.oracles.base import Oracle
+
+logger = logging.getLogger(__name__)
+
+
+class ImagePullSecretMitigationOracle(Oracle):
```
Saadmrp1038, line 18: We should add importance = 1.0 class attribute. Not really needed here. Just to keep consistent with other oracles.

sregym/conductor/problems/missing_image_pull_secret_astronomy_shop.py
```diff
@@ -0,0 +1,1001 @@
...
+                )
+                logger.info(
+                    "Scaled down stale ReplicaSet '%s' for '%s' (revision %d) to 0",
+                    rs.metadata.name,
+                    _TARGET_DEPLOYMENT,
+                    revision(rs),
+                )
+
+    @staticmethod
+    def cleanup_leftovers() -> None:
```
Saadmrp1038, line 952: This method still needs more cleanup handling. After a mid-run crash or exit, the following are not cleaned up when another fault is run:

- `/etc/hosts` entries on cri-dockerd nodes (`# sregym-dockerhub-block` marker lines). Public Docker pulls stay broken on real clusters
- `infra-registry` namespace (master secret, htpasswd secret, runbook)
- Containerd insecure registry config (`/etc/containerd/certs.d/<ip>:5000/`)
- `registry-access-runbook` ConfigMap in app namespace
- Decoy pods in app namespace
- Deployment still repointed to private registry image (pod permanently stuck after registry is cleaned)
- `imagePullPolicy: Always` left on deployment

sregym/conductor/oracles/image_pull_secret_mitigation.py
```diff
@@ -0,0 +1,101 @@
...
+    def evaluate(self) -> dict:
+        print("== Mitigation Evaluation (Missing ImagePullSecret) ==")
+
+        problem = self.problem
+        namespace = problem.namespace
+        v1 = client.CoreV1Api()
+        apps_v1 = client.AppsV1Api()
+
+        # Check 1: imagePullSecret must exist
+        try:
```
Saadmrp1038: This feels too strict. The oracle requires a secret named exactly `hotel-registry-creds` and requires the Deployment to reference it. An agent could provide valid credentials via a differently-named secret or via the default ServiceAccount's imagePullSecrets. The pod would be Running but the oracle rejects it.

Saadmrp1038 (changes requested):
@W-OK-E the real cluster code is working well! Left some comments below.

> However, I sort of feel that the explanation for why the two Calico CRDs lose the apiserver discovery-resync race on AWS but not CloudLab is flaky

I will also test on AWS to see what's wrong. The deployment code might need to be more defensive to account for these weird unexpected behaviours.

> I tried testing Straus on this problem but hit rate limit after the first few API calls :( Please let me know if and when you do test an agent on this and how it goes.

Sure! I will do a set of agent runs after the oracle fix. That's the main blocker right now to doing the runs.

sregym/conductor/problems/missing_image_pull_secret_hotel_reservation.py
W-OK-E: Patched, will be pushed.

sregym/conductor/problems/missing_image_pull_secret_hotel_reservation.py
W-OK-E: Patched, will be pushed.

sregym/conductor/oracles/image_pull_secret_mitigation.py
W-OK-E: Patched, will be pushed.

W-OK-E:
Hey @Saadmrp1038 !!

As requested, the required changes were made and verified to run both on kind and real clusters.

1. To improve cleanup on unexpected exit etc.  `cleanup_leftovers()` now reconstructs all teardown state **from the cluster itself** (deployment `sregym.io/*` annotations, pod labels, `certs.d/*:5000` globs, presence-keyed `daemon.json` removal) instead of from in-memory instance state and so it works in a fresh process after a crash, with no saved registry IP / image / pod.

The following tests were done:

| # | Leftover | Check | Result |
|---|---|---|---|
| 1 | /etc/hosts dockerhub-block markers | grep -c sregym-dockerhub-block on all 3 nodes | 0 / clean + pod pulled a public image successfully |
| 2 | infra-registry ns (master/htpasswd secrets) | kubectl get ns infra-registry | NotFound |
| 3 | containerd insecure config | certs.d/*:5000 + daemon.json insecure-registries on all nodes | none / {"live-restore":true} |
| 4 | runbook ConfigMap | kubectl get cm registry-access-runbook | NotFound |
| 5 | decoy pods | kubectl get pods -l custom-logger=true | none |
| 6 | Deployment repointed to dead private image | image field | yinfangchen/hotelreservation:latest (original public), pod Running 1/1 |
| 7 | imagePullPolicy: Always | policy field | IfNotPresent (restored); annotations cleared; secret ref dropped |


2.  Oracle loosening - Replaced the strict mechanism checks with an **outcome check**: target pods Running+Ready so that legitimate alternate fixes are accepted.

There's a finding too here, on your point earlier:

> An agent could provide valid credentials via a differently-named secret or via the default ServiceAccount's   imagePullSecrets. The pod would be Running but the oracle rejects it.

The "default ServiceAccount imagePullSecrets" path does **not** work standalone for this fault. The fault pins `imagePullSecrets: [hotel-registry-creds]` in the Deployment template, and the built-in `ServiceAccount` admission plugin copies a ServiceAccount's imagePullSecrets into a pod **only when the pod declares none of its own** (it does merge/append but pod-level takes full precedence) and so the SA secret is never appended; kubelet only ever sees the deleted secret and the pod enters `ImagePullBackOff`.


Let me know if I missed something here or if more changes are required.

sregym/conductor/oracles/image_pull_secret_mitigation.py
W-OK-E: The "default ServiceAccount imagePullSecrets" path does not work standalone for this fault (detailed below). However, the oracle has been relaxed and does not demand strict names equality for secrets and registries.

sregym/conductor/problems/missing_image_pull_secret_astronomy_shop.py
W-OK-E: `cleanup_leftovers` has been tightened and tested.

sregym/conductor/oracles/image_pull_secret_mitigation.py
Saadmrp1038: @W-OK-E I see. I am guessing deployment patch + SA would've worked then (?). Although at that point you would have to ask what the agent is even doing XD

sregym/conductor/oracles/image_pull_secret_mitigation.py
Saadmrp1038: @W-OK-E I think this is still missing (again it does not really matter here just for consistency).

Saadmrp1038:
@W-OK-E Nice work! I am testing again.

---
A nit. Let's change all labels that have `sregym` word in it to something else. That means `sregym.io/*`, `sregym-dockerhub-*`, `sregym-node-op-*` etc.

Saadmrp1038 (changes requested):
@W-OK-E when the image is already cached on the nodes from previous runs we should be able to directly push it to the private registry instead of pulling from Docker Hub. That would be way faster. Can you please test it out?

Saadmrp1038 (commented):
@W-OK-E I ran some agent tests. The `sregym.io/fault-active`,  `sregym.io/original-image` and `sregym.io/original-image-pull-policy` annotations on the deployment are giving the agent the answer. It reads the labels and understands that it's an intentional fault injection and can also understand what was injected. It reverts the fault with kubectl patch by changing the image ref to the public one and the image policy to `IfNotPresent`. The public image is still cached on the node so `IfNotPresent` means no pull happens and the Docker Hub block never triggers. The app becomes healthy and the agent submits.

Oracle does catch it (anti-cheat works), but the agent thinks it solved the problem.

Two things should be done:

1. The annotations shouldn't contain `sregym` or expose the fault details directly.
2. Clean up the cached public image from nodes during injection so that the public ref actually breaks.

W-OK-E:
@Saadmrp1038 I have a slightly different plan in mind, which stems from my current analysis:

1. Cleaning up the cached images from the nodes would work but has potential issue - if any of the other pods relying on the image (with PullPolicy `IfNotPresent`)  fail/are deleted/restart, they will enter `ImagePullBackOff`, and the agent would need to redirect it to the private registry and setup auth for them too. Although interesting, I need your advice on whether this happens often enough or we can just ignore this and remove the cached image.

2. Another idea I had was why not just make everything pull from the private registry with only one pod missing the ImageSecrets. But this would introduce an unwanted workaround, the agent can just copy the image secret from the namespace for other pods and apply it to the `recommendation` pod and that would be pretty easy I think.

3. I think the most failsafe approach would be if I could extend the hotel-reservation app and add a new pod that can ONLY ever pull from a private registry, it would still involve pulling a public image (so that we don't introduce long-term auth dependencies on private docker images) and pushing that into a registry, but that image would be unique and so we can remove it from cache without affecting any other pods. I understand that I must be careful so as to not break the hotel-reservation app, but I can try. Which means in default state - the hotel-reservation will have the secret stored, and fault-injection would just delete the secret associated with the pod and launch decoys, but I'd need  your advice on this.

Let me know what path seems the most feasible to you!

Saadmrp1038:
@W-OK-E I gave it some thought. Your points are valid. It slipped my mind that `hotel-reservation` uses one image for everything. Rather than trying to force it here, which requires us to make weird complex decisions, can
 we simply switch to another app that uses unique images for each service? That should make everything much simpler. Check out `astronomy-shop`.

W-OK-E:
Hey @Saadmrp1038 ! I have implemented the problem in `astronomy-shop` as well as `blueprint-hotel-reservation`   for reasons described below.

## What Changed

- Two new `missing_image_pull_secret` problem variants that fix the shared-image purge problem in the original hotel-reservation (8 services share one image, so purging the target's image breaks all siblings).
- **astronomy-shop**: each service has a unique tag (`:2.2.0-payment`), so purging + blocking is fully isolated; disables grafana/jaeger/shrinks opensearch via `_CapacityFittedAstronomyShop` to fit small clusters; blocks both docker.io + ghcr.io. 
- **blueprint-hotel-reservation**: each service has a unique repo (`docker-geo-service-container`), and it is considerably lighter than `astronomy-shop` (~8 pods, no Helm, no resource requests); blocks docker.io only; includes readiness gate before the block to prevent collateral ImagePullBackOff on sibling services.
- Oracle (`image_pull_secret_mitigation.py`) now reads `problem.target_pod_label` and `problem.gated_image_re` instead of hardcoding hotel-reservation values and is also loosened to allow the pathways discussed earlier to work.
- Shell state persistence (`shell.py`): exported vars/functions/cwd carry across CLI commands; fixed predictable tmpfile (now mkstemp + 0600 + atexit cleanup) and cwd-override bug (explicit cwd now resolved to absolute path before applying). Present in #903 
- Both problems registered in `registry.py` and `conductor.py` (`cleanup_leftovers` in `fix_kubernetes`).
- Have also removed the old variant on `hotel-reservation` app.

## What I have tested

- Both variants: inject → oracle-fail → legitimate fix → oracle-pass → recover → cleanup on AWS (cri-dockerd) and kind clusters.
- Anti-cheat verified: repoint-to-public blocked (image purged + registry blocked), flip-to-IfNotPresent blocked (private image also purged), repoint-to-nginx rejected by oracle gate.
- Decoy pods: blueprint uses `registry.k8s.io/pause:3.9` (777lefty images lack `sleep`/shell); astronomy-shop uses sibling otel-demo images (have coreutils).
- `cleanup_leftovers()` crash-recovery path verified standalone (with `_ensure_kube_config` fix for static invocation).
- Purge moved before docker.io block in inject sequence — fixes silent purge failure when busybox node-op pod can't pull from blocked docker.io. Verified that both the public and the private images are purged and the agent cannot just pull from cache.
- One minor caveat : The pull-secret registries use `nginx:alpine` public docker images which are cached on the nodes. And the agent can indeed choose to just point to that and get the pod running, but the oracle would not pass it.

W-OK-E:
Also, I am aware that the current changes are extremely bulky with docstrings and comments, but they just address some of the points that we have discussed in this PR. Besides any other changes, I will clean them up too once you give a go @Saadmrp1038.

sregym/conductor/problems/missing_image_pull_secret_astronomy_shop.py
```diff
@@ -0,0 +1,1415 @@
...
+        ]
+        for s in self._EXTRA_HELM_SETS:
+            extra_args += ["--set", s]
+        self.helm_configs["extra_args"] = extra_args
+
+        Helm.install(**self.helm_configs)
+        Helm.assert_if_deployed(self.helm_configs["namespace"])
+
+
+class MissingImagePullSecretAstronomyShop(Problem):
```
munimthahmid, line 141: I’m not sure we need this problem for both Astronomy Shop and Blueprint Hotel Reservation. The failure and expected diagnosis are basically the same, and most of the implementation is duplicated between the two files.

Since Blueprint is much lighter and already gives us a different image for each service, I would prefer keeping the Blueprint version and dropping the Astronomy version. Is there a specific benchmark reason why we need both?

sregym/conductor/problems/missing_image_pull_secret_blueprint_hotel_reservation.py
```diff
@@ -0,0 +1,1444 @@
...
+        secret = client.V1Secret(
+            metadata=client.V1ObjectMeta(name=_SECRET_NAME, namespace=self.namespace),
+            type="kubernetes.io/dockerconfigjson",
+            data={".dockerconfigjson": base64.b64encode(dockerconfig.encode()).decode()},
+        )
+        try:
+            self.core_v1.create_namespaced_secret(namespace=self.namespace, body=secret)
+        except client.exceptions.ApiException as e:
+            if e.status == 409:
+                self.core_v1.replace_namespaced_secret(name=_SECRET_NAME, namespace=self.namespace, body=secret)
```
munimthahmid, line 1281: I think recovery will fail if the agent has already recreated `geo-service-registry-creds`, which is also the fix we expect it to make.

`create_namespaced_secret()` will return 409, but this new Secret object does not have the existing `resourceVersion`, so the replace request will also fail. Could we read the existing Secret and copy its `resourceVersion`, or patch it instead? It would also be good to test recovery after the agent has recreated the Secret.

tianyin:
@munimthahmid : @W-OK-E informed us earlier that he may not have time to continue working on the problem. 

Are you trying to take the problem over and merge it? (that will be great!)

If you are purely doing the review and expect @W-OK-E to revise it PR, it may not happen.

W-OK-E:
> @munimthahmid : @W-OK-E informed us earlier that he may not have time to continue working on the problem.
> 
> Are you trying to take the problem over and merge it? (that will be great!)
> 
> If you are purely doing the review and expect @W-OK-E to revise it PR, it may not happen.

Hey @tianyin, I can review and refine the PR over the weekends, that would be totally fine for me!

sregym/conductor/problems/missing_image_pull_secret_astronomy_shop.py
W-OK-E: Ahh yes! 
You are right. It's just that the blueprint application can run quicker and on low-end systems since I had some issues getting Astronomy shop and other applications running. 

We can choose to not include the `astronomy-shop` implementation.

tianyin:
Orz

sregym/conductor/problems/missing_image_pull_secret_blueprint_hotel_reservation.py
W-OK-E: I will check and make the changes.

sregym/conductor/problems/missing_image_pull_secret_blueprint_hotel_reservation.py
```diff
@@ -0,0 +1,1444 @@
...
+        Safe to call repeatedly; a no-op once a config is already active. Needed by the
+        static crash-recovery path, which can run before any instance has loaded config.
+        """
+        try:
+            config.load_incluster_config()
+        except config.ConfigException:
+            config.load_kube_config()
+
+    @staticmethod
+    def cleanup_leftovers() -> None:
```
munimthahmid, line 950: I’m a little worried about this cleanup because `fix_kubernetes()` runs it on every conductor startup.

It removes all `certs.d/*:5000` entries and the full `docker.io` and `ghcr.io` directories. The cri-dockerd path also replaces the complete `daemon.json`. If the cluster already has a registry mirror or another insecure registry configured, we will remove that unrelated configuration.

Could we save the original node configuration before changing it and restore that exact state during cleanup?

W-OK-E: There were a bunch of alternatives here, however most would break other problem statements at startup. However, your concern is valid, I will recheck and see if there is a workaround or try your suggested pathway.

sregym/conductor/oracles/image_pull_secret_mitigation.py
```diff
@@ -0,0 +1,106 @@
...
+        target_deployment = problem.target_deployment
+        target_container = problem.target_container
+        # Per-problem overrides; default to the original hotel-reservation values so
+        # the legacy problem keeps working unchanged.
+        pod_label = getattr(problem, "target_pod_label", None) or _DEFAULT_POD_LABEL
+        gated_image_re = getattr(problem, "gated_image_re", None) or _DEFAULT_GATED_IMAGE_RE
+
+        # Primary criterion: the target workload's pods must be Running + Ready with
+        # no image-pull failure. Accepts ANY working credential mechanism.
+        pods = v1.list_namespaced_pod(
```
munimthahmid, line 44: I think the oracle can be bypassed here because it selects pods only by label, without checking whether they belong to the target Deployment.

For example, the target Deployment could be scaled to zero and replaced with a separate healthy pod using the same label. The oracle would still pass even though the actual workload is disabled.

Could we also check that the Deployment has at least one replica, its rollout is ready, and the pods belong to its ReplicaSet?

sregym/conductor/problems/missing_image_pull_secret_blueprint_hotel_reservation.py
```diff
@@ -0,0 +1,1444 @@
...
+            desired = deploy.spec.replicas or 1
+            status = deploy.status
+            observed_current = (status.observed_generation or 0) >= (deploy.metadata.generation or 0)
+            updated = status.updated_replicas or 0
+            total = status.replicas or 0
+            ready = status.ready_replicas or 0
+            if observed_current and updated == desired and total == desired and ready == desired:
+                return
+            time.sleep(3)
+        logger.warning("Deployment '%s' rollout not complete after %ds", deployment_name, timeout)
```
munimthahmid, line 1358: `_wait_for_deployment_ready()` only logs a warning when it times out, so injection continues even if the private registry and valid Secret never produced a healthy pod.

In that case, the benchmark may be testing a registry setup failure instead of the deleted image pull Secret. I think this baseline check should raise on timeout rather than continuing.

sregym/conductor/oracles/image_pull_secret_mitigation.py
munimthahmid: This one is still missing :) Could you add `importance = 1.0` here for consistency?

munimthahmid:
@W-OK-E Thanks for your patience, and sorry it took me so long to get back to this. I’ve left a few inline comments. Please let me know once you’ve addressed them, and I’ll take another look.

munimthahmid:
@W-OK-E, let me know if you still intend to work on the fixes, otherwise, I think I will fix the remaining issues and merge the PR if it's good enough

W-OK-E:
> @W-OK-E, let me know if you still intend to work on the fixes, otherwise, I think I will fix the remaining issues and merge the PR if it's good enough

Hey @munimthahmid ! Yes I am working on it right now. I apologize for the delay, just got a bit busy with the internship workload, changes will be pushed in a day or two.

tianyin:

PR #886, opened 2026-06-11
Add cumulative_admission_webhook_timeout_hotel_reservation problem

riyadomf:
# Add cumulative_admission_webhook_timeout_hotel_reservation

/validate-problem cumulative_admission_webhook_timeout_hotel_reservation

## The real-world failure

In October 2024, [Kubernetes issue #128162](https://github.com/kubernetes/kubernetes/issues/128162) documented a class of admission-controller failure that isn't as obvious as it sounds. A chain of `MutatingWebhookConfiguration`s, each individually set to `failurePolicy: Ignore`, can collectively block pod creation when their backends are unreachable. The issue was closed as "not planned". The upstream view is that operators should configure their webhook chains more carefully, but the trap remains current and reproduces on the 1.32.0 cluster used here.

Here is the picture an SRE actually walks into. Someone installed a handful of compliance webhooks months ago: a pod resource validator, an audit logger, an image policy checker, a tenant quota validator. Each was set to `failurePolicy: Ignore` so they would never block the cluster if they went unhealthy. Later, a security audit added a default-deny ingress `NetworkPolicy` to the namespace where the webhook backends live. Nothing broke immediately. Pods that were already running stayed running. The next time a Deployment in the affected namespace tried to recreate a pod (rolling update, eviction, OOM-kill), pod creation began failing with `Timeout: request did not complete within requested timeout - context deadline exceeded`.

The mechanism is cumulative, and that's the trap. Per-webhook `Ignore` only applies *after* that webhook's `timeoutSeconds` elapses. With four webhooks at 10 seconds each, the kube-apiserver waits 40 seconds before the first `Ignore` would fire. The apiserver's global admission deadline is about 30 seconds. The global timeout fires first, so admission fails outright. The error message names neither the webhooks nor the NetworkPolicy.

## The simulation

`inject_fault()` reproduces the scenario in eight steps:

1. Creates a `policy-system` namespace for the webhook backends.
2. Generates a throwaway self-signed CA via the `cryptography` library so the `caBundle` schema is satisfied. The CA is never validated at runtime because the NetworkPolicy drops the connection before TLS.
3. Deploys four backend Deployments and Services in `policy-system`, each running `busybox:1.36 sleep infinity`. The backends never serve traffic. That's intentional.
4. Installs four real `MutatingWebhookConfiguration`s pointing at those backends. Each has `namespaceSelector` matching `hotel-reservation`, `timeoutSeconds: 10`, `failurePolicy: Ignore`. Names: `pod-resource-validator`, `audit-log-enforcer`, `image-policy-checker`, `tenant-quota-validator`.
5. Installs four decoy `MutatingWebhookConfiguration`s with ecosystem-realistic names: `cert-manager-webhook`, `istio-sidecar-injector`, `kyverno-resource-mutating-webhook-cfg`, `linkerd-proxy-injector-webhook-config`. They share the same `namespaceSelector`, so listing webhooks for the application namespace returns 8 hits, not 4. The decoys are kept inert by `objectSelector`s that require labels no application pod carries (for example `linkerd.io/inject: enabled`), or by `rules` targeting CRDs that aren't installed. Their backend Services point at non-existent endpoints in `policy-system`, but those services are never called because the selectors don't match. Decoys are deliberately visible but non-causal: a diagnostic distractor.
6. Creates the `default-deny-ingress` NetworkPolicy in `policy-system`, blocking all ingress. This is the actual fault.
7. Deletes the running `recommendation` pod in `hotel-reservation`. The ReplicaSet tries to recreate it, admission fails, the deployment goes to `0/1 ready`, the `recommendation` Service loses its endpoints.
8. `frontend` calls to `recommendation` degrade. The application impact is visible from the user side, not just a cluster-state discrepancy.

`recover_fault()` reverses the destructive steps in safe order: NetworkPolicy first, so subsequent kubectl calls aren't themselves admission-blocked.

I picked `recommendation` deliberately. It's a leaf service in Hotel Reservation, so the cascade is bounded. The rest of the application keeps running, but the failure is real and observable from the frontend side.

## Why diagnosis is hard

The error message gives nothing away. `context deadline exceeded` looks identical to scheduler timeouts under resource pressure, etcd contention, or any of a dozen other apiserver-side issues. It names no webhook and no NetworkPolicy.

Each webhook individually looks safe. Reading any single `kubectl describe mutatingwebhookconfiguration` shows `failurePolicy: Ignore`, which an operator reads as "if I time out, you'll skip me." The trap exists only in the sum across all four. Recognizing it requires enumerating the chain, adding the timeouts, and comparing to the apiserver's global deadline. None of that is hinted at by any single resource.

The cause is one network layer below the symptom. The visible signal is pod admission failure. The first investigative impulse is to look at the webhook backends, and the backends are `busybox sleep`, which *looks* like the cause: "the webhook server isn't listening." Even if the backends were real webhook servers, the NetworkPolicy in `policy-system` would still drop the request before TLS handshake. The agent has to follow the chain through this plausible-looking false bottom into another namespace's NetworkPolicy.

Decoys add diagnostic noise. The first webhook listing returns 8 hits, not 4. The decoy names are realistic enough (`cert-manager`, `istio`, `kyverno`, `linkerd`) that a competent operator would recognize them as the kind of cluster tooling that legitimately installs admission webhooks. The decoys' `objectSelector`s are visible, so the trail is followable. It just requires more investigation than reading the first row of `kubectl get`.

## The mitigation oracle

A generic "are pods Running?" check would be defeated by `kubectl delete mutatingwebhookconfiguration --all`. The custom oracle in `sregym/conductor/oracles/cumulative_admission_webhook_timeout_mitigation.py` has four properties evaluated in order. Evaluation short-circuits on the first failure.

**Property 1: spec is in an accepted fix shape.** Any one of these passes:

1. The `default-deny-ingress` NetworkPolicy is deleted.
2. The NetworkPolicy is patched to allow ingress from anywhere.
3. Cumulative `timeoutSeconds` across surviving real webhooks targeting `hotel-reservation` is at most 25 seconds (5-second safety margin under the 30-second global deadline).
4. At least one webhook's `namespaceSelector` no longer matches `hotel-reservation`.
5. At least one (but not all) of the four real webhook configurations is deleted.

If none of these holds, Property 1 returns False with an explicit reason that names the accepted shapes. The "policy plane must remain present" rejection (when all four real webhooks have been deleted) is the part that makes this oracle work. Real production operators can't delete compliance webhooks at will, and the oracle enforces that constraint the way production would. Same invariant style as `cronjob_sidecar_blocks_completion`.

**Property 2: workload intact.** `recommendation` deployment exists, `spec.replicas >= 1`, not paused, no `deletionTimestamp`. Rejects "fix" attempts that remove the affected workload.

**Property 3: pod is healthy.** `status.ready_replicas == spec.replicas`. The pod actually reached Ready, not just exists.

**Property 4: fresh probe pod admits successfully.** The oracle creates a real pod (`oracle-probe-<8-random-hex>`) in `hotel-reservation` via the kubernetes Python client and polls for Running with a 90-second deadline. Random suffix prevents pre-write attacks. Cleanup runs in a finally block via `contextlib.suppress(ApiException)`. If admission is still broken, this property fails regardless of what the spec looks like.

Verified rejected shortcuts:

| Shortcut | Caught by |
|---|---|
| Delete all 4 real webhooks | Property 1 ("policy plane must remain present") |
| Delete `recommendation` deployment | Property 2 |
| `kubectl scale deployment recommendation --replicas=0` | Property 2 |
| Patch all webhooks to `failurePolicy: Fail` | Property 4 (probe still fails) |
| Add a NetworkPolicy that "allows" something but doesn't unblock the apiserver | Property 4 |

### Validation lifecycle

`tests/integration/validate_problem.py` runs the canonical inject → oracle-must-fail → recover → oracle-must-pass contract. The problem passes 6/6 stages:

```
PROBLEM VALIDATION — cumulative_admission_webhook_timeout_hotel_reservation
================================================================
  ✅  Resolve problem in registry
  ✅  Deploy application
  ✅  Inject fault
  ✅  Oracle fails after fault injection
  ✅  Recover fault
  ✅  Oracle passes after recovery
Result: PASSED  (279s)
```

I also re-ran the validation after three back-to-back agent runs to confirm the cluster baseline survives agent-driven state churn. That run also passed in 293s.

### Three Claude Sonnet 4.5 runs

Each run is one invocation of `main.py --problem cumulative_admission_webhook_timeout_hotel_reservation --agent claudecode --model claude-sonnet-4-5 --judge-model gemini/gemini-2.5-flash`. The judge runs at temperature 0, so identical diagnosis text produces identical scores.

| Run | TTL | TTM | Acc | D1 | D2 | D3 | yes/9 | Diagnosis | Mitigation |
|---|---|---|---|---|---|---|---|---|---|
| A | 353s | 430s | 33 | 0.67 | 0.0 | 0.33 | 4 | FAIL | PASS (partial deletion) |
| B | 125s | 316s | 11 | 0.0 | 0.0 | 0.33 | 1 | FAIL | FAIL (delete-all rejected) |
| C | 146s | 477s | 100 | 1.0 | 1.0 | 1.0 | 9 | PASS | FAIL (delete-all rejected) |

Diagnosis pass requires composite ≥ 0.78 and no dimension scored 0 (SREGym §App-A threshold). Sonnet 4.5 did not pass both diagnosis and mitigation in any of three attempts.

### Run A

The agent listed all 8 MutatingWebhookConfigurations targeting `hotel-reservation`, looked at the decoy `objectSelector` fields without falsely blaming any decoy, drilled into only 2 of the 4 real webhooks (`pod-resource-validator`, `tenant-quota-validator`), and submitted:

> *"The recommendation service deployment is failing because the pod-resource-validator mutating webhook is not responding. The webhook is configured to intercept all pod creations in the hotel-reservation namespace with a 10-second timeout, but the webhook pod is running only a busybox sleep loop instead of an actual webhook server. This causes all recommendation pod creation attempts to timeout after 10 seconds, preventing the deployment from becoming available."*

The "10 seconds" claim is wrong. The actual failure is at the apiserver's ~30-second global deadline, not at any individual webhook's 10-second per-call timeout. The agent inspected the `policy-system` namespace 15 times during the run (for backend pod logs) but never looked at NetworkPolicies in it.

Per-question judge breakdown:

| Q | Ans | Evidence |
|---|---|---|
| D1-Q1 | No | Names a single webhook as the fault origin; ground truth says the failure originates apiserver-side from the aggregate deadline across a chain. |
| D1-Q2 | Yes | Distinguishes affected deployment from the webhook cause. |
| D1-Q3 | Yes | Identifies a webhook as not healthy. |
| D2-Q1 | No | Attributes failure to "busybox sleep loop", not NetworkPolicy or cumulative timeout. |
| D2-Q2 | No | Misses the apiserver's global 30-second deadline, the NetworkPolicy, and the `policy-system` namespace. |
| D2-Q3 | No | Attributes fault to application misbehavior, not network policy misconfiguration. |
| D3-Q1 | Yes | Doesn't blame healthy components. |
| D3-Q2 | No | Omits kube-apiserver, multiple webhooks, NetworkPolicy, frontend degradation. |
| D3-Q3 | No | "10 seconds" timing inconsistent with ground truth's ~30-second apiserver deadline. |

For mitigation the agent deleted the two webhooks it had been investigating. By coincidence, that left two real webhooks alive (policy plane intact) and dropped cumulative timeout to 2 × 10s = 20s, below the 25-second safety threshold. Property 1 accepted. The mitigation passed, but diagnosis was 4/9, so the joint outcome is still FAIL.

### Run B

The kind cluster had real load from prior runs: kind-worker at 100% of CPU limits, kind-worker3 at 125%, because the cluster runs `kube-system`, `openebs`, `observe`, `chaos-mesh`, `khaos`, and `hotel-reservation` simultaneously. The agent saw the CPU numbers, saw the same `context deadline exceeded` error in the ReplicaSet, and submitted:

> *"The recommendation service is unavailable because its pod cannot be scheduled due to insufficient CPU resources. The cluster nodes have reached their CPU limit capacity (kind-worker at 100%, kind-worker3 at 125%), preventing the recommendation deployment from creating a pod."*

The diagnosis is internally plausible. `context deadline exceeded` is genuinely ambiguous in isolation. It can come from scheduler timeouts under resource pressure or from the apiserver admission deadline. The agent picked the former without inspecting admission webhooks. The CPU pressure the agent observed comes from the kind cluster carrying days of accumulated work; a fresh cluster per run might produce different diagnoses, but persistent clusters are how SREGym normally evaluates, so this run reflects realistic conditions.

Per-question judge breakdown:

| Q | Ans | Evidence |
|---|---|---|
| D1-Q1 | No | Ground truth fault origin is NetworkPolicy/webhooks/kube-apiserver; diagnosis blames cluster nodes and CPU. |
| D1-Q2 | No | Attributes fault to insufficient CPU on cluster nodes, not the ground-truth origin. |
| D1-Q3 | No | Misidentifies CPU exhaustion on nodes. |
| D2-Q1 | No | Identifies CPU exhaustion as mechanism; ground truth is NetworkPolicy blocking webhooks. |
| D2-Q2 | No | Reports CPU limits (100%, 125%), unrelated to NetworkPolicy and webhook configuration. |
| D2-Q3 | No | Attributes fault to insufficient CPU, an incorrect and unrelated fault type. |
| D3-Q1 | No | Blames cluster nodes that are not contributors. |
| D3-Q2 | No | Misses kube-apiserver, admission webhooks, NetworkPolicy. |
| D3-Q3 | Yes | Correctly quotes the `context deadline exceeded` error string. |

For mitigation the agent first ran `kubectl patch deployment/recommendation` to reduce CPU limit from `1` to `500m`. That didn't help. The agent then deleted all four real webhook configurations. Property 1 rejected with the "policy plane must remain present" message.

### Run C

The agent traced the cross-layer dependency exactly. Final submission:

> *"The recommendation service deployment is failing to create pods due to admission webhook timeouts. The root cause is a default-deny-ingress NetworkPolicy in the policy-system namespace that blocks all ingress traffic, preventing the Kubernetes API server from reaching the mutating webhook services (audit-log-enforcer, image-policy-checker, pod-resource-validator, tenant-quota-validator). When the API server attempts to create pods in the hotel-reservation namespace, it tries to call these webhooks for validation but cannot reach them within the 10-second timeout, causing pod creation to fail with context deadline exceeded errors."*

All 9 checklist questions answered Yes. Composite 1.0. The agent named the affected deployment, the NetworkPolicy, the namespace boundary between `policy-system` and `hotel-reservation`, the cumulative-timeout mechanism, and the apiserver's role.

Per-question judge breakdown:

| Q | Ans | Evidence |
|---|---|---|
| D1-Q1 | Yes | Identifies `default-deny-ingress` NetworkPolicy in `policy-system`. |
| D1-Q2 | Yes | Names the affected deployment alongside the NetworkPolicy root cause. |
| D1-Q3 | Yes | NetworkPolicy in `policy-system` as the fault component. |
| D2-Q1 | Yes | NetworkPolicy blocks ingress, preventing kube-apiserver from reaching webhook services. |
| D2-Q2 | Yes | NetworkPolicy, 10-second per-webhook timeout, unreachable backends. |
| D2-Q3 | Yes | NetworkPolicy blocking all ingress as the fault type. |
| D3-Q1 | Yes | Names specific webhooks; doesn't blame inert ones. |
| D3-Q2 | Yes | Mentions `recommendation` deployment, NetworkPolicy, kube-apiserver, webhook services. |
| D3-Q3 | Yes | Symptoms (pod creation failing with `context deadline exceeded`) match the mechanism. |

For mitigation, despite having the right answer, the agent deleted all four real webhooks. Property 1 rejected with the same "policy plane must remain present" message. The destructive shortcut was mechanically simpler than opening the NetworkPolicy or narrowing namespaceSelectors, and the agent took it.

## Files

| Path | Purpose |
|---|---|
| `sregym/conductor/problems/cumulative_admission_webhook_timeout_hotel_reservation.py` | Problem class. `inject_fault`, `recover_fault`, helpers for CA generation, webhook config build, decoy management, NetworkPolicy management. |
| `sregym/conductor/oracles/cumulative_admission_webhook_timeout_mitigation.py` | Custom 4-property mitigation oracle. |
| `sregym/conductor/problems/registry.py` | One import, one dict entry. |

## Test plan

- [x] `validate_problem.py` passes 6/6 stages (279s)
- [x] `validate_problem.py` re-run after three back-to-back agent runs passes (293s)
- [x] Three Claude Sonnet 4.5 agent runs captured; per-question judge breakdowns are inline above under each run's section
- [x] Oracle rejects deleting all four real webhooks (verified Runs B, C)
- [x] Oracle accepts the legitimate partial-deletion shape (verified Run A)


Saadmrp1038 (changes requested):
@riyadomf I ran the fault on both kind and the real cluster. On the real cluster the fault sometimes does not work.

I did some testing to isolate the cause. The issue appears to be that Calico does not enforce ingress NetworkPolicy on traffic originating from the host network namespace when the destination pod is on the same node.

Tested with isolated webhook backends on specific nodes:

| Source | Destination | Time | NetworkPolicy enforced? |
|--------|-------------|------|------------------------|
| Regular pod on node0 | Pod on node0 (same node) | 5s (timeout) | ✅ Yes |
| Regular pod on node0 | Pod on node1 (cross node) | 5s (timeout) | ✅ Yes |
| **Host-network pod on node0** | **Pod on node0 (same node)** | **0s (instant)** | ❌ **No** |
| Host-network pod on node0 | Pod on node1 (cross node) | 5s (timeout) | ✅ Yes |

Also tested with single-webhook admission:

| Webhook backend location | Pod admission time | Behavior |
|---|---|---|
| Worker node (node1) | **13s** | NetworkPolicy blocks → 10s timeout → fail open |
| Control-plane node (node0) | **1s** | NetworkPolicy bypassed → instant connection refused → fail open |

The bypass requires BOTH: (1) source is host-network AND (2) target pod is on the same node. Regular pod-to-pod same-node traffic IS properly blocked. The kube-apiserver runs with `hostNetwork: true`.

It works on kind because the kind control-plane node has a `NoSchedule` taint, so backend pods are forced onto worker nodes.

---

Can you please check if this is the case or not. If this is the issue a simple anti-affinity to the backend deployments to exclude the control-plane node should do the trick.

riyadomf:
@Saadmrp1038 You're right. I was able to reproduce the bypass locally before making any changes.

I set up a namespace with a default-deny ingress NetworkPolicy, then pinned one busybox pod to the control-plane node and another to a worker node. A `hostNetwork: true` probe running on the control-plane showed the same behavior:

- Host-network probe → pod on same node (control-plane): ~0.06s, NetworkPolicy bypassed, instant connection refused.
- Host-network probe → pod on worker node: ~10s, NetworkPolicy enforced, connection timed out.

This matches your findings. The only thing that changes is the destination node. I also found that this is a known Calico behavior and aligns with [Calico issue #2632](https://github.com/projectcalico/calico/issues/2632).

For this fault, a backend landing on the control-plane node effectively removes one 10-second timeout from the chain. With one backend there, cumulative delay drops from ~40s to ~30s, which explains the flakiness. With two, the fault often won't trigger at all.

I've pushed a fix that adds node affinity to exclude nodes labeled `node-role.kubernetes.io/control-plane`, ensuring webhook backends are scheduled only on worker nodes regardless of cluster taints.

Verified locally:

- `validate_problem.py` passes all stages.
- After a fresh injection, all four backend pods are scheduled on worker nodes and none on the control-plane.

Thanks for testing this on a real cluster.

sregym/conductor/problems/cumulative_admission_webhook_timeout_hotel_reservation.py
```diff
@@ -0,0 +1,745 @@
...
+            else:
+                raise
+
+    def _generate_ca_bundle(self) -> str:
+        """Generate a throwaway self-signed CA. The certificate is never
+        actually validated at runtime because the NetworkPolicy drops
+        traffic before TLS handshake, but the ``caBundle`` field on the
+        webhook config requires a base64-encoded PEM-encoded cert."""
+        key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+        subject = issuer = x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, "sregym-webhook-ca")])
```
Saadmrp1038: nit: We should use a more realistic/generic name here and not leak the `sregym` label.

sregym/conductor/oracles/cumulative_admission_webhook_timeout_mitigation.py
```diff
@@ -0,0 +1,408 @@
...
+        backend_names = self.problem.WEBHOOK_BACKEND_NAMES
+        app_namespace = self.problem.namespace
+
+        # Check NetworkPolicy
+        np = self._get_network_policy(net_policy_name, policy_ns)
+        if np is None:
+            return True, "NetworkPolicy deleted (accepted fix shape)"
+        # Allow-all check: ingress contains an empty from selector list which means "from anywhere"
+        ingress = (np.spec and np.spec.ingress) or []
+        if any(rule.from_ is None or rule.from_ == [] or rule.from_ is None for rule in ingress):
```
Saadmrp1038: Duplicate condition. `rule.from_ is None` appears as both the 1st and 3rd condition in the `or` chain.

sregym/conductor/problems/cumulative_admission_webhook_timeout_hotel_reservation.py
```diff
@@ -0,0 +1,745 @@
...
+                self.networking_v1.replace_namespaced_network_policy(
+                    name=self.NETWORK_POLICY_NAME, namespace=self.POLICY_NAMESPACE, body=body
+                )
+                logger.info(
+                    f"Replaced existing NetworkPolicy '{self.NETWORK_POLICY_NAME}' in '{self.POLICY_NAMESPACE}'."
+                )
+            else:
+                raise
+
+    def _ensure_mutating_webhook(self, backend_name: str, ca_bundle_b64: str) -> None:
```
Saadmrp1038, line 1007: When `_ensure_mutating_webhook` gets a 409 Conflict (webhook already exists), it attempts `replace_mutating_webhook_configuration` without setting `metadata.resourceVersion`. This fails with 422 Unprocessable Entity.

Faced this when doing the following sequence:
`inject_fault` -> `recover_fault` -> `inject_fault`

Teardown fixes this so normal runs won't encounter it. But when trying to test without undeploying and redeploying the app, this becomes an issue.

The decoy code handles this correctly:

```python
existing = self.admissionregistration_v1.read_mutating_webhook_configuration(name=spec["name"])
body["metadata"]["resourceVersion"] = existing.metadata.resource_version
self.admissionregistration_v1.replace_mutating_webhook_configuration(name=spec["name"], body=body)
```

Saadmrp1038 (changes requested):
@riyadomf Working now!

Running claudecode with sonnet-4.6 I got the following results:

| Run | Diagnosis Score | Diagnosis Success | Mitigation Success | Diagnosis Summary | Mitigation Summary |
|-----|----------------|-------------------|-------------------|-------------------|-------------------|
| 1 | 100/100 | ✅ | ✅ | Named all 4 webhooks, NetworkPolicy, cumulative timeout mechanism, and cross-namespace dependency. Did not blame decoys. | Deleted the NetworkPolicy |
| 2 | 66/100 | ❌ | ✅ | Identified NetworkPolicy and webhooks correctly but blamed "fake busybox" backends as a contributing factor. Omitted global admission deadline, companion webhooks, and frontend degradation. | Deleted the NetworkPolicy |
| 3 | 77/100 | ✅ | ✅ | Identified NetworkPolicy and webhooks with cumulative 40s timeout. Blamed "fake busybox" backends and attributed failure to "replicaset controller's timeout" instead of apiserver admission deadline. | Deleted the NetworkPolicy |
| 4 | 100/100 | ✅ | ❌ | Perfect diagnosis — named all 4 webhooks, NetworkPolicy, cumulative timeout, cross-namespace dependency. | Deleted all 4 real webhooks (oracle rejected: "policy plane must remain present") |
| 5 | 89/100 | ✅ | ❌ | Identified NetworkPolicy and cumulative timeout but lumped decoys with real webhooks as equally causal ("7 fake mutating admission webhooks"). | Deleted all 7 webhooks (4 real + 3 decoys) — oracle rejected |

---

The diagnosis is not too hard. The agent consistently finds the NetworkPolicy and webhook chain. The issue is that in 2/5 runs the agent detects the backends are busybox (via `kubectl exec ... ss -tlnp` or checking pod images), concludes the webhooks are "fake" and shouldn't exist, and deletes them instead of fixing the NetworkPolicy.

The agent's reasoning in those runs: "there are two problems — NetworkPolicy blocks traffic AND webhooks are fake busybox. Deleting either fixes it." The agent chose to delete the webhooks since it viewed them as illegitimate. Technically correct (it would fix admission), but the oracle rejects it because the policy plane must remain present.

If agents can trivially detect the backends are busybox (not real webhook servers), it weakens the problem's realism. A real production incident would have actual webhook servers that are unreachable due to the NetworkPolicy, not dummy containers. The agent shouldn't be able to dismiss the policy plane as "fake" by inspecting the pod image.

---

Let me know what you think would be a good way to handle this.

sregym/conductor/problems/cumulative_admission_webhook_timeout_hotel_reservation.py
riyadomf: Fixed.

`_ensure_mutating_webhook` now follows the same create-or-replace pattern as the decoy helper: on a 409, it fetches the existing object, copies `metadata.resourceVersion`, and performs a replace.

I found the same issue in three other helpers and updated them as well:

* `_ensure_default_deny_network_policy`: fixed the 409 → replace path, which could previously fail with a 422 on re-injection.
* `_ensure_backend_deployment`: now updates existing Deployments instead of logging and skipping, so spec changes propagate correctly.
* `_ensure_backend_service`: now updates existing Services while preserving the immutable `clusterIP`.

riyadomf:
> @riyadomf Working now!
> 
> Running claudecode with sonnet-4.6 I got the following results:
> 
> Run	Diagnosis Score	Diagnosis Success	Mitigation Success	Diagnosis Summary	Mitigation Summary
> 1	100/100	✅	✅	Named all 4 webhooks, NetworkPolicy, cumulative timeout mechanism, and cross-namespace dependency. Did not blame decoys.	Deleted the NetworkPolicy
> 2	66/100	❌	✅	Identified NetworkPolicy and webhooks correctly but blamed "fake busybox" backends as a contributing factor. Omitted global admission deadline, companion webhooks, and frontend degradation.	Deleted the NetworkPolicy
> 3	77/100	✅	✅	Identified NetworkPolicy and webhooks with cumulative 40s timeout. Blamed "fake busybox" backends and attributed failure to "replicaset controller's timeout" instead of apiserver admission deadline.	Deleted the NetworkPolicy
> 4	100/100	✅	❌	Perfect diagnosis — named all 4 webhooks, NetworkPolicy, cumulative timeout, cross-namespace dependency.	Deleted all 4 real webhooks (oracle rejected: "policy plane must remain present")
> 5	89/100	✅	❌	Identified NetworkPolicy and cumulative timeout but lumped decoys with real webhooks as equally causal ("7 fake mutating admission webhooks").	Deleted all 7 webhooks (4 real + 3 decoys) — oracle rejected
> The diagnosis is not too hard. The agent consistently finds the NetworkPolicy and webhook chain. The issue is that in 2/5 runs the agent detects the backends are busybox (via `kubectl exec ... ss -tlnp` or checking pod images), concludes the webhooks are "fake" and shouldn't exist, and deletes them instead of fixing the NetworkPolicy.
> 
> The agent's reasoning in those runs: "there are two problems — NetworkPolicy blocks traffic AND webhooks are fake busybox. Deleting either fixes it." The agent chose to delete the webhooks since it viewed them as illegitimate. Technically correct (it would fix admission), but the oracle rejects it because the policy plane must remain present.
> 
> If agents can trivially detect the backends are busybox (not real webhook servers), it weakens the problem's realism. A real production incident would have actual webhook servers that are unreachable due to the NetworkPolicy, not dummy containers. The agent shouldn't be able to dismiss the policy plane as "fake" by inspecting the pod image.
> 
> Let me know what you think would be a good way to handle this.

@Saadmrp1038 Thanks for the testing.

The busybox backends made it too easy for the agent to dismiss that the policy plane itself was illegitimate. I've updated the design so the backends behave like real admission services: each webhook now runs an HTTPS server that handles AdmissionReview requests, evaluates them against a compliance-style ruleset, and returns proper AdmissionReview responses. Under normal conditions, the webhooks are fully functional and admit the Hotel Reservation workloads.

I also varied the per-webhook timeouts (12s, 8s, 11s, 9s) instead of using four identical 10-second timeouts, so the cumulative timeout is less obvious at a glance.

My goal was to make the policy plane something the agent has to investigate and rule out, rather than something it can immediately dismiss as fake by checking the pod image or process list. The underlying root cause remains the same: the NetworkPolicy prevents the apiserver from reaching otherwise healthy webhook backends. 

I've also incorporated the other fixes you flagged in your inline comments.

After the changes, `validate_problem.py` passes all stages end-to-end.

Saadmrp1038:
@riyadomf The fix worked. The agent is not calling it fake anymore. But it seems the fault is quite easy for the agent to solve.

*claudecode sonnet-4.6 runs*
| Run | Diagnosis Score | Diagnosis Success | Mitigation Success | Diagnosis Summary | Mitigation Summary |
|-----|----------------|-------------------|-------------------|-------------------|-------------------|
| 1 | 100/100 | ✅ | ✅ | Named all 4 webhooks, NetworkPolicy, cumulative timeout, cross-namespace dependency. | Deleted the NetworkPolicy |
| 2 | 89/100 | ✅ | ✅ | Identified NetworkPolicy and webhooks correctly. D3 dinged for calling it "malicious." | Deleted the NetworkPolicy |
| 3 | 100/100 | ✅ | ✅ | Named NetworkPolicy, webhooks, cumulative timeout, cross-namespace dependency. | Deleted the NetworkPolicy |
| 4 | 66/100 | ❌ | ✅ | Named webhooks but said they are "deadlocked." Missed NetworkPolicy as root cause initially. | Patched ConfigMap, eventually fixed |
| 5 | 89/100 | ✅ | ✅ | Correctly identified NetworkPolicy, cumulative timeout exceeding deadline. | Deleted the NetworkPolicy |
| 6 | 100/100 | ✅ | ✅ | Named NetworkPolicy, webhooks, cumulative timeout, cross-namespace dependency. | Deleted the NetworkPolicy |
| 7 | 100/100 | ✅ | ✅ | Named NetworkPolicy, webhooks, cumulative timeout, cross-namespace dependency. | Deleted the NetworkPolicy |

The implementation is solid already. Maybe we can do some minor fiddling to make it a bit harder. I checked the incident link [kubernetes#128162](https://github.com/kubernetes/kubernetes/issues/128162). It seems the NetworkPolicy there was more sophisticated and not a simple deny-all ingress.

> This occurred with Datadog's admission controller, when a default deny policy was applied and the network policy allowing ingress to Datadog's admission controller was missing.

So we could have a NetworkPolicy with partial allows but missing the allow for the apiserver CIDR. We could even have multiple NetworkPolicies. That would reflect the real incident more faithfully and force the agent to reason about what's *absent* rather than spotting an obvious deny-all.

You can explore other ways also.

---

We are not looking for a very hard problem but at least there should be some amount of difficulty where the agent gets confused and fails sometimes (more than the 1/7 diagnosis failure and 0/7 mitigation failure we see here). Problems like these help us see how an agent navigates ambiguity and expose its strengths and weaknesses. Even a problem where the agent solves it but takes a lot of time to reason about it is also interesting.

riyadomf:
@Saadmrp1038, Thanks for the suggestion. I reworked the network layer.

## What changed

- The fault is now a **missing apiserver allow**, not a blanket deny-all. `policy-system` carries a realistic baseline: `default-deny-ingress` plus several targeted allows (metrics from `observe`, intra-namespace, and a `kube-system` control-plane allow). None permits the apiserver→webhook path, so the webhooks still time out and the cumulative wait trips the ~30s deadline.
- `allow-control-plane` is a deliberate **near-miss**: it looks like it permits the control plane, but the apiserver is hostNetwork, so its source isn't a `kube-system` pod and the selector doesn't match. The real fix has to allow the webhook port broadly, not a scoped source.
- Since NetworkPolicy is additive, deleting the deny-all alone no longer works. The other allows keep the backends isolated. The mitigation oracle is restructured to invariants + a runtime probe: backends must stay isolated, "delete all NetworkPolicies / all webhooks" is rejected, and a fresh probe pod is the ground truth for admission recovery.

## Evaluation (claudecode, sonnet-4.5, N=3)

| Run | Diagnosis | Mitigation |
| --- | --- | --- |
| 1 | 100% ✅ (reasoned through the hostNetwork near-miss) | ✅ but only via removing 2 webhooks after failing to add a working allow |
| 2 | 22% ❌ (blamed the decoys) | ❌ deleted all webhooks → rejected |
| 3 | 11% ❌ (missed the NetworkPolicy) | ❌ deleted all webhooks → rejected |

Diagnosis 1/3, mitigation 1/3. 
The near-miss + missing-allow framing confuses agents now, and even the one success took ~18 min and couldn't pin down the right allow.

Saadmrp1038:
@riyadomf Thanks for addressing all the comments! The runs look a lot more interesting now. With claude-sonnet-4.6 the agent can still solve it but it's not as easy as before. It takes quite a lot of time for the agent to diagnose and mitigate the fault, and sometimes times out.

| Run | Diagnosis Score | Diagnosis Success | Mitigation Success | TTL | TTM | Diagnosis Summary | Mitigation Summary |
|-----|----------------|-------------------|-------------------|-----|-----|-------------------|-------------------|
| 1 | 100/100 | ✅ | ✅ | 868s | 1573s | Identified near-miss `allow-control-plane` and hostNetwork reason it doesn't match apiserver. | Patched allow-control-plane, deleted default-deny, then reduced webhook timeouts |
| 2 | 89/100 | ✅ | ❌ (timed out) | 1374s | — | Correctly identified NetworkPolicy issue and webhook chain. | Tried patching allow-control-plane but couldn't get the right allow before timeout |
| 3 | 78/100 | ✅ | ✅ | 998s | 1248s | Lumped all 8 webhooks as causal ("cumulative 60s"). | Patched allow-control-plane with `0.0.0.0/0` on port 443 (after narrow ipBlock failed) |
| 4 | 89/100 | ✅ | ❌ (timed out) | 858s | — | Identified hostNetwork as reason allow-control-plane doesn't work. | Tried ipBlock for node IP + deleted default-deny, still couldn't get admission working |
| 5 | 89/100 | ✅ | ✅ | 986s | 1324s | Identified misconfigured allow-control-plane. | Patched allow-control-plane + reduced webhook timeouts to 3s each |

Diagnosis 5/5, mitigation 3/5. Agents don't get the fix right on the first attempt. They typically try patching `allow-control-plane` with a narrow ipBlock first (which doesn't work), then succeed on later attempts by either broadening to `0.0.0.0/0` or reducing webhook timeouts.

---

Congrats on your first PR merged here @riyadomf 🎉

Saadmrp1038 (approved):

PR #888, opened 2026-06-13
Add runAsNonRoot security context fault for hotel-reservation

talha-asgher:
## Real-World Failure Story

A security hardening initiative rolls out new pod security policies across a cluster.
An engineer patches a deployment to add `securityContext: {runAsNonRoot: true}` — but
the container image (DeathStarBench's Go-based services) has no USER directive and
defaults to UID 0. Kubernetes rejects the container *before it ever starts*, with the
error: "container has runAsNonRoot and image will run as root". The pod lands in
`CreateContainerConfigError` permanently. Engineers initially suspect a crashing
application, but no logs exist because the container never ran.

Real-world references:
- https://github.com/Budibase/budibase/issues/4796
- https://github.com/kubernetes-sigs/node-feature-discovery-operator/issues/197
- https://github.com/rook/rook/issues/13135
- https://github.com/jaegertracing/helm-charts/issues/105
- https://github.com/keycloak/keycloak-operator/issues/458
- https://devblogs.microsoft.com/dotnet/running-nonroot-kubernetes-with-dotnet/

## Simulation

The fault targets the `geo` deployment in the `hotel-reservation` namespace, chosen
because it is a lightweight stateless service whose pods restart cleanly.

**Fault injection:** A strategic merge patch adds `securityContext.runAsNonRoot: true`
to the first container. Since the DeathStarBench image has no USER directive, it
defaults to UID 0, and Kubernetes rejects it at admission.

**Fault recovery:** A second patch nulls out `securityContext`, restoring the
deployment to its pre-fault state and allowing pods to return to Running.

**Oracles:**
- `LLMAsAJudgeOracle` evaluates the diagnosis against a structured root cause
  (component, namespace, description of the admission rejection).
- Standard `MitigationOracle` verifies all pods in the namespace reach Running state.
  Both mitigations are accepted — removing `securityContext` entirely, or adding
  `runAsUser: <non-zero UID>` — since the oracle only checks pod state.

## Problem Runtime Behavior

After fault injection:
- New `geo` pods enter `CreateContainerConfigError` and never progress to Running.
- `kubectl describe pod <geo-pod> -n hotel-reservation` shows the admission error in
  the Events section: *"container has runAsNonRoot and image will run as root"*.
- No application logs are produced — the container never ran.
- Requests routed through the `geo` service fail.

This makes the fault deceptively tricky: an engineer expecting a crashing application
will find no logs to inspect and must pivot to reading pod events instead.

Mitigation is accepted once all pods in the `hotel-reservation` namespace are Running.

## Agent Behavior (Stratus)

Stratus was run against this problem (1 attempt):

| Metric | Value |
|---|---|
| Diagnosis judgment | ✅ Pass |
| Diagnosis accuracy | 89% |
| Mitigation success | ✅ Pass |
| Time to localize (TTL) | ~34 seconds |
| Time to mitigate (TTM) | ~226 seconds |

**Diagnosis breakdown:**
- Fault Localization (D1): 1.00 — correctly identified the `geo` pod as the fault origin.
- Fault Characterization (D2): 1.00 — identified the `runAsNonRoot` / UID 0 conflict and `CreateContainerConfigError`.
- Scope Precision (D3): 0.67 — named the pod but did not explicitly reference `deployment/geo` as the originating component.

Stratus submitted: *"The geo pod is failing to start due to a CreateContainerConfigError.
The root cause is a security context conflict: the container is configured to run as a
non-root user (runAsNonRoot), but the container image itself attempts to run as the root
user. This mismatch prevents the container from being created successfully."*

The agent arrived at the correct fix (removing the security context) without attempting
any illegitimate workarounds.

/validate-problem run_as_non_root_hotel_reservation


Saadmrp1038:
@talha-asgher Thanks for the effort! The implementation is clean and works correctly.

However, this fault is too easy for our benchmark. The error message in pod events literally tells the agent what's wrong and what to fix: "container has runAsNonRoot and image will run as root". There's really no challenge. Just a `kubectl get events` or `kubectl describe pod` reveals the fault to the agent and also what to fix.

We ran Claude Code against this fault (3 runs):

| Run | Diagnosis | Accuracy | Mitigation | TTL | TTM |
|-----|-----------|----------|------------|-----|-----|
| 1 | Pass | 89% | Pass | 108s | 279s |
| 2 | Pass | 100% | Pass | 104s | 257s |
| 3 | Pass | 100% | Pass | 87s | 270s |

3/3 diagnosis and 3/3 mitigation (also very fast TTL and TTM). We need faults of sufficient difficulty that can challenge the agent enough to reveal its strengths and weaknesses. Otherwise that wouldn't really be a good benchmark problem.

Closing this PR for now. Feel free to open a new one or reopen this if you think this can be tweaked to make harder.

---

One thing, the description has odd newlines that look copy-pasted straight from Claude. Also, in `registry.py` only the import for `RunAsNonRootFault` exists the fault itself was not registered. Please be mindful of these stuff.

PR #891, opened 2026-06-15
Add psa_restricted_blocks_recreation problem

### LINKED ISSUE #887: [Problem RFC] Pod Security Admission "restricted" label silently blocks pod recreation
## Summary
 
A namespace gets the Pod Security Admission (PSA) `enforce: restricted` label as part of a security-hardening rollout. Nothing breaks immediately — PSA only gates **admission**, so already-running pods keep running. But the moment a pod is deleted (node drain, eviction, rollout, crash), the ReplicaSet's recreate attempt is **rejected by the kube-apiserver itself** with `violates PodSecurity "restricted:latest"`, because the app's pods run as root / don't set a seccomp profile / don't drop capabilities. The deployment quietly goes under-replicated, and stays that way.
 
The trap: there is no webhook to find, no crash loop, no resource pressure, and no image problem. The Deployment spec, Service, and resources are all valid. The rejection happens at the **namespace policy layer** — a label on the Namespace object — which most diagnostic flows never inspect. The error only surfaces in ReplicaSet events, not in Deployment or Pod status.
 
## Real-World Failure Story
 
Pod Security Admission replaced the deprecated PodSecurityPolicy and went GA in Kubernetes 1.25. The [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/) define three profiles (`privileged`, `baseline`, `restricted`); the `restricted` profile is the strictest and requires pods to explicitly set `runAsNonRoot: true`, drop all capabilities, and set a `seccompProfile`, among others.
 
The classic incident pattern is well documented as a failure *class* (rather than a single named company postmortem):
 
- A platform team enables `enforce: restricted` on a namespace to satisfy a compliance requirement.
- Existing workloads were built before these constraints — they run as root, omit a seccomp profile, or request added capabilities. They violate `restricted` but were never rejected, because **PSA only checks pods at admission time, not pods already running.**
- The violation stays dormant until something forces a pod to be recreated — a node drain, an autoscaler scale-up, a rollout, or a crash. Then the new pod is denied and the service silently degrades.
- A particularly subtle variant: an **injected sidecar** (service mesh, logging agent) adds a `securityContext` that violates the policy, so even a "compliant" app deployment fails admission once the sidecar is merged in.
 
This is documented across vendor and community guidance as a real operational footgun, with dedicated incident-response runbooks:
- [Kubernetes — Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) (enforce/audit/warn modes; enforce rejects violating pods at admission).
- [OneUptime — Troubleshooting PSA rejections after migration](https://oneuptime.com/blog/post/2026-02-09-pod-security-admission-rejections/view) ("a critical service fails to roll out after a policy change because production pods were denied").
- The standard prevention advice — *never go straight to `enforce: restricted`; stage with `audit`/`warn` first* — exists precisely because teams keep hitting this.
 
> Honesty note for reviewers: unlike the webhook-timeout RFC (which maps to a specific filed incident, kubernetes/kubernetes#128162), this is grounded in a documented failure **class** with published runbooks rather than a single named company RCA. The mechanism, symptoms, and fix are all real and reproducible; the "famous postmortem" is what's missing.
 
**Why this is a distinct failure class** from the admission problems already in SREGym:
- `admission_webhook_outage`, `admission_webhook_tls_mismatch`, `mutating_webhook_resource_limits` are all **webhook** problems — the agent's diagnostic path is "enumerate webhook configs, find the broken/Fail one."
- PSA is **built into the kube-apiserver** — there is no `ValidatingWebhookConfiguration` or `MutatingWebhookConfiguration` to find. The policy lives as a *label on the Namespace*. Enumerating webhooks returns nothing; the agent has to look at namespace metadata, a genuinely different place. No existing problem touches PSA or the Pod Security Standards.
 
## Proposed Simulation
 
**Target application:** Hotel Reservation (same single-namespace scoping as the existing admission problems; target a single-replica deployment such as `recommendation`).
 
**Fault injection:**
1. Add the PSA enforce label to the app namespace:
   `pod-security.kubernetes.io/enforce: restricted` (optionally pin `enforce-version`).
2. Confirm the target deployment's pods violate `restricted` (the demo images run as root / lack `runAsNonRoot` / lack a seccomp profile — true for the Hotel Reservation services).
3. Delete one pod of the target deployment. The ReplicaSet's recreate attempt is rejected at admission; the deployment stays under-replicated.
 
(Existing running pods of *other* services are unaffected — PSA doesn't evict them — which reinforces the "everything looks fine except this one deployment" symptom.)
 
**Expected symptoms:**
- The target deployment is under-replicated; the missing pod is never created.
- ReplicaSet events show `Error creating: pods "<name>" is forbidden: violates PodSecurity "restricted:latest": <specific violations>` (e.g. `allowPrivilegeEscalation != false`, `runAsNonRoot != true`, `seccompProfile`).
- No webhook configurations exist to explain it — enumerating `Validating/MutatingWebhookConfiguration` returns nothing relevant.
- Deployment spec, image, Service, and resources are all valid.
 
**Compatibility:** Works on kind — PSA is a built-in admission plugin enabled by default since 1.25 (kind runs ≥1.25). Pure Kubernetes machinery, no host access, no control-plane saturation. Deterministic: the label either is or isn't present, and the violation is structural.
 
## Expected Mitigation
 
Any of:
1. Remove the `pod-security.kubernetes.io/enforce: restricted` label from the namespace (revert the hardening).
2. Relax it to a profile the workload satisfies (e.g. `baseline`), or set `enforce: privileged` for that namespace.
3. (Hardest / most production-correct) Make the workload compliant — add `securityContext` with `runAsNonRoot`, `allowPrivilegeEscalation: false`, dropped capabilities, and a `seccompProfile` — so it passes `restricted`.
 
Once admission stops rejecting the pod, the existing ReplicaSet recreates it with no further changes. The lesson: PSA enforcement is silent until a pod is recreated, and `restricted` rejects workloads that were never built for it.
 
## Oracle Design
 
**Diagnosis oracle:** `LLMAsAJudgeOracle` — validates the agent identified the namespace PSA `restricted` enforcement (not a webhook, not the app) as the root cause.
 
**Mitigation oracle:** `DeploymentReadinessOracle` — passes only when the target deployment's `ready_replicas == desired` and all its pods are Running/Ready. This is the same oracle the existing webhook-outage problem uses: it reports failure while admission is blocked (the recreated pod is absent) and success once the workload can be admitted again. It accepts all three mitigation paths and resists reward-hacking (deleting the deployment / scaling to zero leaves it not-ready).
 
## Implementation Notes
 
- Follows the `admission_webhook_outage.py` lifecycle: in `inject_fault()` apply the namespace label and delete one target pod to surface the blocked recreate; in `recover_fault()` remove the label.
- Apply the label via `client.CoreV1Api().patch_namespace(...)` (or `kubectl label namespace ...`), capturing any prior PSA labels so recovery restores the exact original state.
- Reuse the existing `DeploymentReadinessOracle` and `LLMAsAJudgeOracle` — no new oracle needed.
- Files: `sregym/conductor/problems/psa_restricted_blocks_recreation.py` (new), one import + one registry entry in `sregym/conductor/problems/registry.py`.
- Verify on kind that the target Hotel Reservation deployment's pods actually violate `restricted` before wiring the oracle (they should, since the images run as root without a seccomp profile). If any chosen service happens to already be compliant, pick one that isn't.
 
## References
 
- [Kubernetes — Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/)
- [Kubernetes — Pod Security Standards (baseline / restricted)](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
- [OneUptime — Troubleshooting PSA rejections after migration](https://oneuptime.com/blog/post/2026-02-09-pod-security-admission-rejections/view)

TamimEhsan:
## Summary

  Adds `psa_restricted_blocks_recreation_hotel_reservation`, a new Hotel Reservation benchmark problem.
- Fixes #887.

Labelling the `hotel-reservation` namespace `pod-security.kubernetes.io/enforce=restricted` leaves already-running pods untouched (PSA only gates admission), but the kube-apiserver rejects any **recreated** pod because the demo pods declare no `securityContext` and so violate the restricted Pod Security Standard on four counts: `allowPrivilegeEscalation != false`, unrestricted capabilities, `runAsNonRoot != true`, and missing `seccompProfile`. Deleting a pod therefore leaves the `recommendation` deployment stuck under-replicated, with a `violates PodSecurity "restricted:latest"` event on the  ReplicaSet.

The failure is deterministic by construction: PSA evaluates the pod **spec fields**, not runtime behavior, so an unset `runAsNonRoot` is itself a violation regardless of the image's actual UID.

## Why it's distinct from the existing admission problems

`admission_webhook_outage`, `admission_webhook_tls_mismatch`, and `mutating_webhook_resource_limits` are all **webhook** problems — the diagnostic path is "enumerate webhook configs, find the broken one." Here there is no

`Validating`/`MutatingWebhookConfiguration` to find: the policy is enforced by the apiserver itself via a **namespace  label**, so the agent has to inspect namespace metadata instead. No existing problem touches Pod Security Admission / Pod Security Standards.

## Real-world basis

Modeled on the well-documented PSA hardening footgun: teams enabling `enforce: restricted` (PSA replaced PodSecurityPolicy, GA in 1.25) without staging via `audit`/`warn` first break workloads the moment a pod is recreated (node drain, eviction, rollout). See [Kubernetes — Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) and the RFC in #887 for references.

  ## Code changes

- `sregym/conductor/problems/psa_restricted_blocks_recreation.py` — problem class. Mirrors the `admission_webhook_outage` lifecycle (label namespace → delete one pod of a single-replica deployment → blocked recreate). Captures prior PSA labels for clean restore. Reuses `DeploymentReadinessOracle` and `LLMAsAJudgeOracle` — no new oracle.
- `sregym/conductor/problems/registry.py` — import + registry entry.
- `Problem List.md`, `tests/e2e-testing-scripts/registry.txt` — catalog entries.

  /validate-problem psa_restricted_blocks_recreation_hotel_reservation

## Test plan

- [x] Local validation passes end-to-end (deploy → inject → oracle **fails** at `0/1` → recover → oracle **passes** at  `1/1`), ~164s.
- [x] Verified on kind manually: PSA `restricted` rejects `recommendation` recreation with all four violation messages on the ReplicaSet; removing the label recreates the pod (`1/1`).
- [x] Oracle uses `DeploymentReadinessOracle` (checks `ready_replicas == desired` + namespace pod health), so it resists reward-hacking via delete / scale-to-zero.
- [x] Recovery is correct; note the `recommendation` deployment can take ~60–90s to return to `1/1` after the label is removed, due to the ReplicaSet's exponential `FailedCreate` backoff — well within the validation harness's 600s recovery  window.


Saadmrp1038 (approved):

Saadmrp1038:
@TamimEhsan The implementation is great! Tested end-to-end everything is working correctly.

Saadmrp1038:
@TamimEhsan Did you join our Slack group? If not, could you please do so?
Invite link: https://join.slack.com/t/sregym/shared_invite/zt-3z4ce764m-XRProMeLK468GZ2HDNrfUw

TamimEhsan:
@Saadmrp1038 Thanks! Joined.

PR #892, opened 2026-06-16
Automatically constructing problems from bug reports of large open-sourced systems

yimingsu01:
This PR serves as the first step towards automatic problem construction in SREGym. Due to the nature of SREGym problems, fully automatic construction (i.e., in SWEBench style) is difficult if not impossible. This PR takes the approach of coding agent skills.

I added `/sregym-db-bug-reproduction` and `/sregym-runtime-validation` skills. `/sregym-db-bug-reproduction` takes in a issue description url and attempts to reproduce it on SREGym. To use the skill, you need to make sure the target system is already ported to the benchmark (this PR also includes porting Cassandra to SREGym). The bug reproduction process is:
1. retrieve a bug report by a given url
2. find the buggy commit of the system
3. build the system under buggy commit and deploy by loading the image to the cluster
4. deploy a pod that consumes the system and manifests the bug into a failure.

In this PR, the reproducer is a simple deployment that runs the manifesting CQL query on Cassandra. If the query fails, then the pod goes into an abormal unready state which the agent will see on the cluster. We can incorporate other types of reproducer later.

The `/sregym-runtime-validation` skill is then used to validate the automatically constructed problems by deploying it on a local kind cluster and observe if the failure manifested on the cluster.


Saadmrp1038:
@yimingsu01 
+31,110/-45 :)

Would love a description of how this works!
Also, how do we pick the worthwhile problems. Manual filtering?

yimingsu01:
@Saadmrp1038 Not fully ready yet, but most of the lines come from the automatically constructed Cassandra problems.

yimingsu01:
This pipeline also doesn't do any filtering. It assumes that you have some URLs of bug reports that you want to replicate on the benchmark. It also assumes that you have already ported the target system on SREGym before automatically constructing problems from it.

Issue #826, opened 2026-05-30
[Problem RFC] Node Clock Drift Silently, Breaks TLS While Pods Stay Running


haani-maybe:
Hello! I'm working on the evaluation task and would like to propose a new SREGym benchmark problem based on node clock drift causing TLS handshake failures (reference links attached)

Advancing a worker node's system clock significantly forward causes all pods on that node to fail outbound TLS connections with `certificate has expired or is not yet valid`; this is not because any certificate is actually bad, but because the skewed clock makes even a valid certificate APPEAR expired from that node's perspective. 
The issue here is that the certificates are completely healthy so an agent has to keep digging to find the fault. All pods stay `Running` throughout so hypothetically the main signal is just the x509 errors in application logs that look identical to a cert misconfiguration and that makes it harder to diagnose the root cause. 

Real-world examples this is heavily inspired by are linked here as sources:
[ hashicorp/vault#11150](https://github.com/hashicorp/vault/issues/11150); (this is the main match since NTP stopped running here)
[ D2iQ time drift support article](https://support.d2iq.com/hc/en-us/articles/14802884090772); documents a Kubernetes cluster where etcd began rejecting connections with the same x509 error while `kubectl get pods` appeared fine in terms of result
[Cloudflare leap second](https://blog.cloudflare.com/how-and-why-the-leap-second-affected-cloudflare-dns/); not exact issue but interesting comparison case where NTP correctly stepping the clock backward by one second, mainly used as inspiration to demonstrates small clock drifts cascading
[ kubernetes/kubernetes#76714](https://github.com/kubernetes/kubernetes/issues/76714) ; example production of the error by advancing a system clock a display of the fact that it occurs naturally when cloud VM clocks drift
(The existing problems/ PRs handle genuine expiry or config mismatches, and another interesting match was Problem ID `cpu_clocksource_failure` listed in the Problem List which targets hardware TSC/RTC instability instead, whereas here the clock jumps forward via NTP shutdown and the main visible symptom is TLS validation failure)

For implementing this into SREGym, i'll utilize `hotel_reservation` (used in the quiz too) since it supports TLS natively and the fault has a real TLS surface to break without any additional app wiring needed. The problem would hypothetically inject the fault by targeting one worker node whose init container advances the system clock and disables NTP; recovery would then restore the sync. 

[...]

PR #894, opened 2026-06-18
Add node_clock_drift Problem 

haani-maybe:
<hr>
<h1>Node Clock Drift Causing TLS Handshake Failures on Hotel Reservation</h1>
<p>This PR adds a new SREGym problem, <code>node_clock_drift_hotel_reservation</code>, as part of my evaluation task for UIUC++ SRSE.</p>
<p>It models a worker node's system clock drifting significantly forward after its time-sync service is disabled. All pods on that node fail outbound TLS connections with <code>certificate has expired or is not yet valid</code>, even though the certificate itself is completely healthy. The interesting part for an agent is that <code>kubectl get pods</code> looks completely normal (Running/Ready), and the certificate, inspected on its own, is valid; the broken state only becomes visible once the agent connects the node's clock to the TLS validation failures.</p>
<blockquote>
<p><strong>NOTE for test environment:</strong> this problem can't be simulated on <code>kind</code>. Containers on the same Docker host share the host kernel's clock; drifting one "node" (container) would drift the system clock for all of them. A 3 node CloudLab cluster running Ubuntu 22.04 was used instead.</p>
</blockquote>
<p>Related RFC: #826</p>
<p><code>/validate-problem node_clock_drift_hotel_reservation</code></p>
<h2>1. Real-world failure story</h2>
<p>TLS validation checks a certificate's <code>notBefore</code>/<code>notAfter</code> window against the local system clock, not against any external source of truth. If a node's clock is wrong, even a perfectly valid certificate appears expired or not-yet-valid from that node's perspective.</p>
<p>This pattern recurs across unrelated stacks once time-sync stops running: <code>kubernetes/kubernetes#76714</code> (kubelet TLS failures from clock skew), <code>hashicorp/vault#11150</code> (Vault cert renewal failing the same way), and a D2iQ support article documenting an etcd cluster rejecting connections with the identical x509 error while <code>kubectl get pods</code> reported a fully healthy cluster. The Cloudflare leap-second postmortem was additional inspiration for how small, time-sync-driven clock steps can cascade into TLS failures cluster-wide.</p>
<p>References:</p>
<ul>
<li><code>kubernetes/kubernetes#76714</code>: https://github.com/kubernetes/kubernetes/issues/76714</li>
<li><code>hashicorp/vault#11150</code>: https://github.com/hashicorp/vault/issues/11150</li>
<li>D2iQ time drift support article: https://support.d2iq.com/hc/en-us/articles/14802884090772</li>
<li>Cloudflare leap second postmortem: https://blog.cloudflare.com/how-and-why-the-leap-second-affected-cloudflare-dns/</li>
</ul>
<h2>2. How the failure is simulated on SREGym</h2>
<p>The problem targets the Hotel Reservation app in the <code>hotel-reservation</code> namespace, specifically the <code>frontend</code> deployment and whichever worker node it's scheduled on.</p>
<p><code>hotel_reservation</code> doesn't have a native TLS surface (services talk to each other over plain HTTP, no ingress controller deployed), so fault injection first builds a minimal one to break:</p>
<ol>
<li>Generates a self-signed certificate with a <strong>1-day validity window</strong>, short enough to guarantee it's expired well within the 30-day clock-drift window used later.</li>
<li>Stores it as a <code>kubernetes.io/tls</code> Secret (<code>hotel-frontend-tls</code>).</li>
<li>Stores the same cert as a plain ConfigMap (<code>hotel-frontend-ca</code>), mounted into a sidecar as the CA bundle.</li>
<li>Patches the <code>frontend</code> deployment to add a <code>tls-health-check</code> sidecar that runs <code>openssl verify</code> against the mounted cert every 30 seconds — simulating a service performing routine outbound TLS validation. This is the only way to surface a TLS-shaped log signal, since the app itself has no real TLS surface to break. <code>apt-get install openssl</code> runs once at sidecar startup, before the clock is touched, so it isn't blocked by the drift that follows.</li>
<li>Waits for the sidecar rollout to fully land (new pod <code>Running</code>, all containers ready, sidecar present) before selecting a target node.</li>
<li>Selects the node currently running the frontend pod that has the sidecar.</li>
<li>Deploys a privileged pod (<code>hostPID</code>/<code>hostNetwork</code>/<code>hostIPC</code>) that uses <code>nsenter</code> to <strong>dynamically discover</strong> whichever time-sync service is currently active (queries <code>systemctl list-units --state=running</code> for any unit matching <code>ntp|chrony|timesync</code>), stops and masks it, then advances the system clock 30 days forward via <code>date -s</code>.</li>
</ol>
<p>The injector pod stays alive for the duration of the fault so the clock change persists. Recovery deletes all clock-drift injector pods; it does <strong>not</strong> automatically restore the time-sync service or roll back the clock, since reversing the fault yourself isn't a valid mitigation strategy.</p>
<h3>Bugs found and fixed during testing</h3>
<p><strong>Rollout race condition.</strong> The deployment patch that adds the sidecar triggers a rolling update. During testing, the old frontend pod (no sidecar) could still be <code>Running</code> on its original node at the exact moment node selection ran, while the new pod (with sidecar) was being scheduled to a <em>different</em> node, causing the clock drift to target a node the sidecar never ran on. Fixed by adding <code>_wait_for_sidecar_rollout()</code>, which polls until a frontend pod is <code>Running</code> with the sidecar container present and ready, and tightening node selection to only match pods that actually have the <code>tls-health-check</code> container.</p>
<p><strong>Hardcoded time-sync service name.</strong> The original injector tried to mask <code>systemd-timesyncd</code>, <code>chrony</code>, and <code>ntpd</code> by fixed name. On the actual CloudLab test cluster, none of those three matched; the real service was running as legacy <code>ntpd</code> under the systemd unit name <code>ntp.service</code> (no "d"). This silently failed to disable time-sync at all, since <code>systemctl stop ntpd</code> doesn't match a unit literally named <code>ntp</code>. Fixed by querying <code>systemctl list-units</code> for any <em>currently running</em> service matching <code>ntp|chrony|timesync</code> and acting on whatever is actually found, rather than guessing names. This makes the fault portable across environments without needing to maintain an exhaustive daemon name list.</p>
<h2>3. Problem runtime behavior and mitigation oracle</h2>
<p>After injection, pod health looks normal:</p>
<pre><code class="language-bash">kubectl get pods -n hotel-reservation
</code></pre>
<p><code>frontend</code> shows <code>2/2 Running</code>, no restarts. The diagnostic signal isn't in pod status — it's in the sidecar's behavior and the node's own clock:</p>
<pre><code class="language-bash">kubectl logs &lt;frontend-pod&gt; -c tls-health-check
</code></pre>
<p>Expected output (recurring every 30s once drift hits):</p>
<pre><code>error 10 at 0 depth lookup: certificate has expired
</code></pre>
<p>Cross-checking the node directly confirms the root cause:</p>
<pre><code class="language-bash">kubectl debug node/&lt;node&gt; --profile=sysadmin -it --image=ubuntu:22.04 -- bash
nsenter --target 1 --mount --pid -- date
nsenter --target 1 --mount --pid -- systemctl list-units --type=service --state=running | grep -iE 'ntp|chrony|timesync'
</code></pre>
<p><code>date</code> reports a timestamp ~30 days ahead of the rest of the cluster. The second command reveals which time-sync service is actually masked; on the CloudLab test cluster this was <code>ntp.service</code>; <strong>the specific service name is environment-dependent</strong>, which is why the oracle (below) doesn't check for any specific daemon by name.</p>
<p>A correct diagnosis identifies the actual cause: the certificate is healthy, the node's clock is wrong, and that skew is what makes TLS validation fail.</p>
<p>The canonical mitigation, from a privileged pod on the affected node (substituting whatever service name was found above <code>ntp</code> as an example):</p>
<pre><code class="language-bash">nsenter --target 1 --mount --uts --ipc --net --pid -- systemctl unmask ntp
nsenter --target 1 --mount --uts --ipc --net --pid -- systemctl start ntp
</code></pre>
<p>Restarting the frontend pod alone should <strong>not</strong> pass as the node's clock is still skewed, so a new pod scheduled on the same node would immediately hit the same TLS failure once the sidecar runs again.</p>
<h3>Oracle behavior</h3>
<p>The default pod-health oracle can't grade this either, since pods stay <code>Running/Ready</code> throughout, so this PR adds <code>NodeClockDriftMitigationOracle</code>.</p>
<p>It passes only when:</p>
<ol>
<li>The node's clock is within 60 seconds of the conductor's own clock.</li>
<li>No pod on the affected node is in a non-<code>Running</code>/<code>Succeeded</code> phase, or shows a <code>waiting</code>/<code>terminated</code> container state referencing <code>certificate</code> or <code>tls</code>.</li>
</ol>

<h2>4. Agent behavior and validation</h2>
<p>Manual testing (<code>kubectl</code>/<code>cli.py</code>): confirmed the fault injects correctly (clock confirmed skewed via <code>kubectl debug node</code>, x509 errors confirmed appearing in sidecar logs), and the oracle correctly returns <code>False</code> while the clock remains skewed and <code>True</code> once the time-sync service is unmasked, restarted, and the clock resyncs.</p>
<p>Stratus, three runs (judge matching model):</p>

Run | Model | Diagnosis
-- | -- | --
1 | gpt-4o-mini | Failed; attributed restarts to a PVC delay on the geo pod
2 | gpt-4o | Failed; attributed restarts to a Jaeger tracing config issue
3 | gpt-4o | Failed; attributed restarts to a Jaeger tracing misconfiguration/connectivity issue

<p><code>gpt-4o-mini</code> was expected to underperform; testing with <code>gpt-4o</code> was included as a stronger baseline but produced similarly unsuccessful diagnoses, plan to test against a more capable model as well if this approach for the problem is okay.</p>
<p>In all runs the agent noticed several pods had nonzero restart counts and investigated logs, events, the Jaeger dependency graph, and Prometheus CPU/memory metrics (all of which were clean) and never inspected the <code>frontend</code> pod specifically (despite it being the only pod with 2/2 containers instead of 1/1), nor checked node clocks, time-sync status or any time related detail on any node. This matches the intended difficulty: the certificate itself is valid, so log-level and metric-level investigation alone doesn't surface the fault, and the agent has to connect node clock state to the TLS symptom.</p></body></html>
</body>
</html>


haani-maybe:
Quick note: I'll run a few more agent runs for further testing purposes on different models if this solution approach is okay, for now I've only tested on gpt 4o with Stratus. 

I'd appreciate any feedback or any type of suggestions for the code, for both the mitigation oracle as well as the injection

sregym/conductor/problems/node_clock_drift.py
```diff
@@ -0,0 +1,413 @@
...
+            }
+        }
+
+        try:
+            self.kubectl.patch_deployment("frontend", self.namespace, patch_body)
+            print("[TLS] Added TLS verification sidecar to frontend deployment")
+        except Exception as e:
+            print(f"[TLS] Warning: Could not patch frontend deployment with sidecar: {e}")
+
+    def _wait_for_sidecar_rollout(self, timeout: int = 120) -> None:
```
Saadmrp1038: During testing, the diagnostic signal (x509 errors in sidecar logs) seems to be delayed 5+ minutes after fault injection. The signal did eventually appear, but an agent investigating immediately after injection would see empty sidecar logs for several minutes.

sregym/conductor/problems/node_clock_drift.py
```diff
@@ -0,0 +1,413 @@
...
+        print(f"Target node: {self.target_node}")
+
+        self._advance_node_clock(self.target_node)
+        print(f"Advanced system clock on {self.target_node} by {self.clock_drift_seconds}s (30 days)")
+
+        time.sleep(10)
+        print("Node clock skewed. x509 certificate errors will now appear in sidecar log")
+
+    @mark_fault_injected
+    def recover_fault(self):
```
Saadmrp1038, line 318: `recover_fault` has to undo everything `inject_fault` does. Right now it only deletes the injector pods. Need to fix the clock skew and NTP. Also, currently if the run is stopped midway the cluster stays in a broken state. Need to add a global clean up in `conductor.py`. Check the `fix_kubernetes` function.

sregym/conductor/problems/registry.py
Saadmrp1038, line 1: This problem will corrupt the entire host clock if run on a kind cluster since kind nodes share the host kernel clock. It needs to be added to `non_emulated_cluster_problems` in the registry.

sregym/conductor/oracles/node_clock_drift_mitigation.py
```diff
@@ -0,0 +1,177 @@
...
+        self.core_v1 = client.CoreV1Api()
+
+    def evaluate(self) -> dict:
+        print("== Node Clock Drift Mitigation Evaluation ==")
+
+        results = {}
+
+        try:
+            if not hasattr(self.problem, 'target_node') or self.problem.target_node is None:
+                target_node = self._find_affected_node(self.problem.namespace)
```
Saadmrp1038, line 26: This fallback just returns the first pod's node, which may not be the node with clock drift (our real cluster setup have three nodes). Since this is only hit when `self.problem.target_node` is `None`, it's unlikely to fire during normal operation. But if it does, the oracle could check the wrong node and produce a misleading result.

Saadmrp1038 (changes requested):
@haani-maybe Thanks for the effort!

I ran and tested the problem end-to-end on a 3-node CloudLab cluster. I found some issues. I left inline comments below with details. Please ping me when you've addressed them.

sregym/conductor/problems/node_clock_drift.py
haani-maybe: updated, 
I think this delay was because of the wait for openssl, I added in a wait for the sidecar rollout (wait_for_sidecar_rollout) so that the clock drift only happens once that's done and the sidecar is ready and inject_fault now waits for the sidecar rollout to finish
also added a marker for the sidecar (/tmp/sidecar-ready)

sregym/conductor/problems/node_clock_drift.py
haani-maybe: added a global cleanup and fixed the recover_fault

sregym/conductor/problems/registry.py
haani-maybe: added! 
(quick side note but non_emulated_cluster_problems is currently empty, minus this one i added, is that supposed to be the case?)

sregym/conductor/oracles/node_clock_drift_mitigation.py
haani-maybe: updated, the find_affected_node now finds the frontend pod and checks for sidecar presence so the target_node has to be the same as the node with the clock drift

haani-maybe:
@Saadmrp1038 hii, updated! sorry for the delay

the changed code is all the fixes you mentioned (elaborated in the replies to your code review comments) + one tiny thing more, i added a guard to ensure the node we're injecting on is one of the worker nodes (not control node)
this wasn't necessary since control node's are tainted so pods arent scheduled on them, but while I was running it I saw it happen on 2 runs out of around 10, which i assume was an issue on cloudlab setup end, so i added that just incase. 

although once again the control node IS tainted within standard setup so this wouldn't be a problem, but just incase someone else is running the problem and there's an issue in setup due to which the taint isn't present or misconfigured, the guard might help clarify 
(<code>_WORKER_ONLY_AFFINITY </code> in <code>node_clock_drift.py</code>, line 40 onwards as well as <code>is_control_plane_node</code>, same file, line 98 and onwards)

please let me know if the fixes are fine and if there's anything else to add/edit :)

sregym/conductor/problems/node_clock_drift.py
```diff
@@ -0,0 +1,591 @@
...
+                            self.stopped_services = services_str.split()
+                        break
+            except ApiException as e:
+                print(f"Warning: could not read injector logs to determine stopped services: {e}")
+
+        except ApiException as e:
+            print(f"Failed to create injector pod: {e}")
+            raise
+
+    def _restore_node_clock(self, node: str) -> None:
```
Saadmrp1038, line 440: Can we make this idempotent. By that I mean if time is already correct do not modify it again. If NTP syncs first the clock is already correct, and then this function will mess up node time.

sregym/conductor/problems/node_clock_drift.py
```diff
@@ -0,0 +1,591 @@
...
+            }
+        }
+
+        try:
+            self.kubectl.patch_deployment("frontend", self.namespace, patch_body)
+            print("[TLS] Added TLS verification sidecar to frontend deployment")
+        except Exception as e:
+            print(f"[TLS] Warning: Could not patch frontend deployment with sidecar: {e}")
+
+    def _wait_for_sidecar_rollout(self, timeout: int = 300) -> None:
```
Saadmrp1038: Increase the timeout. Sometimes it takes a lot of time to install openssl in CloudLab. In a couple of runs it took 300s+ resulting in timeout from this function.

sregym/conductor/problems/node_clock_drift.py
```diff
@@ -0,0 +1,591 @@
...
+            nsenter --target 1 --mount --uts --ipc --net --pid -- date
+            nsenter --target 1 --mount --uts --ipc --net --pid -- \
+                date -s "+{self.clock_drift_seconds} seconds"
+            nsenter --target 1 --mount --uts --ipc --net --pid -- date
+
+            echo "Clock drift injection complete. Keeping pod alive:"
+            tail -f /dev/null
+        """
+
+        pod_name = f"clock-drift-inject-{int(time.time() * 1000)}"
```
Saadmrp1038: We should use generic name here and not expose fault related info.

sregym/conductor/problems/node_clock_drift.py
```diff
@@ -0,0 +1,591 @@
...
+        self.injector_pod_name = pod_name
+
+        pod_spec = {
+            "apiVersion": "v1",
+            "kind": "Pod",
+            "metadata": {
+                "name": pod_name,
+                "namespace": self.clock_injector_namespace,
+                "labels": {"app": "clock-drift-injector", "inject-role": "drift"},
```
Saadmrp1038: We should use generic name here and not expose fault related info.

sregym/conductor/oracles/node_clock_drift_mitigation.py
```diff
@@ -0,0 +1,180 @@
...
+        except Exception as e:
+            print(f"Error getting node time: {e}")
+            return None
+        finally:
+            try:
+                self.core_v1.delete_namespaced_pod(pod_name, "default", grace_period_seconds=0)
+            except Exception:
+                pass
+
+    def _check_pod_health(self, namespace: str, target_node: str) -> bool:
```
Saadmrp1038, line 131: I think if pod does not exist it should return False. Otherwise, if an agent deletes the deployment it will Pass.

sregym/generators/fault/inject_remote_os.py
```diff
@@ -335,6 +336,186 @@ def recover_disk_pressure_all(self):
...
 
+    def recover_clock_drift(self):
+        """Detect leftover clock-drift injector/restore pods from interrupted
+        run, restore affected node by doing: unmask/restart timesync
+        service + clean up of leftover pods
+        """
+        try:
+            leftover_pods_raw = self.kubectl.exec_command(
+                "kubectl get pods -n default "
+                "-l 'app in (clock-drift-injector,clock-drift-restorer)' "
```
Saadmrp1038: We should use generic name here and not expose fault related info.

sregym/conductor/problems/node_clock_drift.py
```diff
@@ -0,0 +1,591 @@
...
+
+        except ApiException as e:
+            print(f"Failed to create injector pod: {e}")
+            raise
+
+    def _restore_node_clock(self, node: str) -> None:
+        """Step the node's clock back by the same offset it was advanced + restore
+        whichever time-sync service was masked during injection.
+        """
+        services = self.stopped_services or []
```
Saadmrp1038: If `self.stopped_services` is empty (because injection crashed before log parsing, or `ApiException` happened), then services = [] -> the restore script prints "No recorded stopped services to restore" and skips NTP unmask entirely. The node stays with NTP masked.

---

This is a bit out of the wild example but since we are messing with a very sensitive value (node clock time) it's better to be careful and handle this.

Saadmrp1038 (changes requested):
@haani-maybe Sorry for the delayed review.
I tested the fault thoroughly and did not find any major bugs. Only found some minor things. Can you please address them. Meanwhile I will do some agent runs.

sregym/conductor/problems/node_clock_drift.py
haani-maybe: Added a time skew checker that skips if NTP already self-corrected into restore_node_clock

sregym/conductor/problems/node_clock_drift.py
haani-maybe: Doubled the timeout (600s)

sregym/conductor/problems/node_clock_drift.py
haani-maybe: made the prefix node-probe instead of clock-drift now, to prevent this

sregym/conductor/oracles/node_clock_drift_mitigation.py
haani-maybe: Updated to add the check to return False if no pods

sregym/conductor/problems/node_clock_drift.py
haani-maybe: Updated to just have the label "node-probe"

sregym/generators/fault/inject_remote_os.py
haani-maybe: Updated to use generic node-probe

sregym/conductor/problems/node_clock_drift.py
haani-maybe: updated so now even when services is empty it still falls back to discovering the masked time sync services

haani-maybe:
> @haani-maybe Sorry for the delayed review. I tested the fault thoroughly and did not find any major bugs. Only found some minor things. Can you please address them. Meanwhile I will do some agent runs.

@Saadmrp1038 No worries! I added in all the required changes, let me know if there's anything else

Saadmrp1038 (approved):
@haani-maybe Working great now!

I did 5 agent runs on this (claudecode+sonnet-5). 4 times the agent failed both diagnosis and mitigation. The only time it succeeded was when I added a hint ("Check node clock") in the prompt. Without the hint, agents get stuck investigating the sidecar's cert expiry or startup restarts and never make the leap to checking the node's system clock.

So, seems like it's quite a hard problem which is exactly what we want for the benchmark. Also, thanks for taking the time. It took quite a while for me to review this PR.

Congrats on the first PR merged here 🎉

tianyin:
WoW This is incredible persistence!

haani-maybe:
> @haani-maybe Working great now!
> 
> I did 5 agent runs on this (claudecode+sonnet-5). 4 times the agent failed both diagnosis and mitigation. The only time it succeeded was when I added a hint ("Check node clock") in the prompt. Without the hint, agents get stuck investigating the sidecar's cert expiry or startup restarts and never make the leap to checking the node's system clock.
> 
> So, seems like it's quite a hard problem which is exactly what we want for the benchmark. Also, thanks for taking the time. It took quite a while for me to review this PR.
> 
> Congrats on the first PR merged here 🎉

No worries at all! Glad I could start contributing here :)

tianyin:
@haani-maybe I'm not sure if you are already in the Slack or are here for the UIUC+ program. If so, let me know so I can add you into the channels.

haani-maybe:
> @haani-maybe I'm not sure if you are already in the Slack or are here for the UIUC+ program. If so, let me know so I can add you into the channels.

@tianyin  I joined the Slack (reu-2026-general and benchmark channel) when I got the evaluation task email for the UIUC+ program, are there any other channels I have to join for the program?

tianyin:
Can you DM me on the Slack? I can't find you... (and I'm guessing you don't want to reveal your name/ID on GitHub)

PR #896, opened 2026-06-19
 Fix MCP retry error handling and CRLF parsing in CI

Saadmrp1038:
Found some bugs over the last 2 weeks. This PR fixes them.

- Handle `anyio.BrokenResourceError`, `anyio.ClosedResourceError`, and `McpError` in the MCP tool retry loop so transient connection issues don't crash the agent. Return a descriptive error string instead of re-raising after retries are exhausted.
- Strip `\r` from PR description lines in the problem-validation workflow to avoid parse failures from CRLF line endings.


tianyin (approved):
LGTM

PR #899, opened 2026-06-21
Decouple failure configmaps from mongo pods in `hotel-reservation`

### LINKED ISSUE #882: Distracting configmaps in `hotel-reservation`
During agent runs ` failure-admin-geo / failure-admin-rate` configmaps are a major red herring and can distract the agent on more subtle faults. Ideally these should not be present on runs where they are not needed.

Saadmrp1038:
Fixes #882

Implemented based @yimingsu01's suggestion (See https://github.com/SREGym/SREGym/issues/882#issuecomment-4683841548).

The configmaps are now managed separately from the base deployment:

- Unrelated faults (default): configmaps get populated with scripts and patched into MongoDB deployments as noise. Same end result for the agent as before.
- Related faults (`revoke_auth`, `storage_user_unregistered`): no failure configmaps created. Injection and recovery pipe scripts via stdin so nothing is written to the pod. So, no fault related data leakage to the agent.

Related PR: https://github.com/SREGym/SREGym-applications/pull/5

/validate-problem skip


tianyin (approved):
LGTM

Issue #898, opened 2026-06-20
Feature Request: Persist shell session state across `SREGym>` CLI commands


W-OK-E:
This request partially stems from the testing I have been doing for #883 :) Keeping track of the namespaces, registries and IPs involved was slightly annoying and I found that the usual `export VAR=VALUE` method does not work in the `SREGym>` cli.

Currently, every `SREGym>` command is executed in a fresh shell process with no retained session state between commands which means workflows like the following do not behave as users typically expect from an interactive shell:

```bash
SREGym> export NS=astronomy-shop
SREGym> kubectl get pods -n $NS   # NS is lost
```
---

#### I am proposing a minor change for shell session state to persist across CLI commands in the dev/debug cli that would:
* Preserve:
  * exported environment variables
  * shell function definitions
  * working directory (`$PWD`)
* Restore this state before executing the next command

This would make a sequence of `SREGym>` commands behave much closer to a normal interactive shell session while still keeping each command isolated in its own subprocess.

---

## Implementation Approach

Instead of maintaining a single long-lived piped shell process, the implementation snapshots shell state after each command and restores it before the next one.

After every command:

* exported env vars are captured via:

  ```bash
  declare -px
  ```
* shell functions via:

  ```bash
  declare -pf
  ```
* current working directory via:

  ```bash
  $PWD
  ```
#### These are written to a per-PID temporary state file. Before the next command executes, the shell sources this state file to restore continuity.
---
## Scope

This change would affect only the developer/debug CLI path (`cli.py`).

There would be no behavior changes for:

* `main.py`
* agent execution flows
* non-interactive runtime behavior

---

If I have missed something here, or if there are other ways that already do what I am requesting for, or if this does not add much value to the current debugging state, please let me know. If the maintainers do approve,  I you approve, I can go ahead and open a PR.

Saadmrp1038:
@W-OK-E It's interesting that you use it like this. I never run any commands in the cli session and just open another terminal window by the side. But yeah definitely worth adding this feature. Feel free to give a PR.

tianyin:
This could be a useful feature!

W-OK-E:
PR raised in #900

PR #900, opened 2026-06-21
Shell state persistence

W-OK-E:
This PR emerges from the discussions in #883 and #898 to enable shell-state persistence across `SREGym>` cli commands used for debugging.

## What Changed

`Shell.exec` now persists **exported variables, shell functions, and the working directory** across successive `SREGym>` CLI commands, so a sequence of commands behaves like one interactive shell session:

```bash
export REGISTRY_IP=$(kubectl get svc ... -o jsonpath='{.spec.clusterIP}')
echo "$REGISTRY_IP"          # <- still set on the next command
```
Previously each command ran in a fresh `bash` with no memory of the last one, so `export`, functions, and `cd` were lost between commands.

## How it works

After each command we snapshot the exported environment (`declare -px`), function  (`declare -pf`), and `$PWD` to a per-process state file (one per CLI PID, the temp dir), and source it before the next command. Each command still runs in
its own short-lived `bash` process; the state file is what carries context forward.

## Scope

Dev/debug CLI only — `cli.py` is unchanged; it simply calls `Shell.exec`. No behavior change for `main.py` / agent runs. The change lives entirely in `shell.py

One simple thing to point out - The changes only take place AFTER the debug shell has entered into a problem session i.e. after `start <problem-id` has executed since `cli.py` holds the session in a while loop checking for the `start` and `exit` commands.

Let me know if this seems fine or if any changes are required. 

Also, apologies for the delay in #883 @Saadmrp1038, I had hooked up a RAID HDD to run the local evaluations and it failed so kind of working things out there, but it will be done soon too.

### DIFF
--- a/sregym/conductor/oracles/mitigation.py
+++ b/sregym/conductor/oracles/mitigation.py
@@ -11,6 +11,11 @@ _ROLLOUT_POLL_INTERVAL = 5
 class MitigationOracle(Oracle):
     importance = 1.0
 
+    def __init__(self, problem):
+        super().__init__(problem)
+        deployments = self.problem.kubectl.list_deployments(self.problem.namespace)
+        self.replica_count = {dep.metadata.name: dep.spec.replicas for dep in deployments.items}
+
     def _wait_for_rollouts(self, kubectl, namespace):
         """Wait for all deployments in the namespace to finish rolling out."""
         deadline = time.monotonic() + _ROLLOUT_SETTLE_SECONDS
@@ -19,7 +24,7 @@ class MitigationOracle(Oracle):
             all_settled = True
             for dep in deployments.items:
                 status = dep.status
-                desired = dep.spec.replicas or 1
+                desired = dep.spec.replicas if dep.spec.replicas is not None else 1
                 if (
                     (status.updated_replicas or 0) < desired
                     or (status.ready_replicas or 0) < desired
@@ -42,6 +47,26 @@ class MitigationOracle(Oracle):
         # Wait for any in-progress rollouts to finish so we don't evaluate
         # a transient state where old pods are gone and new ones haven't crashed yet.
         self._wait_for_rollouts(kubectl, namespace)
+
+        deployments = kubectl.list_deployments(namespace)
+        current_deps = {dep.metadata.name: dep for dep in deployments.items}
+
+        for name in self.replica_count:
+            if name not in current_deps:
+                print(f"❌ Deployment '{name}' was deleted")
+                results["success"] = False
+                return results
+            dep = current_deps[name]
+            desired = dep.spec.replicas if dep.spec.replicas is not None else 1
+            if desired == 0:
+                print(f"❌ Deployment '{name}' was scaled to 0")
+                results["success"] = False
+                return results
+            ready = dep.status.ready_replicas or 0
+            if ready < desired:
+                print(f"❌ Deployment '{name}' has {ready}/{desired} replicas ready")
+                results["success"] = False
+                return results
 
         pod_list = kubectl.list_pods(namespace)
 
--- a/sregym/conductor/problems/registry.py
+++ b/sregym/conductor/problems/registry.py
@@ -26,8 +26,8 @@ from sregym.conductor.problems.ephemeral_port_range_hotel_reservation import Eph
 from sregym.conductor.problems.expired_tls_hotel_reservation import ExpiredTlsHotelReservation
 from sregym.conductor.problems.failed_readiness_probe import FailedReadinessProbe
 from sregym.conductor.problems.faulty_image_correlated import FaultyImageCorrelated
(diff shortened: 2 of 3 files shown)


sregym/service/shell.py
```diff
@@ -1,33 +1,57 @@
...
 
 
 class Shell:
     """Interface to run shell commands. Currently used for development/debugging with cli.py"""
 
-    @staticmethod
-    def exec(command: str, input_data=None, cwd=None):
-        """Execute a shell command on localhost."""
+    # One snapshot file per CLI process, reused across exec() calls.
+    _state_file = os.path.join(tempfile.gettempdir(), f"sregym_shell_state_{os.getpid()}.sh")
```
munimthahmid: This file is stored in shared `/tmp`, and its name is predictable. Because it stores exported variables such as API keys and is later sourced, another user could read it or create it first with their own commands. Could we use a randomly named private file with `0600` permissions and delete it on exit?

sregym/service/shell.py
```diff
@@ -1,33 +1,57 @@
...
             input_data = input_data.encode("utf-8")
 
+        state = shlex.quote(cls._state_file)
+        # If a cwd is explicitly requested, it wins over the persisted directory.
+        start_dir = shlex.quote(cwd) if cwd else "."
+        # Restore prior state -> run the user's command -> snapshot exported vars, functions,
+        # and cwd for the next call. The user command's own exit code is preserved via exit.
+        wrapped = (
+            f"[ -f {state} ] && source {state} 2>/dev/null; "
+            f"cd {start_dir} 2>/dev/null; "
```
munimthahmid: This issue will not block the current SREGym CLI, so it is minor, but it should still be fixed. 

For example, after `Shell.exec("cd /tmp")`, calling `Shell.exec("pwd", cwd="tests")` should run in `<repo>/tests`, but it returns `/tmp`. The persisted `/tmp` state is restored before the relative `cwd` is applied, so `tests` is resolved as `/tmp/tests`. Since that directory does not exist and the `cd` error is suppressed, the command remains in `/tmp`. 

This also contradicts the comment above, which says that an explicitly requested `cwd` wins over the persisted directory.

munimthahmid (changes requested):
@W-OK-E I tested this through the SREGym CLI, and the intended persistence works well for exported variables, shell functions, and the working directory. I found two issues and left inline comments about them. Could you please address them? Otherwise, the feature looks good.

sregym/service/shell.py
W-OK-E: Patched - resolves to absolute `cwd` not relative.

W-OK-E:
Hello @munimthahmid !

Thank you for reviewing the PR and the suggestions, I have made a few changes w.r.t the points raised. Kindly review them and let me know if it seems fine.

sregym/conductor/oracles/image_pull_secret_mitigation.py
munimthahmid, line 1: Seems like you mistakenly added the problem from #883 here. Can you remove it from this PR?

sregym/service/shell.py
```diff
@@ -1,33 +1,66 @@
...
 
 
 class Shell:
     """Interface to run shell commands. Currently used for development/debugging with cli.py"""
 
-    @staticmethod
-    def exec(command: str, input_data=None, cwd=None):
-        """Execute a shell command on localhost."""
+    # One snapshot file per CLI process, reused across exec() calls.
+    _state_file = os.path.join(tempfile.gettempdir(), f"sregym_shell_state_{os.getpid()}.sh")
```
munimthahmid: `_make_state_file()` is added but never used, so the random filename and cleanup do not take effect. Could this be changed to `_state_file = _make_state_file()`?

munimthahmid (changes requested):
@W-OK-E  The persistence feature works well, and the relative cwd issue is fixed. I left two comments about the unused state-file helper and changes that belong in #883

sregym/service/shell.py
W-OK-E: Apologies for the silly errors, I just realized messed up the git worktrees there, fixing it right now.

sregym/conductor/oracles/image_pull_secret_mitigation.py
W-OK-E: Apologies, removed.

W-OK-E:
Sincere apologies @munimthahmid again!  Will be careful handling multiple work-trees hereon.

munimthahmid (changes requested):
@W-OK-E  Sorry for the confusion. My earlier comment was attached to the oracle file, but I meant all changes from #883. The problem implementation and registry changes are still included here. Could you remove those as well?

W-OK-E:
I should be apologizing @munimthahmid. Have closed this since the current worktree is a mess. Primarily due to the hacks I was doing to test #883 on my NAS. I will create a new PR. Apologies for the inconvenience.

PR #901, opened 2026-06-22
Add nightly rebalance problem, RFC 862

iamyeezuschrist:
### 1. The failure background
Capacity-rebalancing automation is common on shared clusters: a scheduled controller scales down or tightens the resources of lower-priority processing workloads at off-peak hours to free capacity for more demanding apps (e.g. scaling down service pods overnight to hand the resources to MR/OLAP jobs). Generally, it is a nice way to utilize resources properly in a larger company, however, if rebalancing overshoots, it may drive a service's memory so low that the container can no longer initialize. The pod gets OOMKilled before serving a single request: no metrics exported, no logs. Since an external controller sets the resource limits for the pods, a one-off manual limit increase won't help mitigating the fault.

### 2. How to simulate it on SREGym

- **App:** Hotel Reservation, namespace `hotel-reservation`. 
- **Target:** the `recommendation` deployment (a lower-priority service).
- **Injection layer:** Direct K8s API only (CronJob + RBAC + deployment patch). 

- A CronJob **`vpa-updater` in `kube-system`** is deployed. It is named to
  mimic the VerticalPodAutoscaler updater, a real controller that patches pod resource limits — in order to blend in with platform components rather than being an obvious "chaos" pod. Schedule is `* * * * *`; each tick runs a one shot `kubectl patch` that drives `recommendation`'s memory limit down to `4Mi`. `inject_fault` also applies the squeeze once immediately so the fault is instantly live.
- `4Mi` is below the service's working set, so the `recommendation` pod is
    **OOMKilled during init -> CrashLoopBackOff** (or `CreateContainerError` when the squeezed limit falls below the container runtime's minimum). The crashed pod emits no metrics/logs of its own, observability blackout.
- **Metastable:** raising the limit is reverted on the next tick.
- `inject_fault` deletes any leftover CronJob first, so
  re-injection is idempotent.

### 3. How the agent should behave

The agent cannot diagnose from the `recommendation` service's metrics/logs/traces. It must reason from control-plane signals: pod/deployment status, `kubectl get events` (OOMKilled), and — the key tell - the memory limit **repeatedly reverting on a cycle**, which points to an existence of an external controller. A correct diagnosis localizes **both** the `recommendation` deployment **and** the `vpa-updater` CronJob in `kube-system` as the recurring root cause.

**Expected mitigation:** suspend or delete the `vpa-updater` CronJob **and** restore a sane memory limit on `recommendation`. A fix that only raises the limit is rejected, the next tick reverts it.

### 4. Oracle

- **Mitigation oracle** passes only when all hold: (a) the `vpa-updater` CronJob is absent or suspended (so the squeeze can't recur), (b) `recommendation`'s memory limit is restored to a healthy value, if the deployment originally had a limit -> must be ≥ it, if there was none, oracle accepts either no memory limit or any limit >= the 64Mi fallback floor, (c) target pods are Running/ready with no recent OOMKilled. Simply bumping the limit fails (a).

- **Diagnosis oracle:** the standard LLM-as-a-judge against the structured `root_cause`, written in the PR.

/validate-problem nightly_rebalance_oom_hotel_reservation

### 5. Real-world context

**Root cause.** A misconfigured VerticalPodAutoscaler (VPA), set up incorrectly by the core infra team, drove the affected pods' memory limits far too low. The containers couldn't even start since they were OOMKilled while bringing up their base runtime layer. Because the pod never came up, nothing was exported except the OOM events themselves.

**Mitigation at the time.** The on-duty engineer diagnosed it fairly quickly from the pod's configuration in the control plane: the memory limit was far below any value the team would normally set. The fix was to kill and remove the VPA daemon from the cluster, restore the deployment's default limits, and restart the pods.

This is what the simulation mirrors: the injected actor is a `vpa-updater`, the only healthy signal during the fault is control-plane state (that's why I say dark in it's own telemetry), and the oracle requires removing the controller and restoring a sane limit, not just bumping it.


sregym/conductor/oracles/nightly_rebalance_oom_mitigation.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def _memory_limit_sane(self, service: str, namespace: str) -> bool:
+        original = getattr(self.problem, "_original_memory_limit", None)
+        threshold = _parse_mem_to_bytes(original) if original else _FALLBACK_FLOOR_BYTES
+        dep = self.problem.kubectl.get_deployment(service, namespace)
+        for container in dep.spec.template.spec.containers:
+            limits = (container.resources.limits or {}) if container.resources else {}
+            mem = limits.get("memory")
+            if mem is None:
+                continue
```
ermias19: Instead of restoring, the mitigation  is actually removes the limit `mem`entirely instead of restoring. 

may we add one more check for the original memory limit? WDYT.

sregym/conductor/problems/nightly_rebalance_oom.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+        self.diagnosis_oracle = LLMAsAJudgeOracle(problem=self, expected=self.root_cause)
+        self.mitigation_oracle = NightlyRebalanceOOMMitigationOracle(problem=self)
+        self.app.create_workload()
+
+    @mark_fault_injected
+    def inject_fault(self):
+        print("== Fault Injection ==")
+        dep = self.kubectl.get_deployment(self.faulty_service, self.namespace)
+        container = dep.spec.template.spec.containers[0]
```
ermias19: minor issue tho, container is looking only the first index 0, its okay for recomodation single container but fragile. WDYT ?

sregym/conductor/problems/nightly_rebalance_oom.py
```diff
@@ -0,0 +1,220 @@
...
+                        print(f"Fault confirmed live: {last}")
+                        return
+            print(f"Waiting for squeeze to take effect... {last}")
+            time.sleep(5)
+        raise RuntimeError(f"Target {self.faulty_service} did not become unhealthy in time; last={last}")
+
+    def _restore_memory_limit(self):
+        if self._target_container is None:
+            return
+        limits = {"memory": self._original_memory_limit or "256Mi"}
```
munimthahmid: This does not fully restore the original deployment spec. The baseline Hotel Reservation `recommendation` container has `limits.cpu` only and no `limits.memory`; after `recover_fault()` this code leaves `memory: 256Mi` on the deployment.

sregym/conductor/problems/nightly_rebalance_oom.py
```diff
@@ -0,0 +1,220 @@
...
+        self.kubectl.patch_deployment(self.faulty_service, self.namespace, self._squeeze_patch())
+
+    def _wait_for_target_unhealthy(self, timeout: int):
+        unhealthy_reasons = {
+            "CrashLoopBackOff",
+            "CreateContainerError",
+            "RunContainerError",
+            "CreateContainerConfigError",
+            "Error",
+            "ImagePullBackOff",
```
munimthahmid: `ImagePullBackOff` / `ErrImagePull` are unrelated to the memory-squeeze mechanism, since the container never starts. Accepting them here could make the injection look successful for an image-pull issue.

sregym/conductor/problems/nightly_rebalance_oom.py
```diff
@@ -0,0 +1,220 @@
...
+        self._wait_for_target_unhealthy(timeout=180)
+        print(
+            f"Service: {self.faulty_service} | squeeze={self.squeeze_memory} | actor={self.actor_namespace}/{self.actor_name}\n"
+        )
+
+    @mark_fault_injected
+    def recover_fault(self):
+        print("== Fault Recovery ==")
+        self._delete_cronjob()
+        time.sleep(5)  # let any in-flight tick finish before restoring
```
munimthahmid: This fixed sleep assumes any in-flight CronJob is gone before we restore the deployment. Since `_delete_cronjob()` uses background deletion, that may be flaky on slower clusters or while the kubectl image is still starting. Can we poll until the CronJob and any active actor Jobs/Pods are gone before restoring the memory limit?

sregym/conductor/problems/nightly_rebalance_oom.py
```diff
@@ -0,0 +1,220 @@
...
+from sregym.conductor.oracles.nightly_rebalance_oom_mitigation import NightlyRebalanceOOMMitigationOracle
+from sregym.conductor.problems.base import Problem
+from sregym.service.apps.hotel_reservation import HotelReservation
+from sregym.service.kubectl import KubeCtl
+from sregym.utils.decorators import mark_fault_injected
+
+
+class NightlyRebalanceOOM(Problem):
+    actor_name = "vpa-updater"
+    actor_namespace = "kube-system"
```
munimthahmid, line 20: Since this fault creates a recurring actor in `kube-system`, can we add a global cleanup path for interrupted runs? If a run is killed after injection and before  `recover_fault()`, app cleanup will not remove `CronJob/vpa-updater` or the kube-system ServiceAccount.

sregym/conductor/problems/nightly_rebalance_oom.py
```diff
@@ -0,0 +1,220 @@
...
+                "template": {
+                    "spec": {"containers": [{"name": self._target_container, "resources": {"limits": limits}}]}
+                }
+            }
+        }
+        self.kubectl.patch_deployment(self.faulty_service, self.namespace, patch)
+
+    def _create_cronjob(self):
+        patch_json = json.dumps(self._squeeze_patch(), separators=(",", ":"))
+        cmd = f"kubectl -n {self.namespace} patch deployment {self.faulty_service} --type=strategic -p '{patch_json}'"
```
munimthahmid: One benchmark-quality concern: this command makes the root cause very explicit once the agent inspects the CronJob, since it names the target deployment and includes the exact `4Mi` patch. That may make the task easier than intended despite the `vpa-updater` name blending in.

Can we make the actor less self-revealing, for example, by moving the target/limit into a less direct policy/config layer or adding realistic decoy scheduled jobs/controllers so the answer is not exposed by a single CronJob command?

munimthahmid (changes requested):
Thanks for the PR. This is a very interesting problem.

However, it would be better if we had a bit more context on the real-world failure. You mentioned in the RFC that this issue occurred in your company. Can you add a small section in the PR briefly explaining the root cause and the mitigation that was done at the time? That would be great.

In the Oracle section, can we adjust the wording for condition (b)? The implementation is more flexible than "restored to >= its original baseline". if the original deployment has no memory limit, the oracle accepts either no memory limit or any memory limit above the fallback threshold (`64Mi`). So can you update the PR description to match that?

Other than these, I have left a few inline comments. Can you address them? Thanks!

sregym/conductor/oracles/nightly_rebalance_oom_mitigation.py
iamyeezuschrist: @ermias19 you're right, that's quite a dull mistake from my side. 

updated the restoring with checks

https://github.com/SREGym/SREGym/pull/901/changes/98d154fb38da42358c78a51d7ab9b0304033212e#diff-76f8fa5a62a252dd4882f05f8f0d90eadb35290b5151bcfee9c2cd55d3e6b5d0L121

updated the oracle

https://github.com/SREGym/SREGym/pull/901/changes/98d154fb38da42358c78a51d7ab9b0304033212e#diff-c31d027253563ee8e28ae8d6408863adef656cc6d81a23a88a266b84b1a55d5cR72

sregym/conductor/problems/nightly_rebalance_oom.py
iamyeezuschrist: indeed, random sidecar could result into unexpected behavior.

use the service name for identification now, fallback to [0]

https://github.com/SREGym/SREGym/pull/901/changes/98d154fb38da42358c78a51d7ab9b0304033212e#diff-76f8fa5a62a252dd4882f05f8f0d90eadb35290b5151bcfee9c2cd55d3e6b5d0R57

sregym/conductor/problems/nightly_rebalance_oom.py
iamyeezuschrist: yep, the errors list is too broad. left only the backoff and create error (more related to original problem)

sregym/conductor/problems/nightly_rebalance_oom.py
iamyeezuschrist: changed for polling, thanks for pointing that out. sleep is flaky. 

https://github.com/SREGym/SREGym/pull/901/changes/98d154fb38da42358c78a51d7ab9b0304033212e#diff-76f8fa5a62a252dd4882f05f8f0d90eadb35290b5151bcfee9c2cd55d3e6b5d0R190

sregym/conductor/problems/nightly_rebalance_oom.py
iamyeezuschrist: https://github.com/SREGym/SREGym/pull/901/changes/98d154fb38da42358c78a51d7ab9b0304033212e#diff-76f8fa5a62a252dd4882f05f8f0d90eadb35290b5151bcfee9c2cd55d3e6b5d0R210

added the global cleanup for stability.

we ignore the 404's to silence the errors when trying to remove a non-existing resource. not sure if it's the best approach here, couldn't come up with better choice tbh.

sregym/conductor/problems/nightly_rebalance_oom.py
iamyeezuschrist: maybe you're right, I only tried the Haiku and it wasn't capable of resolving the problem from cronjob name, but Sonnet and Opus weren't tested and maybe it's a loophole that can be fixed easily to improve the problem.

i've changed it a bit so the agent has to first find cron, then read the policy config and make some conclusions. WDYT? 

https://github.com/SREGym/SREGym/pull/901/changes/98d154fb38da42358c78a51d7ab9b0304033212e#diff-76f8fa5a62a252dd4882f05f8f0d90eadb35290b5151bcfee9c2cd55d3e6b5d0R133

iamyeezuschrist:
> However, it would be better if we had a bit more context on the real-world failure. You mentioned in the RFC that this issue occurred in your company. Can you add a small section in the PR briefly explaining the root cause and the mitigation that was done at the time? That would be great.

added several sentences on how it was handled in production, not much to say there tbh, we saw the limits were extremely low, the product context was fairly an advantage here, since we know of the VPA's existence in our system. 

though I hope it adds some clarity on the context and the reasons. please share WDYT. thanks!


> In the Oracle section, can we adjust the wording for condition (b)? The implementation is more flexible than "restored to >= its original baseline". if the original deployment has no memory limit, the oracle accepts either no memory limit or any memory limit above the fallback threshold (`64Mi`). So can you update the PR description to match that?


you're right, thanks, updated it for more transparency.


> Other than these, I have left a few inline comments. Can you address them? Thanks!

thanks for mentioning those! i indeed overlooked quite obvious issues. addressed the comments with links to additions and fixes.

sregym/conductor/problems/nightly_rebalance_oom.py
iamyeezuschrist: reworked here as well

https://github.com/SREGym/SREGym/pull/901/changes/98d154fb38da42358c78a51d7ab9b0304033212e#diff-76f8fa5a62a252dd4882f05f8f0d90eadb35290b5151bcfee9c2cd55d3e6b5d0L121

sregym/conductor/problems/nightly_rebalance_oom.py
```diff
@@ -0,0 +1,272 @@
...
+                cron_gone = True
+            jobs = batch.list_namespaced_job(self.actor_namespace, label_selector=label).items
+            pods = core.list_namespaced_pod(self.actor_namespace, label_selector=label).items
+            if cron_gone and not jobs and not pods:
+                return
+            time.sleep(3)
+        print(f"⚠️ Actor {self.actor_namespace}/{self.actor_name} not fully gone after {timeout}s; restoring anyway")
+
+    @classmethod
+    def cleanup_leftover_actor(cls):
```
munimthahmid, line 228: When running end to end  with Codex, I hit a cleanup issue between repeated runs: a stale `kube-system/vpa-updater` Job pod was left stuck in `ContainerCreating` after its ConfigMap/ServiceAccount had already been deleted, which blocked the next run at kube-system readiness.

Could we make this cleanup more complete? For example, delete the CronJob/Jobs using foreground deletion, or explicitly wait for the `app=vpa-updater` Jobs/Pods to be gone before deleting the ConfigMap/ServiceAccount. It would also be good to clean up the Role/RoleBinding here too.

sregym/conductor/oracles/nightly_rebalance_oom_mitigation.py
```diff
@@ -0,0 +1,90 @@
...
+    def evaluate(self) -> dict:
+        print("== Nightly Rebalance OOM Mitigation Evaluation ==")
+        results = super().evaluate()
+        if not results.get("success"):
+            return results
+
+        service = self.problem.faulty_service
+        namespace = self.problem.namespace
+
+        if self._rebalancer_active():
```
munimthahmid: I think this mitigation check may be a bit too strict. The injected issue is that `vpa-updater` repeatedly applies an unsafe `4Mi` memory limit. Deleting/suspending the CronJobfixes that, but fixing the policy so it no longer applies an unsafe value seems like a valid mitigation too.

In a Codex high-reasoning run, the agent found `vpa-updater`, updated the policy to `128Mi`, restored `recommendation`, and the service stayed healthy after the CronJob ran again. The run still failed because the CronJob remained active. 

I’d prefer accepting mitigation if either the CronJob is removed/suspended, or it remains active but no longer applies an unsafe memory limit and the target service stays healthy.

munimthahmid (changes requested):
@iamyeezuschrist  I left some inline comments. Can you review them?

I ran the Codex agent 3 times on this problem (2 times with medium reasoning and 1 time with high reasoning), and it failed all 3 times. In the medium runs, it found the visible symptom quickly: `recommendation` gets a bad `4Mi` memory limit and OOMs. But it missed the actual root cause, which is that `kube-system/vpa-updater` is repeatedly applying that limit.

The high-reasoning run got further. Codex still submitted the diagnosis too early, so diagnosis failed, but during mitigation it did eventually find `vpa-updater` / `vpa-updater-policy`, updated the policy to `128Mi`, restored `recommendation`, and verified it stayed healthy after the CronJob ran again. It still failed because the oracle requires the CronJob itself to be removed/suspended. I think that case might be due to the mitigation oracle being a bit too strict (check inline comment).

Overall, I think the problem is medium-hard for the agent. The symptom is clearly visible, but the root cause is harder because the agent has to investigate why the bad limit keeps coming back. SOTA models might be able to fix it quickly, but this still seems like a reasonable benchmark.

I also left one cleanup-related inline comment from repeated runs. Could you also pull latest `main` and resolve the small merge conflict?

sregym/conductor/oracles/nightly_rebalance_oom_mitigation.py
iamyeezuschrist: i've changed the deletion as the miltigation for survives next tick function

https://github.com/SREGym/SREGym/pull/901/changes/7cec120ca6802ae1419ef3e895e48f0a12ff56d3#diff-c31d027253563ee8e28ae8d6408863adef656cc6d81a23a88a266b84b1a55d5cR62

it polls the cycle and checks if the cronjob is deleted || the limits are sane.

please share WDYT, i believe it's a more fair scenario. couldn't think of any potential flaws that can abuse that judgement.

sregym/conductor/problems/nightly_rebalance_oom.py
iamyeezuschrist: i've changed the deletion logic a bit, see teardown_actor (L180-183) 

https://github.com/SREGym/SREGym/pull/901/changes/7cec120ca6802ae1419ef3e895e48f0a12ff56d3#diff-76f8fa5a62a252dd4882f05f8f0d90eadb35290b5151bcfee9c2cd55d3e6b5d0R180

https://github.com/SREGym/SREGym/pull/901/changes/7cec120ca6802ae1419ef3e895e48f0a12ff56d3#diff-76f8fa5a62a252dd4882f05f8f0d90eadb35290b5151bcfee9c2cd55d3e6b5d0R185

and also added polling for actor to be gone (L206), please see and share your opinion. it has to be better rn.

thanks for your time reviewing again!

munimthahmid (approved):
Approved after adding ba486ac3, which requires the mitigation oracle to observe a fresh rebalancer tick before passing an active CronJob mitigation. This closes the reproduced no-tick oracle bypass while preserving the safe-policy path. Let me do some agent run.

munimthahmid:
@iamyeezuschrist I ran this 3 times with Codex 5.6 Sol: medium x2 and xhigh x1.

Results:

- Medium attempt 1: diagnosis passed `89/100`, mitigation failed.
- Medium attempt 2: diagnosis passed `78/100`, mitigation failed.
- Xhigh attempt 1: diagnosis passed `89/100`, mitigation passed.

Pattern was pretty consistent: medium found the `recommendation` OOM / bad `4Mi` memory limit, but missed the `vpa-updater` CronJob reapplying it, so mitigation failed. Xhigh found `vpa-updater` / `vpa-updater-policy`, fixed that durable source, and passed.

So I think this is a good addition to the benchmark. One thing I noticed after these run is diagnosis judge can still pass even when the answer misses `vpa-updater`, but I think that’s probably okay here since mitigation oracle catches it. All good from my end.

@Saadmrp1038 can do a final check and merge it.

tianyin:
@munimthahmid I think you should just feel free to merge. If it's pass you, I'm confident that it works. 

With the scale of the project, let's avoid bottleneck of making @Saadmrp1038 the only person who can merge. (I know he's a superman)

Saadmrp1038:
@munimthahmid Yeah. Feel free to merge if you think the PR is okay !!!

munimthahmid:
@iamyeezuschrist I updated the diagnosis judge to explicitly require `vpa-updater`, so now it is consistent with the mitigation oracle. I also did 6 ( :) ) more runs with Codex 5.6 Sol.

Results:

- Medium: diagnosis failed `56/100`, mitigation failed.
- Xhigh attempt 1: diagnosis failed `56/100`, mitigation failed.
- Max attempt 1: diagnosis failed `56/100`, mitigation failed.
- Max attempt 2: diagnosis failed `44/100`, mitigation failed.
- Xhigh attempt 2: diagnosis failed `56/100`, mitigation failed.
- Xhigh attempt 3 (with a small hint to check cluster-wide reconcilers and scheduled jobs): diagnosis passed `100/100`, mitigation passed.

The pattern was pretty consistent again: the agent found the `recommendation` OOM / bad `4Mi` memory limit, but very often missed the `vpa-updater` CronJob that keeps reapplying it. In the last run with the hint, it found `vpa-updater` from the cluster and fixed the durable source. 

So I think the problem is genuinely hard for the agent. `vpa-updater` feels like a blind spot for Codex rather than an issue with the problem itself.

Overall great work @iamyeezuschrist. Merging it!

iamyeezuschrist:
@munimthahmid thx for your great work and review!

and for the last commit btw.. i genuinely coded it worse, happy if the contribution makes any good for the benchmark.

tianyin:
@iamyeezuschrist It's good stuff! I'll reach out soon regarding the next step.

PR #909, opened 2026-06-24
Isolate problem IDs from agent runtime artifacts

### LINKED ISSUE #905: Benchmark problem ID leaks into agent container through environment variables and /logs
During a GPT-5.5 agent run on _calico_route_reflector_label_drift_hotel_reservation_, the agent submitted an incorrect diagnosis and received a score of 0.0. It then searched /logs and read:

Problem ID (harness): calico_route_reflector_label_drift_hotel_reservation

Immediately afterward, the agent said:

> The harness identifies this as a Calico route-reflector label drift scenario, which matches the symptoms: pods are Running, Services resolve, but pod-to-pod streams and MongoDB handshakes fail. I’m inspecting Calico BGP peer selectors and node labels now so the fix targets the routing control plane instead of cycling application pods.

It then inspected the relevant Calico resources and successfully mitigated the problem.

Although PR #844 removed the /get_problem HTTP endpoint, the problem ID is still passed into the agent container through SREGYM_PROBLEM_ID and written to agent-readable logs.
The agent can therefore obtain it from either the environment or /logs.

A correct fix should keep the real problem ID entirely in the host harness. Agent containers should use opaque run IDs or generic artifact names, with the host associating artifacts with the real problem after the agent exits.

munimthahmid:
Fixes #905 

This patch prevents evaluated agents from seeing the real benchmark problem ID during a run.

Changes:

* Replaces the real problem ID with an opaque ID such as anon_... during agent execution.
* Stores live artifacts under a temporary .runtime/.../anon_... path.
* Creates the real semantic result path only after the agent has stopped.
* Restores the real problem ID in final artifact filenames and structured CSV/JSON fields.
* Before publishing, verifies that temporary artifacts do not contain the real problem ID. If they do, publication fails and the artifacts are retained for investigation.
* Repairs container-created file ownership before publishing so the host can process the results.

Validation:

* Ran end-to-end canaries on Claude Code and Codex.
* The real problem ID was absent from the live agent environment, /logs path, driver logs, and session artifacts.
* Final semantic CSV/JSON outputs contained the real problem ID as expected.

### Another potential issue I noticed

* Mounting the full host ~/.codex directory into agent containers may expose historical sessions, logs, previous problem IDs etc. It is currently mounted for every agent, not only Codex. For Codex subscription authentication, we could instead create a private per-run Codex home containing only a copy of auth.json, mount it only into the Codex container, securely preserve any refreshed credentials if necessary, and remove the temporary directory after the run. API-key-based Codex runs and other agents would not need this mount.

/validate-problem skip


tianyin:
This is a very important and timely fix! Thank you @munimthahmid !

I skim the code and it looks fine.

And, given that you are the one who reported it, you probably have the right setup to validate it. 

Feel free to merge it. I approved.

tianyin (approved):
LGTM

tianyin:
> Another potential issue I noticed

the bigger question is whether there are other side channels but I guess the question is unanswerable and we will have to fix whenever we see.

tianyin:
@munimthahmid Feel free to merge it if you think it's ready.

Saadmrp1038:
@munimthahmid why not use a deterministic string to number conversion? Some kind of cryptography algo maybe.

munimthahmid:
@Saadmrp1038 Sorry, I missed the comment before merging :). 
The runtime ID doesn’t need to encode the real problem ID because the host already keeps the mapping and restores it after the run. We could use keyed deterministic encryption, but I feel that would add key management without providing a benefit over a fresh random ID.

Saadmrp1038:
I see. That's fine then.

> Mounting the full host ~/.codex directory into agent containers may expose historical sessions, logs, previous problem IDs etc.

Can you review this in the tlaps-bench PR please. I was thinking of just copying what we do there :3

PR #912, opened 2026-06-26
Fail alert oracle on Prometheus query errors

munimthahmid:
## Summary

This PR updates the benchmark-side `AlertOracle` to fail closed when querying Prometheus alerts fails.

Previously, `_query_firing_alerts()` caught Prometheus query errors and returned `[]`. Since `evaluate()` treats an empty list as “no firing alerts,” a Prometheus query failure could incorrectly mark mitigation as successful.

Now the oracle raises a `RuntimeError` instead. The existing conductor error handling records the mitigation result as `success=False` with an error message.

## Why

A failed Prometheus query is different from Prometheus being reachable with no firing alerts. The oracle should not pass mitigation without evidence that alerts are actually silent.


Saadmrp1038 (approved):

Saadmrp1038:
LGTM

PR #913, opened 2026-06-27
feat(problem): add limitrange_quota_exhaustion_hotel_reservation problem

kabirshahriar468:
## Description
Adds a new benchmark problem: `limitrange_quota_exhaustion_hotel_reservation`.

This problem introduces a deceptive, real-world multi-tenant governance failure. It simulates a scenario where a platform team applies restrictive namespace governance (`LimitRange` and `ResourceQuota`) without auditing the existing workloads, causing silent failures during pod recreation.

### The Failure Mechanism
1. **Invisible Defaults**: A `LimitRange` is created that silently injects default memory requests (`128Mi`) into the Hotel Reservation pods, which ship without explicit memory specifications.
2. **Quota Chokehold**: A `ResourceQuota` is applied that strictly caps aggregate namespace memory to exactly fit `N-1` pods (where N is the total number of running containers).
3. **The Trigger**: The `recommendation` pod is deleted. Current usage drops, but when the `ReplicaSet` controller attempts to recreate the pod, the API server rejects it with a `FailedCreate` event because the injected defaults leave zero headroom for the replacement.

### Why this is a valuable SREGym benchmark
This problem tests advanced diagnostic reasoning in LLM agents. It is a "Deceptive Failure" because:
- **No obvious pod logs:** The `recommendation` pod is never created, so there are no crash logs or `OOMKilled` statuses to read.
- **Hidden specs:** Inspecting the deployment YAML reveals no memory requests, meaning the agent cannot simply read the deployment spec to find the resource leak.
- **Diagnostic Leap:** To solve this, the agent must inspect `ReplicaSet` events, identify the quota rejection, and correlate it with the `LimitRange` admission controller rules that are injecting the invisible memory limits.

## Validation
I have run this locally via `cli.py` and verified:
- [x] Fault successfully injects without race conditions during rollout.
- [x] Unmitigated state correctly fails `submit()` (Mitigation Oracle).
- [x] Deleting the `ResourceQuota` and `LimitRange` allows the pod to recreate.
- [x] `submit(diagnosis)` passes both the Mitigation Oracle and the Diagnosis Oracle successfully after manual intervention.


Saadmrp1038:
@kabirshahriar468 Thanks for the PR!

Unfortunately, I think we have faults that are quite similar to this (`namespace_memory_limit`). These kinds of faults can be very quickly diagnosed by the agent by seeing the kubectl events which defeats the purpose of a hard fault. We are looking for faults that are hard for the agents to reason about or solve so that it will reveal the agents' strengths and weaknesses.

I am closing this PR. Please feel free to send another PR or preferably an RFC first to validate the fault.

PR #914, opened 2026-06-28
Agent internet access control via egress proxy

### LINKED ISSUE #911: Restrict internet access during agent evaluation
## Summary

The agent container is carefully isolated *inside* the cluster (filtered kubeconfig, hidden `chaos-mesh`/`khaos` namespaces, out-of-band grading), but it has **unrestricted outbound internet access**, and the Claude Code client enables `WebSearch`/`WebFetch`. Because the problem definitions, `root_cause` strings, fault injectors, and oracle logic all live in the **public** `SREGym/SREGym` repository, an agent can in principle fetch the exact expected answer for the problem it's being graded on — bypassing the cluster-side isolation entirely.

This is an evaluation-integrity gap, not a crash. Current leaderboard results implicitly assume agents do not look the benchmark up online; the harness does not enforce that assumption.

## Evidence (current `main`)

1. **No network egress restriction on the agent container.** `sregym/service/container_runner.py` launches the agent with `network_mode = "host"` (default, line ~26). On Linux `--network=host` shares the host network stack; on macOS the container falls back to its own netns but still has outbound NAT to the internet. There is no egress firewall, no DNS allowlist, no `--cap-drop`, and no offline flag anywhere in the launch path. At minimum, `Bash` → `curl`/`wget` reaches the public internet.

2. **Web tools are explicitly enabled for Claude Code.** `clients/claudecode/claudecode_agent.py` `ALLOWED_TOOLS` (line ~86) includes `"WebFetch"` (line ~94) and `"WebSearch"` (line ~103), passed to the CLI via `--allowedTools`. Codex has no dedicated web tool but runs with `--dangerously-bypass-approvals-and-sandbox` plus shell access, so it can reach the network as well.

3. **The answers are public.** Each problem's `root_cause` (the exact text the diagnosis `LLMAsAJudgeOracle` grades against), the fault injector, and the mitigation oracle are all in the open-source repo and trivially discoverable by problem ID.

## Impact

- A diagnosis submission can be informed by the verbatim expected `root_cause`. The grader is an LLM-as-judge over free text, so exact copy-paste isn't *automatically* rewarded, but an agent that has read the expected answer has a clear edge.
- Mitigation oracles only check cluster state, so foreknowledge of the intended fix shortcuts the investigation the benchmark is meant to measure.
- The effect is silent — nothing in the logs distinguishes "diagnosed it" from "looked it up."

## Reproduction (conceptual)

1. Run any agent with web/shell tools (e.g. `python main.py --agent claudecode --model <m> --problem <id>`).
2. Inside the container the agent can run e.g. `curl https://raw.githubusercontent.com/SREGym/SREGym/main/sregym/conductor/problems/<id>.py` and read the `root_cause` / injector directly.
3. The problem ID is available to the agent (`HARNESS_PROBLEM_ID_ENV` / task context), so the lookup is fully directed.

## Proposed direction: make internet access configurable (not removed)

Internet access and `WebSearch`/`WebFetch` are legitimately useful — agents should be able to consult upstream docs, CVE databases, vendor runbooks, etc., just like a real SRE. The goal is **control**, not removal. Suggest a tiered policy, configurable per run, with a sane default.

**Tier 1 — on/off flag (minimal):**
A `--internet-access {on,off}` flag (or `network_access` in `agents.yaml` / run config). `on` keeps today's behavior; `off` enforces an egress-locked container that can still reach only the in-cluster API endpoint and the model-provider API. Lets each run/leaderboard decide its own policy, with the keep-tools-available case preserved.

**Tier 2 — allow/deny lists (finer):**
A per-run egress policy with `allow` and `deny` domain lists, enforced at the network layer (so it covers `Bash`→`curl` too, not just `WebSearch`/`WebFetch`). For example:
```yaml
internet_access:
  mode: filtered            # off | on | filtered
  deny:                     # always blocked
    - github.com/SREGym
    - raw.githubusercontent.com/SREGym
    - "*.SREGym mirrors"
  allow:                    # optional; if set, only these are reachable
    - kubernetes.io
    - "*.datadoghq.com"
    - cve.mitre.org
```
Enforcement options: drop `network_mode: host` and route container egress through an HTTP(S) proxy / DNS filter that applies the lists. This keeps web tools fully available while making it impossible to fetch the benchmark's own answers.

**Default:** suggest shipping with the benchmark answers denied out of the box (at least `github.com/SREGym` + raw/mirrors on the deny list), so results are trustworthy by default while leaving the rest of the internet open. Teams that want a fully open or fully sandboxed run can flip the mode.

Keeping `WebSearch`/`WebFetch` enabled is fine under any of these — the control belongs at the **egress layer**, since an agent with `Bash` can `curl` regardless of which tools are in `ALLOWED_TOOLS`.

## Open questions

- Preferred granularity to start with — a simple `on/off` flag now, or go straight to allow/deny lists?
- Where should the policy live — a CLI flag, a field in `agents.yaml`, a separate run config, or all three?
- What default should the public leaderboard use (fully open, or benchmark-answers-denied)?
-

TamimEhsan:
- Closes #911

## Problem

Opus 4.8 was observed querying the SREGym GitHub repo during benchmark runs, reading problem definitions and root cause strings — bypassing  all cluster-side isolation. Any agent with shell access can curl regardless of ALLOWED_TOOLS, so enforcement must happen at the network layer.

## Solution

Adds a --internet-access {on,off,filtered} CLI flag (default: filtered) that runs a MITM proxy (mitmdump) on the host. The proxy is injected  into agent containers via http_proxy/https_proxy env vars.

  - filtered (default): blocks requests to SREGym GitHub paths (github.com/SREGym, raw.githubusercontent.com/SREGym,
  api.github.com/repos/SREGym, codeload.github.com/SREGym) while allowing all other traffic
  - off: blocks all outbound internet traffic
  - on: no proxy, unrestricted access (previous behavior)

  Uses mitmdump as a subprocess to intercept HTTPS and inspect full URL paths — not just hostnames. A combined CA bundle (system CAs + proxy CA)  is mounted into the container so HTTPS interception works without breaking API provider connections (Anthropic, OpenAI, Bedrock, etc.).

  Proxy env vars are also injected into non-containerized agent subprocesses, so the policy applies regardless of container_isolation setting.

## Files

  | File | Change |
  |------|--------|
  | `sregym/service/egress_proxy.py` | New — mitmdump-based MITM proxy with path-level deny list |
  | `main.py` | `--internet-access` CLI flag, proxy lifecycle |
  | `sregym/agent_launcher.py` | Proxy env injection for both container and bare-process agents |
  | `sregym/service/container_runner.py` | Proxy env vars + CA bundle mount in containers |
  | `sregym/conductor/conductor.py` | `internet_access` field on `ConductorConfig` |
  | `tests/service/test_egress_proxy.py` | 16 unit tests + 1 integration test |
## Prerequisites

  mitmdump must be installed when using filtered or off mode:
```
  uv tool install mitmproxy
```
## Test plan

  - [x] Unit tests pass: uv run pytest tests/service/test_egress_proxy.py -k "not integration" (16 tests)
  - [x] Integration test passes: uv run pytest tests/service/test_egress_proxy.py -m integration (starts real proxy, verifies SREGym blocked /
  kubernetes allowed)
  - [ ] Full run with --internet-access filtered — verify agent can reach API providers but not SREGym repo
  - [ ] Full run with --internet-access on — verify no behavioral change from current main

### DIFF
--- a/main.py
+++ b/main.py
@@ -30,6 +30,7 @@ from sregym.conductor.conductor_api import request_shutdown, run_api
 from sregym.conductor.constants import StartProblemResult
 from sregym.run_artifacts import ArtifactFinalizationError, RunArtifacts
 from sregym.service.container_runner import ContainerRunner, ExecInput
+from sregym.service.egress_proxy import DEFAULT_DENY_DOMAINS, EgressProxy
 
 LAUNCHER = AgentLauncher()
 logger = logging.getLogger(__name__)
@@ -493,6 +494,18 @@ def main(args):
 
     logger.info(f"🔧 Config — agent: {args.agent}, agent_model: {agent_model}, judge_model: {judge_model}")
 
+    # Start egress proxy if internet access is restricted
+    internet_access = args.internet_access
+    egress_proxy = None
+    if internet_access != "on":
+        deny = list(DEFAULT_DENY_DOMAINS) if internet_access == "filtered" else []
+        egress_proxy = EgressProxy(mode=internet_access, deny_domains=deny)
+        port = egress_proxy.start()
+        LAUNCHER.set_proxy(f"http://host.docker.internal:{port}", ca_bundle=egress_proxy.ca_bundle_path)
+        logger.info(f"🛡️ Internet access: {internet_access} (proxy port {port})")
+    else:
+        logger.info("🌐 Internet access: unrestricted")
+
     # Only build/check agent container image if the agent requires it
     agent_reg = (
         get_agent(args.agent, path=Path(os.path.dirname(os.path.abspath(__file__))) / "agents.yaml")
@@ -510,7 +523,11 @@ def main(args):
         install_script=agent_reg.install_script if agent_reg else None,
     )
 
-    conductor_config = ConductorConfig(deploy_loki=not args.use_external_harness, enable_noise=args.noise)
+    conductor_config = ConductorConfig(
+        deploy_loki=not args.use_external_harness,
+        enable_noise=args.noise,
+        internet_access=internet_access,
+    )
     conductor = Conductor(config=conductor_config)
 
     # Start the driver in the background; it will call request_shutdown() when finished
@@ -553,6 +570,14 @@ def main(args):
 
         # Give driver a moment to finish setting results
         driver_thread.join(timeout=5)
+
+        # Log blocked requests (read from file while proxy is still up),
+        # then stop the process and clean up temp files.
+        if egress_proxy:
+            for req in egress_proxy.blocked_requests():
+                logger.info(f"🚫 {req}")
+            egress_proxy.stop()
+            egress_proxy.cleanup()
 
     # When API shuts down, collect results from driver
     results = _driver_results
@@ -640,6 +665,13 @@ if __name__ == "__main__":
         type=str,
         default=None,
         help="Resume from a previous results CSV file. Problems already in the CSV will be skipped.",
+    )
+    parser.add_argument(
+        "--internet-access",
+        type=str,
+        choices=["on", "off", "filtered"],
+        default="filtered",
+        help="Agent internet access: 'on' (unrestricted), 'off' (cluster-only), 'filtered' (deny SREGym repo, default)",
     )
     args = parser.parse_args()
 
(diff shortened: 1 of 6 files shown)


Saadmrp1038:
Implemented in #968
Thanks for the initiative!

Issue #915, opened 2026-06-28
Support Local LLMs


Saadmrp1038:
We need to be able to support local LLMs like Qwen/Deepseek.
Most likely expose a local LLM API endpoint to the SREGym Docker environment.

Stratus might need some extension to support local LLMs. Our opencode agent does support local llms just need to expose the endpoint I believe. The other agents are fully commercial so I don't think they have support for local LLMs.

tianyin:
close; thanks @munimthahmid for the work!

PR #917, opened 2026-06-29
Support local LLM endpoints for Stratus and OpenCode

munimthahmid:
## Summary

This is the first step toward local LLM support in SREGym (#915). This PR adds local endpoint support for Stratus and OpenCode, so models such as Qwen or DeepSeek served through Ollama or other compatible local runtimes can be used by passing an API base URL.

## Changes

- Add `AGENT_API_BASE` / `AGENT_API_KEY` for the agent model endpoint
- Add `JUDGE_API_BASE` / `JUDGE_API_KEY` for the judge model endpoint
- Forward generic agent/judge endpoint env vars into the SREGym Docker environment
- Pass endpoint config into the shared LiteLLM backend
- Add per-run OpenCode config for local OpenAI-compatible endpoints
- Automatically route loopback agent endpoints into Docker on macOS and WSL
- Update Stratus preflight to use the same backend path as the agent and verify tool calling support

## Why update Stratus preflight?

Stratus depends on tool calls during normal execution. The previous preflight only checked a minimal text completion, which can pass even if the configured local model cannot return tool calls in the shape Stratus needs. The updated preflight catches that early before starting a benchmark run.

I think it makes the local LLM support safer to use. But open to discussion about this.

## Validation

- Stratus preflight passed with `gpt-5`
- Stratus preflight passed with local `ollama_chat/qwen2.5:3b`
- End-to-end agent run with `ollama_chat/qwen2.5:3b` entered diagnosis and executed `get_services`, then hit the run timeout
- End-to-end agent run with `ollama_chat/qwen3-1.7b-fast` produced a diagnosis after multiple tool calls, then failed in mitigation due to the model context limit
- OpenCode local endpoint calls passed with `local/qwen2.5:3b`; an end-to-end run connected successfully and exported its session, but the model did not solve the task
- Ruff passed
- `git diff --check` passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,8 +140,9 @@ Set the required environment variable for your provider before running:
 
 #### Local LLMs
 
-SREGym supports local models through Ollama and OpenAI-compatible servers such
-as vLLM and LM Studio. The examples below use Ollama.
+SREGym supports local models through Ollama and OpenAI-compatible servers such as vLLM and LM Studio. The examples below use Ollama.
+
+Set `AGENT_API_KEY` as well if the endpoint requires authentication.
 
 **Stratus with Ollama:**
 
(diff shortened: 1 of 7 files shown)


main.py
Saadmrp1038, line 1: I don't think the cli params are needed. In the readme we can just tell the user to pass these values via environment variables. Feels cleaner to me.

---

On the same note, README needs to be updated to add instructions regarding local LLMs.

Saadmrp1038 (changes requested):
Thanks @munimthahmid!

The preflight update is fine.

One thing though, what value did you use for `AGENT_API_BASE` in your testing? I'm asking because I expected a host-local endpoint (`localhost:11434`) to need some networking work to be reachable from inside the agent container. We hit a similar thing with `MCP_SERVER_PORT`. It worked on Linux but broke on WSL and macOS, and we had to rewrite to `host.docker.internal`.

```bash
        if self.config.network_mode == "host" and platform.system() == "Darwin" or self.config.network_mode == "host":
            env_vars["API_HOSTNAME"] = "host.docker.internal"
            mcp_port = env_vars.get("MCP_SERVER_PORT", os.environ.get("MCP_SERVER_PORT", "9954"))
            env_vars["MCP_SERVER_URL"] = f"http://host.docker.internal:{mcp_port}"
```

---
Also, it seems there's some logic duplication here...
Instead of  `if (host and Darwin) or host` we can simply do `if host` and update the comment. I think at some point we decided to just avoid the complexity and aways route to host.docker.internal . Can u please address this also?

Saadmrp1038:
@munimthahmid also can you check with opencode if the local LLM works please? I do want to add more open source agents so that people can test more smootly with local LLMs.

main.py
munimthahmid: Removed the cli params.

I tested with `AGENT_API_BASE=http://127.0.0.1:11434`. It works on my Linux since the agent container uses host networking.

I didn’t auto-rewrite this to `host.docker.internal` because that breaks my local Ollama setup (`host.docker.internal:11434` gives connection refused). For macOS/WSL, users can set `AGENT_API_BASE=http://host.docker.internal:11434`.

Lmk if you’d prefer some other approach.

munimthahmid: Also cleaned up the existing duplicated `network_mode == "host"` condition

munimthahmid:
@Saadmrp1038 Yes, I’m already looking into OpenCode next. SREGym’s OpenCode wrapper doesn’t actually support local LLMs out of the box yet, even though OpenCode upstream does.(ex. It rejects `ollama/qwen2.5:3b` as an unknown provider)

Saadmrp1038:
Interesting. Thanks for looking into it.

README.md
```diff
@@ -138,6 +138,36 @@ Set the required environment variable for your provider before running:
...
+export JUDGE_API_KEY="..."
+python main.py --agent stratus --model ollama_chat/qwen2.5:3b --judge-model gpt-5
+```
+
+On Linux host networking, `http://127.0.0.1:11434` can reach a local model
+server bound to loopback. On Docker Desktop for macOS or WSL, use
+`http://host.docker.internal:11434` if `127.0.0.1` resolves inside the agent
+container instead of the host. The local model server must listen on an address
+reachable from the Docker container.
+
```
Saadmrp1038: Let's handle it ourself instead of asking the user since it is quite easy to handle programmatically.

Saadmrp1038 (changes requested):

README.md
munimthahmid: Done.  Actually, I thought about it earlier, but then I felt that, as I can't really test on WSL or macOS, let's leave it up to the user. But you are right, it's quite straightforward, and we can handle this programmatically as well.

munimthahmid:
@Saadmrp1038 added local LLM support with OpenCode

Saadmrp1038 (commented):
@munimthahmid I think in current code logic `JUDGE_MODEL_ID` always gets passed to `LiteLLMBackend`. So, that means if someone passes a `provider/model` string in the `--judge-model` arg expecting to use local LLM as a judge it might get blocked due to `litellm` not recognizing it. Can you check this out? Ideal case is litellm will use that local llm as a judge. I am not sure if litellm needs something else to use local LLMs.

The edge case I am thinking about is someone runs opencode with local model using the `provider/model` format that opencode expects. If they just pass `--model` param that mode will also be used for the judge and the naming format might clash with litellm.

This could also be a nothing burger and maybe litellm is already compatible with opencode's naming format. Can you please test this out though?

munimthahmid:
@Saadmrp1038 your observation is correct. LiteLLM is not compatible with the OpenCode style format. OpenCode expects the local LLM model to be in `local/model` format, whereas LiteLLM expects it to be `provider/model`. And for this edge case specifically, I added this in the README:

`When --judge-model is not set, SREGym reuses the agent model and endpoint for the judge. This works directly for Stratus because its model identifier is LiteLLM-compatible. OpenCode's local/ provider is OpenCode-specific, so its judge must use a LiteLLM identifier such as ollama_chat/qwen3-coder:30b.`

However, if we want to handle this programmatically, one approach can be to normalize the `local/model` name to `openai/model` only for the judge path, since the OpenCode local endpoint is OpenAI-compatible. Then I think it should work in most scenarios.

Saadmrp1038:
@munimthahmid Sounds good. 
> if we want to handle this programmatically

Let's try to handle this programmatically if it's easy to do. Along with #922 user will also get fast feedback on wrong model string.

munimthahmid:
@Saadmrp1038 handled it programmatically now

Saadmrp1038 (approved):

Saadmrp1038:
LGTM

> os.environ["JUDGE_API_KEY"] = "dummy"

Also, I find it weird that litellm does not handle this natively.

PR #924, opened 2026-07-02
Preflight check the judge model

munimthahmid:
Preflight checks the judge model through LiteLLM before starting the benchmark, catching incompatible model configurations early. External-harness runs are excluded because they do not use the judge.


Saadmrp1038 (approved):

Saadmrp1038:
image

Issue #925, opened 2026-07-03
[Problem RFC] A Stuck 'ACCESS EXCLUSIVE' lock on product-catalog's table blocks all the reads (astronomy-shop)


saihariG:
## **Summary**

If one database session holds an **`ACCESS EXCLUSIVE`** lock on product-catalog's **`catalog.products`** table and doesn't release it, every read of that table queues behind the lock and never completes.

**product-catalog** and the four services that call it (frontend, checkout, recommendation, product-reviews) hangs. The product-catalog pod is **`Running`**, and Postgres itself is up with normal CPU and memory, but nothing crashes. The cause is only visible inside the database **(`pg_locks` / `pg_stat_activity`)**, which is why it isn’t easy to diagnose the issue.

## **Real-world Failure Story**

This is a common, well-documented Postgres failure. **GoCardless** had 15 seconds of API downtime. https://gocardless.com/blog/zero-downtime-postgres-migrations-the-hard-parts 

In order to add a foreign key constraint, Postgres takes `AccessExclusive` locks on both the table with the constraint, and the one it references while it adds the triggers which enforce the constraint. When a lock can't be acquired because of a lock held by another transaction, it goes into a queue. Any locks that conflict with the queued lock will queue up behind it. As `AccessExclusive` locks conflict with every other type of lock, having one sat in the queue blocks all other operations on that table.

This mechanism where a waiting **`ACCESS EXCLUSIVE`** lock blocks even non-conflicting reads, is documented:
- PostgreSQL manual: https://www.postgresql.org/docs/current/explicit-locking.html
- https://xata.io/blog/migrations-and-exclusive-locks

Also, GitLab treats this as a known hazard: https://gitlab.com/gitlab-org/gitlab/-/issues/38060

This is hard to diagnose because the database is not down and looks healthy the whole time. Responders tend to chase the wrong layer in a similar **_"database degraded, not down"_** incident, **ImprovMX** first tuned connection pools and caches before finding the real cause. Reference: https://improvmx.com/blog/post-mortem-2025-10-08  

## **How this differs from existing SREGym problems?**

_**astronomy-shop's**_ existing product-catalog problems make the service hard-fail (the flagd `productCatalogFailure` flag & `incorrect_image` problem). Existing database problems are auth/memory faults on valkey and mongodb. None of them covers Postgres lock contention, where the database stays up and healthy but every query on a table simply waits. 

## **Proposed simulation on SREGym**

[...]

PR #927, opened 2026-07-05
postgres_lock_contention_product_catalog problem (astronomy-shop)

saihariG:
Problem RFC: https://github.com/SREGym/SREGym/issues/925 

This is a common, well-documented Postgres failure. **GoCardless** had 15 seconds of API downtime. https://gocardless.com/blog/zero-downtime-postgres-migrations-the-hard-parts 

In order to add a foreign key constraint, Postgres takes `AccessExclusive` locks on both the table with the constraint, and the one it references while it adds the triggers which enforce the constraint. When a lock can't be acquired because of a lock held by another transaction, it goes into a queue. Any locks that conflict with the queued lock will queue up behind it. As `AccessExclusive` locks conflict with every other type of lock, having one sat in the queue blocks all other operations on that table.

This mechanism where a waiting **`ACCESS EXCLUSIVE`** lock blocks even non-conflicting reads, is documented:
- PostgreSQL manual: https://www.postgresql.org/docs/current/explicit-locking.html
- https://xata.io/blog/migrations-and-exclusive-locks

Also, GitLab treats this as a known hazard: https://gitlab.com/gitlab-org/gitlab/-/issues/38060

This is hard to diagnose because the database is not down and looks healthy the whole time. Responders tend to chase the wrong layer in a similar **_"database degraded, not down"_** incident, **ImprovMX** first tuned connection pools and caches before finding the real cause. Reference: https://improvmx.com/blog/post-mortem-2025-10-08  

App: **astronomy-shop** 
product-catalog reads from PostgreSQL table `catalog.products`.

- **`inject_fault()`:**
  
1. Raises the postgres memory limit (100Mi -> 512Mi). Under the fault, blocked reads pile up connections; at 100Mi Postgres OOM-kills within minutes, which would restart it and release the lock (the fault would self-destruct). 512Mi keeps it pure lock contention.
  2. Deploys a hidden Job (`pg_lock_holder.yaml`) that runs `BEGIN; LOCK TABLE catalog.products IN ACCESS EXCLUSIVE MODE; SELECT pg_sleep(...)` in a reconnecting loop, so the lock survives DB restarts.
  3. Waits until a read of `catalog.products` genuinely blocks before returning.
  
- **`recover_fault()`:** removes the lock-holder and **terminates the blocking backend** (`pg_terminate_backend`) - deleting the pod alone doesn't release the lock because the backend is parked in `pg_sleep`. (Same action a real SRE/agent must take.)

- **Mitigation oracle:** reads `catalog.products` with a short `lock_timeout`; accepts mitigation only when the read succeeds (a pod-health oracle can't detect this fault, since every pod stays Running).

- Product pages, recommendations, checkout, and product-reviews modules slows down or throws error
- The product-catalog pod stays Running ; The requests hangs & time-out and the logs show database query timeouts
- The PostgreSQL pod stays Running and healthy like normal CPU / memory
- In `pg_stat_activity`, 30 product-catalog `SELECT`s pile up with `wait_event_type=Lock` behind the lock holder.
Screenshot 2026-07-03 at 5 02 22 PM

- Naive fixes like restarting product-catalog won't work rather just makes the new pod wait on the same lock. Only releasing the lock works
- Validated end-to-end via `tests/integration/validate_problem.py` (deploy -> inject -> oracle fails -> recover -> oracle passes -> PASS)

## **4. Agent behavior**

Ran Stratus with two models. **Both failed**, each blamed a *different* app-layer but neither inspected `pg_locks`/`pg_stat_activity` or found the lock:

- **Gemini 2.5 Flash:** diagnosed a product-catalog, diagnosis judged incorrect.
- **Gemini 2.5 Pro:** diagnosed an OOM: *"product-catalog is crashing due to an out-of-memory error, memory limit set to 20Mi."* Diagnosis scored **11/100**. It then tried to mitigate by raising product-catalog's memory to 256Mi (never touching the database), mitigation failed, and it timed out (1800s).

The LLM judge's breakdown of the Pro run captures the intended difficulty:

> *"identifies product-catalog as the fault origin instead of the postgresql database"*
> *"identifies an OOM error, whereas the ground-truth mechanism is an ACCESS EXCLUSIVE lock"*
> *"claims the service is crashing due to OOM, but the ground truth states the service hangs while pods stay healthy"*

**_Therefore the "agents stop at the app layer and never look in the database"_**

## **Notes**
- Results above are Gemini 2.5 Flash/Pro - a different agent/model may fare differently.
- The lock-holder Job currently runs in the app namespace (visible to the agent). Neither model exploited this, but it could be moved to a hidden namespace to fully remove the shortcut.

/validate-problem postgres_lock_contention_product_catalog


sregym/conductor/problems/postgres_lock_contention_product_catalog.py
```diff
@@ -0,0 +1,197 @@
...
+        def released() -> bool:
+            self._terminate_lock_holder_backend()
+            return self._catalog_read_status() == "ok"
+
+        self._wait_until(released, timeout=120)
+
+        logger.info("Postgres lock-contention fault recovered.")
+        return True
+
+    """
```
SuMayaBee: This docstring is placed above `_terminate_lock_holder_backend` instead of inside it, so it never actually attaches to the function. `help()` or IDE hover won't show it. Same thing happens for `_set_postgres_memory` (line 139), `_namespace_exists` (line 152), `_catalog_read_status` (line 158), and `_wait_until` (line 185). Just needs to move to the first line inside each `def`.

sregym/conductor/problems/postgres_lock_contention_product_catalog.py
```diff
@@ -0,0 +1,197 @@
...
+
+    The fix is to end the blocking session (pg_terminate_backend on its pid) or remove the process
+    holding the lock.
+    """
+
+    POSTGRES_DEPLOY = "postgresql"
+    # postgres ships with a 100Mi limit and gets OOMKilled under the connection pile-up within minutes,
+    # which would restart it and release the lock (the fault would self-destruct).
+
+    # Raising the limit so the fault is pure lock contention. (We do NOT restore it on recovery: see recover_fault.)
```
SuMayaBee, line 36: The memory limit is raised to 512Mi here and never set back after recovery. Is the app redeployed fresh each run? If not, later runs on the same cluster could start with this higher limit already in place.

sregym/conductor/problems/postgres_lock_contention_product_catalog.py
```diff
@@ -0,0 +1,197 @@
...
+        self._set_postgres_memory(self.FAULT_MEMORY_LIMIT)
+        self.kubectl.exec_command(
+            f"kubectl rollout status deploy/{self.POSTGRES_DEPLOY} -n {self.namespace} --timeout=180s"
+        )
+
+        # Deploying the lock-holder (takes ACCESS EXCLUSIVE on catalog.products).
+        self.kubectl.apply_configs(self.namespace, self.manifest_path)
+
+        if not self._wait_until(lambda: self._catalog_read_status() == "locked", timeout=180):
+            logger.warning("Lock not confirmed held within timeout; fault may not be fully established.")
```
SuMayaBee: If the lock isn't confirmed within 180s, this just logs a warning and `inject_fault` still returns True below. Should this fail instead? Otherwise a broken fault injection could look like a normal run to the agent.

SuMayaBee (commented):
I like the problem, the "everything stays Running but reads just hang" angle is hard for agents to catch. Left some comments below.

sregym/conductor/problems/postgres_lock_contention_product_catalog.py
saihariG: Good One! This doesn't leak across runs. `astronomy-shop` is undeployed and redeployed fresh at the start of every run (both validate_problem.py and main.py do a full undeploy + redeploy), so postgres always starts at 100Mi. The 512Mi bump only lives for the duration of a single run.

We deliberately don't restore it in `recover_fault `because restoring the limit restarts postgres pod, which transiently wipes catalog.products and would break the mitigation oracle's post-recovery read. I've added a NOTE comment documenting this.

sregym/conductor/problems/postgres_lock_contention_product_catalog.py
saihariG: Fixed! Moved the docstring inside each of the five helper def's

sregym/conductor/problems/postgres_lock_contention_product_catalog.py
saihariG: **Fixed!** `inject_fault `now raises RuntimeError, so a failed injection fails loudly and the harness can catch/retry. 

**Re-validated:** It still passes. Lock is confirmed on a normal run, so the raise only fires on a genuine failure.

saihariG:
> I like the problem, the "everything stays Running but reads just hang" angle is hard for agents to catch. Left some comments below.

All three addressed, thanks for the review @SuMayaBee

sregym/conductor/oracles/postgres_lock_mitigation.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    def __init__(self, problem):
+        super().__init__(problem=problem)
+
+    def evaluate(self) -> dict:
+        print("--- Mitigation Evaluation (Postgres lock contention) ---")
+
+        status = self.problem._catalog_read_status()
+        if status == "ok":
```
munimthahmid: `pg_terminate_backend` kills only the current PostgreSQL session, while the Job reconnects after two seconds and reacquires the lock. This oracle can accept a successful read during that gap even though the fault immediately returns. Please remove the reconnect loop or require the database to remain readable beyond the reconnect interval.

sregym/conductor/problems/postgres_lock_contention_product_catalog.py
```diff
@@ -0,0 +1,197 @@
...
+
+        # Give postgres headroom so the connection pile-up behind the lock
+        # does not OOM-kill it.
+        self._set_postgres_memory(self.FAULT_MEMORY_LIMIT)
+        self.kubectl.exec_command(
+            f"kubectl rollout status deploy/{self.POSTGRES_DEPLOY} -n {self.namespace} --timeout=180s"
+        )
+
+        # Deploying the lock-holder (takes ACCESS EXCLUSIVE on catalog.products).
+        self.kubectl.apply_configs(self.namespace, self.manifest_path)
```
munimthahmid: This creates `pg-lock-holder` in the agent-visible app namespace, exposing the fault label and psql command. Please run this helper outside the agent’s Kubernetes credential scope, as required by the RFC.

munimthahmid (changes requested):
@saihariG , I left a couple of inline comments. Please address them when you have a chance.

sregym/conductor/oracles/postgres_lock_mitigation.py
saihariG: I removed the reconnect loop, the holder now acquires the lock once and holds it in a single session (`backoffLimit: 0`, `restartPolicy: Never`), so `pg_terminate_backend` on the blocking pid is a permanent fix and the fault no longer immediately returns. Also, I hardened the oracle, it now requires `catalog.products` to remain readable for 8s, well beyond the old 2s reconnect interval instead of accepting a single read.

sregym/conductor/problems/postgres_lock_contention_product_catalog.py
saihariG: Done. The lock-holder now deploys into the hidden `khaos` namespace rather than the app namespace, so it's outside the agent's Kubernetes credential scope. `k8s_proxy` filters `khaos/chaos-mesh` out of agent views, so the agent can no longer see or delete the helper (or read its label/psql command) and must mitigate through the database.

munimthahmid (approved):
@saihariG  I made a couple of small fixes to the fault setup and mitigation oracle. I also did two  end-to-end runs with Codex 5.6 Sol High.

Results:

- Attempt 1: diagnosis passed `89/100`, mitigation passed.
- Attempt 2: diagnosis passed `89/100`, mitigation passed.

The problem feels moderately hard. A strong model like Codex 5.6 Sol can investigate and uncover the underlying cause, but it still needs several diagnostic steps to trace the issue to PostgreSQL, identify the blocking session, and mitigate it correctly.

Overall great work @saihariG . Merging it!

PR #929, opened 2026-07-10
Extract standalone ATIF converter

### LINKED ISSUE #926: Convert the ATIF converter into a library
All the ATIF conversion code is already done. We just need to bundle it into a library. Sample usage should be:

```python
import atif_converter
atif_trajectory = atif_converter.convert(session_file)  # The file passed would be different for each agent
```

We would only need to tell the user which file they should pass to the function for each agent.

Saadmrp1038:
### Summary

- Add a copyable `atif_converter/` package with ATIF v1.7 models.
- Support Claude Code, Codex, Copilot, Gemini, OpenCode, and Stratus.
- Add automatic format detection and explicit agent overrides.
- Update SREGym to use the standalone converter directly.
- Preserve SREGym metadata, Claude multi-file handling, and Stratus stages.
- Remove the old duplicated adapters and models from `sregym/traces`.

Closes #926


tianyin (approved):
good stuff

PR #933, opened 2026-07-12
Chores: Redesign AlertOracle to ignore pre-existing alerts

### LINKED ISSUE #745: astronomy-shop Grafana sidecar trips ContainerCPUThrottling, breaking AlertOracle mitigation grading
## Issue

The OTel demo Helm chart shipped with `SREGym-applications/astronomy-shop` sets a `cpu: 100m` limit on Grafana's k8s-sidecar container. The sidecar's work is bursty (decoding ConfigMaps and reloading Grafana on each event), so it hits CFS throttling on every astronomy-shop deploy and during periodic re-syncs. This causes Prometheus to chronically fire `ContainerCPUThrottling` on the Grafana pod in the `astronomy-shop` namespace.

`AlertOracle` (`sregym/conductor/oracles/alert_oracle.py`) fails mitigation when any alert is firing in the problem's namespace. Because this alert is always firing, **every astronomy-shop problem graded by AlertOracle gets mitigation FAIL by default** — independent of whether the agent actually fixed the injected fault.

## Affected problems

The 11 astronomy-shop problems whose mitigation is graded by AlertOracle: ad_service_failure, ad_service_high_cpu, ad_service_image_slow_load, ad_service_manual_gc, cart_service_failure, failed_readiness_probe, kafka_queue_problems, loadgenerator_flood_homepage, payment_service_failure, payment_service_unreachable, product_catalog_service_failure.

## Not a cluster-size issue

CFS throttling is enforced per-container regardless of node capacity, so this affects any benchmark host running the chart as-is.

SuMayaBee:
## Problem

Our `astronomy-shop` has a chronic `ContainerCPUThrottling` alert from its Grafana sidecar (100m CPU limit). This alert fires all the time, unrelated to any agent action. Since `AlertOracle` fails the run if any alert is firing, every agent working on astronomy-shop failed automatically, regardless of what they did.

## Fix

Added `capture_baseline()` to `AlertOracle`. The conductor calls it right before the fault is injected, while the app is still healthy. It snapshots whatever alerts are already firing at that point and ignores them for the rest of the run. Any alert that appears later, after the fault, still fails the agent as before.

```mermaid
flowchart TD
    A["App healthy<br/>ContainerCPUThrottling already firing (chronic)"] --> B["capture_baseline()<br/>snapshot: firing alerts saved as baseline"]
    B --> C["Fault injected"]
    C --> D["Agent attempts mitigation"]
    D --> E{"AlertOracle checks<br/>currently firing alerts"}
    E -->|"alert is in baseline"| F["Ignored — pre-existing noise"]
    E -->|"alert is new"| G["FAIL — caused by agent's changes"]
    F --> H{"any non-baseline<br/>alerts still firing?"}
    H -->|"no"| I["PASS"]
    H -->|"yes"| G
```

## Verified on a live cluster

```
STEP 1 — alerts firing before baseline capture:
  ContainerCPUThrottling, DeploymentNotReady, HighRequestErrorRate,
  KubePodNotReady, PendingPodsDetected, PodSchedulingFailure, PodStatusError

STEP 2 — capture_baseline() runs:
  📋 AlertOracle baseline for astronomy-shop: ignoring pre-existing alerts
  [ContainerCPUThrottling, DeploymentNotReady, HighRequestErrorRate,
   KubePodNotReady, PendingPodsDetected, PodSchedulingFailure, PodStatusError]

STEP 3 — same alerts still firing, oracle checked again:
  (nothing firing — all baseline alerts filtered out)

RESULT: PASS
```

## Tests

4 new unit tests, all passing:
- pre-existing chronic alert is ignored
- a newly triggered alert still fails the run
- behavior is unchanged when no baseline is captured
- explicit `exclude_alerts` still works alongside the baseline

Closes SREGym#745


SuMayaBee:
I've resolved the merge conflict with upstream/main. Also bumped CodeQL CLI to v2.26.3 to fix the broken codeql-custom check.

tianyin:
@SuMayaBee You should really talk to @Saadmrp1038 (ping him on Slack). I talked to him about this PR and it seems that he wants to retire the alerts.

Saadmrp1038:
@tianyin she pinged me a few days back about this. She will notify me again after resolving the conflicts.

Saadmrp1038:
Oh already resolved!!! I will review it.

SuMayaBee:
> @tianyin she pinged me a few days back about this. She will notify me again after resolving the conflicts.

Yeah, the conflicts have been resolved, ready to review!

Saadmrp1038 (changes requested):
Thanks for the PR! I tested this and found a case where the oracle can incorrectly pass mitigation.

I first caused `ContainerCPUThrottling` on one pod and captured it in the baseline. I then caused the same alert on a second pod. Prometheus showed the alert firing for both pods, but the oracle returned no alerts because it filters using only the alert name.

This means a pre-existing alert can hide a new alert with the same name caused by the fault. We should track the full alert instance, including labels such as the pod or service, instead of only `alertname`.

---

On a side note, as Tianyin mentioned earlier, I am still unsure if we should add anything to patch the alert oracle at all. I need to think about this a bit, whether this is proper or not. I might shelve the whole alert oracle and refactor/disable all the faults related to it. I will need to investigate faults that depend on the alert oracle more before coming to a conclusion.

SuMayaBee:
I fixed this. Baseline now matches on the full alert label set, not just alertname, so a pre-existing alert on one pod doesn't mask a new one with the same name on a different pod. Added tests for both cases.

Saadmrp1038:
The PR looks more or less alright. I will do some more testing in 3-4 days and merge it then.

Saadmrp1038:
LGTM

Saadmrp1038 (approved):

Issue #867, opened 2026-06-05
[Problem RFC] PagerDuty-Style Kafka Producer Leak Exhausts Broker Memory


vsmart-06:
## Motivation

Real-world failures tied to well-known companies tend to produce the most compelling benchmark problems - they come with detailed postmortems, concrete timelines, and clear lessons. Kafka-related problems also seem underrepresented in the current problem set, with only one existing entry covering a completely different failure class (consumer lag).

### Real-World Failure Story

**Postmortem link:** https://www.pagerduty.com/eng/august-28-kafka-outages-what-happened-and-how-were-improving/

In August 2025, PagerDuty experienced a major platform outage caused by a Kafka producer leak. A new feature introduced a bug where a fresh Kafka producer was instantiated per API request rather than reusing a shared instance. At peak traffic this generated millions of producer connections per hour. The Kafka broker's JVM heap filled up tracking metadata for all these producers, triggering garbage collection thrashing and eventually full broker unavailability. Since Kafka was central to nearly every async operation at PagerDuty, the outage cascaded platform-wide. Critically, an initial mitigation that simply restarted the broker failed because the leaking producers immediately re-exhausted memory - the root cause had to be addressed directly.

## Proposed Simulation on SREGym

**App:** Astronomy Shop

The astronomy shop uses Kafka as the message bus between the checkout service (producer) and the accounting and fraud detection services (consumers). The Kafka broker runs with a constrained JVM heap (`-Xmx400M -Xms400M`) and a 600Mi memory limit, making it susceptible to memory exhaustion under abnormal producer load.

The fault injects a sidecar container into the checkout deployment. The checkout pod already has one container (checkout) and one init container (wait-for-kafka), and the checkout container already has KAFKA_ADDR: kafka:9092 set as an environment variable, confirming Kafka connectivity from within the pod. The sidecar uses this same address to continuously create new Kafka producer connections without reusing or closing them, mimicking the per-request producer instantiation bug from the PagerDuty incident.

The observable symptoms for an agent would be: initially checkout requests appear to succeed at the application level but orders do not appear in accounting or fraud detection as the broker becomes slow under memory pressure. As the producer leak continues and the broker's JVM heap is fully exhausted, checkout itself will also begin failing since it can no longer publish to Kafka at all. The Kafka broker pod will show memory pressure or get OOMKilled.

### Mitigation Oracle

[...]

PR #935, opened 2026-07-13
Kafka Producer Leak Causing Broker OOM on Astronomy Shop

vsmart-06:
Implements [#867](https://github.com/SREGym/SREGym/issues/867)

This PR adds a new SREGym problem, `kafka_producer_leak`, inspired by a real PagerDuty incident where excessive Kafka producer creation caused broker heap exhaustion.

## Real-World Failure Story
Postmortem link: https://www.pagerduty.com/eng/august-28-kafka-outages-what-happened-and-how-were-improving/

In August 2025, PagerDuty experienced a major platform outage caused by a Kafka producer leak. A new feature introduced a bug where a fresh Kafka producer was instantiated per API request rather than reusing a shared instance. At peak traffic this generated millions of producer connections per hour. The Kafka broker's JVM heap filled up tracking metadata for all these producers, triggering garbage collection thrashing and eventually full broker unavailability. Since Kafka was central to nearly every async operation at PagerDuty, the outage cascaded platform-wide. Critically, an initial mitigation that simply restarted the broker failed because the leaking producers immediately re-exhausted memory - the root cause had to be addressed directly.

## How the failure is simulated on SREGym
The problem targets the checkout deployment in the astronomy-shop namespace. 

### Fault injection
1. Patches the Kafka deployment to raise KAFKA_MESSAGE_MAX_BYTES to 20MB, allowing oversized messages through.
2. Adds an order-creator sidecar to the checkout deployment that spawns 20 threads, each creating a new idempotent Kafka producer per iteration and sending a 10MB payload to the orders topic. 

The concurrent large-message sends exhaust the broker's JVM heap through network receive buffer allocation pressure, triggering an OOMKill.

Recovery removes the sidecar and reverts KAFKA_MESSAGE_MAX_BYTES, then force-deletes the old checkout pod to break the rolling update deadlock caused by the broker being unavailable during the update.

## Problem runtime behavior and mitigation oracle
After injection, the Kafka broker OOMKills repeatedly (exit code 137). The same behaviour may be observed in the accounting and fraud-detection pods as well. The order-creator sidecar logs show hundreds of concurrent librdkafka producer instances terminating simultaneously with undelivered 10MB messages, directly implicating the sidecar as the cause.

The mitigation oracle (KafkaProducerLeakOracle) inherits MitigationOracle and additionally checks that the order-creator container is absent from the checkout deployment spec and that the broker pod is healthy.
The canonical mitigation is removing the order-creator sidecar from the checkout deployment.

## Notes

1. I definitely tested out the problem manually and I can see everything working, but I don't have access to any keys (that I'm aware of) to test it out with an actual agent. Any suggestions to test that out would be much appreciated! 
2. Right now I have a small Python script that creates the producers with the large payloads in the sidecar container. While investigated various approaches at implementing the problem, I found out that if I used a custom Java image instead, then the same outcome is observed, but the logs that come with the Python version are no longer visible, which may make the problem significantly more difficult for an agent to detect.

/validate-problem kafka_producer_leak


sregym/conductor/oracles/kafka_producer_leak_mitigation.py
```diff
@@ -0,0 +1,14 @@
...
+
+class KafkaProducerLeakOracle(MitigationOracle):
+    def evaluate(self) -> dict:
+        results = super().evaluate()
+
+        if results["success"]:
+            deployment = self.problem.kubectl.get_deployment(self.problem.faulty_service, self.problem.namespace)
+            for c in deployment.spec.template.spec.containers:
+                if c.name == "order-creator":
+                    results["success"] = False
```
Saadmrp1038: The inherited deployment baseline is empty because this oracle is constructed before the app is deployed. From my testing found the below unsafe mitigaiton attempts being passed:

- remove the sidecar and scale only `checkout` to 0 -> {'success': True}
- restore `checkout` and scale only `kafka` to 0 -> {'success': True}
- scale both to 0 -> {'success': True}

We should explicitly require both `checkout` and `kafka` to exist, request at least one replica, and have every desired replica Ready before accepting the mitigation.

---

Also, the oracle is too strict because it requires the agent to remove `order-creator` instead of checking whether the producer leak stopped. I fixed the `order-creator` sidecar and kept it. The broker was healthy, but the oracle still returned {'success': False} solely because container name matched.

---

Deleting the `checkout` Deployment raises `ApiException: 404 Not Found` here instead of returning {'success': False}. Please handle Kubernetes API errors and missing required resources as failed mitigation results.

sregym/generators/fault/inject_app.py
```diff
@@ -471,6 +471,87 @@ def recover_missing_env_variable(self, deployment_name: str, env_var: str, env_v
...
+
+        encoded = base64.b64encode(script.encode()).decode()
+
+        producer = client.V1Container(name="order-creator", image="python:3.12-slim", command=["sh", "-c", f"pip install confluent-kafka && python3 -u -c \"import base64; exec(base64.b64decode('{encoded}'))\""])
+
+        deployment.spec.template.spec.containers.append(producer)
+
+        self.kubectl.update_deployment(deployment_name, self.namespace, deployment)
+
+        print(f"Injected sidecar container 'order-creator' in '{deployment_name}'")
```
Saadmrp1038, line 562: The conductor starts the diagnosis stage immediately after this method returns. But it takes some time for Kafka OOM to appear. We should wait a little for the OOM to actually happen. Otherwise, initially the agent might see that everything is healthy and get confused.

Saadmrp1038 (changes requested):
@vsmart-06 Thanks for the effort! I left some inline comments below. Mainly about oracle behavior and fault injection handling. Please ping me when you have addressed them.

sregym/conductor/oracles/kafka_producer_leak_mitigation.py
vsmart-06: I'm not sure I understood what you meant by the inherited deployment baseline is empty because that basically had the functionality that would prevent those 3 unsafe mitigation attempts. Could you explain a bit more about what you meant there? I'll work on fixing the other issues though thanks!

Saadmrp1038: That logic would prevent these cases if it had captured the Deployments properly. The issue is when that capture happens. `MitigationOracle.__init__()` creates the baseline once.

*sregym/conductor/oracles/mitigation.py*
```python
class MitigationOracle(Oracle):
    importance = 1.0

    def __init__(self, problem):
        super().__init__(problem)
        deployments = self.problem.kubectl.list_deployments(self.problem.namespace)
        self.replica_count = {dep.metadata.name: dep.spec.replicas for dep in deployments.items}
```

`KafkaProducerLeakOracle` is instantiated inside the problem constructor before `Astronomy Shop` is deployed in the conductor lifecycle. So when the capture happens there's no Deployments in the namespace.

It’s actually a bug in our generic oracle implementation that I just noticed. Just handle it directly during `evaluate()` so it does not depend on that baseline for now. It should return failure if either Deployment is missing, scaled to zero, or does not have all desired replicas Ready.

I'll need to fix the baseline bug.

vsmart-06: I see alright, I'll try it out then. I was the one who introduced that addition in the default mitigation oracle so I guess my bad for that as well :(. How would we check that all the original deployments are still there in their original state then? And since you said that I can't identify the sidecar by its name, how do I make sure that the functionality is still missing? Because technically I think an incorrect mitigation could increase the heap limits for the broker and the fault would just not appear for a while and things would look good for all intents and purposes (this was something attempted in the actual PagerDuty issue as well). How do I make sure that the sidecar has truly been removed?

Saadmrp1038: To fix the baseline I think it should be captured after the app is fully deployed and Ready. Immediately before fault injection may be a good place. I’m not sure what the cleanest implementation is right now. So, checking `checkout` and `kafka` directly in `evaluate()` is also fine for this PR.

For Kafka, I don’t think the sidecar name needs to be checked. Capture the heap and memory limits before injection and reject increases. Another better way could be, observe Kafka for 60–120s to confirm it does not restart and its memory usage does not increase too much.

Saadmrp1038: These are just ideas from my quick understanding of the fault. If there are better ways to do it feel free to implement that!

sregym/generators/fault/inject_app.py
vsmart-06: I've added a sleep for 90s which should be enough for at least 1 OOMKill to register so this problem should be fixed

sregym/conductor/oracles/kafka_producer_leak_mitigation.py
vsmart-06: I've made some changes to make the mitigation oracle better and from my testing the problems you gave were fixed. My only issue is that in the mitigation oracle we currently wait 2 minutes for rollbacks, but this time period might have be to be increased to 5 minutes in case some of the pods are stuck in CrashLoopBackOff which caps at 5 minutes before it retries

sregym/generators/fault/inject_app.py
```diff
@@ -526,7 +535,11 @@ def task(thread_id: int):
 
         self.kubectl.update_deployment(deployment_name, self.namespace, deployment)
 
+        time.sleep(90)
```
Saadmrp1038: We should poll here.

sregym/conductor/oracles/kafka_producer_leak_mitigation.py
```diff
@@ -1,14 +1,67 @@
...
+            for p in pods.items:
+                if "kafka" in p.metadata.name:
+                    for c in p.status.container_statuses:
+                        if "kafka" in c.name:
+                            rcnt_1 = c.restart_count
+                            break
                     break
 
+            time.sleep(120)
+
```
Saadmrp1038: We should poll here. Then it would be safe to use a larger value.

Saadmrp1038 (changes requested):
Most of the issues I mentioned before are now fixed. Only two nits about the large sleep values remain. We should poll instead whenever possible.

Saadmrp1038:
Works great now!

Saadmrp1038 (approved):

Issue #937, opened 2026-07-17
Document hardware requirements / recommended cluster spec


dynomisto:
The README and the paper (arXiv:2605.07161) don't document minimum or recommended hardware specs for running the benchmark. We'd like to try SREGym on AWS and it's not clear what instance type and cluster topology to use.

It would be helpful to document:
- Minimum and recommended instance type / node spec
- Minimum number of worker nodes
- Recommended disk size (logs and Docker images accumulate)
- Any problems known to require more resources than the defaults
- Whether the CloudLab `c220g5` (used in the paper's evaluation) is the reference hardware

Context: the paper reports Claude Code uses ~1.47M tokens/problem on average, so a full run is a significant time and cost investment — knowing upfront what hardware is needed would help people plan their setup.

HacksonClark:
It does vary by problem -- most problems can be ran on a 16GB M2 MacBook pro! For the trainticket problems you would likely want a 32gb machine or so.

> Minimum number of worker nodes
We typically use 3 nodes, though most problems don't require a multi-node setup.

tianyin:
> We'd like to try SREGym

@dynomisto You can start with SREGym-Lite, which are 20 problems @Saadmrp1038 prepared.

https://github.com/SREGym/SREGym/issues/890#issuecomment-4931028656

@Saadmrp1038 can report the hardware configurations.

tianyin:
Let us know if you encounter problems running them.

Saadmrp1038:
@dynomisto Thanks for the interest!
I would recommend the SREGym-lite problem set for your run https://github.com/SREGym/SREGym/issues/890#issuecomment-4931028656. 

The easiest AWS setup is one ec2 instance running the `kind` setup. These 20 faults were selected so that they can be ran easily on kind so I don't see any reason to do kubernetes setup as that will cost way more. The SREGym-lite problem set does not have any problems using the `train-ticket` app so the VM requirements to run them are also quite low. My reccomendation will be: 

Minimum: 8 vCPU, 16 GB RAM, 100 GB disk (e.g., `c7a.2xlarge`)
Comfortable: 16 vCPU, 32 GB RAM, 200 GB disk (e.g., `c7a.4xlarge`)

The minimum should be enough imo. The latter if you want more headroom to run addtional stuff in the VM at the same time.

tianyin:
@Saadmrp1038 Could you put the SREGym-Lite on the `README.md` and also provide the hardware specs you just mentioned?

[...]

PR #939, opened 2026-07-18
Add SREGym-Lite benchmark suite

Saadmrp1038:
Adds SREGym-Lite as a well-tested, easy to run starting point for the benchmark.

ref #937


tianyin (approved):
Congrats for this excellent PR!

tianyin:
Let me merge it. I think we should start adding the results onto the leaderboard.

PR #942, opened 2026-07-19
fix(oracles): capture MitigationOracle baseline after the app is deployed

### LINKED ISSUE #941: MitigationOracle's replica-count guard never fires — baseline is snapshotted before the app is deployed
## Summary

`MitigationOracle` snapshots per-deployment replica counts to reject reward hacks
that delete a deployment or scale it to 0 (#874). The snapshot is always empty,
so all three replica checks are skipped.

## Cause

The oracle is constructed before the app exists:

```
conductor.py:403   get_problem_instance()  ->  MitigationOracle.__init__   [snapshot]
conductor.py:428   undeploy_app()          ->  tears down the namespace
conductor.py:431   deploy_app()            ->  app created
```

`list_namespaced_deployment` returns an empty list for an absent namespace rather
than raising, so `replica_count` stays `{}` and the guard loop at
[mitigation.py:54](https://github.com/SREGym/SREGym/blob/main/sregym/conductor/oracles/mitigation.py#L54)
iterates over nothing. Nothing errors — the feature is just inert.

## Impact

Only bites multi-deployment apps, which is the normal case. Zeroing one of
hotel-reservation's ~15 deployments leaves the rest `Running`, so the pod-liveness
check passes and the oracle reports a successful mitigation on a cluster that was
never fixed.

62 problem files construct `MitigationOracle`; three oracles subclass it and
inherit the defect. No test covers the base class.

## Reproduction

kind, k8s v1.35.0. Two healthy deployments, then
`kubectl scale deploy/web-b --replicas=0`. Same cluster state evaluated by two
oracles differing only in when they were constructed:

```
A  pre-deploy   replica_count={}                       success = True    <- hack accepted
B  post-deploy  replica_count={'web-a': 1, 'web-b': 1} success = False   <- "Deployment 'web-b' was scaled to 0"
```

## Proposed fix

Move the snapshot into a `capture_baseline()` hook called from
`Conductor._inject_fault()` — after deploy, before injection — with a no-op
default on `Oracle` so other oracles are unaffected.

Happy to send a PR; the change and tests are ready.

---

*AI assistance disclosure: I found this with Claude Code (Claude Opus 4.8), which
also drafted the fix and tests. The reproduction above was verified on a local
kind cluster.*

JamesWeng0428:
Fixes #941.

## What

`MitigationOracle` snapshotted per-deployment replica counts in `__init__`, which
runs before `deploy_app()`. The namespace didn't exist yet, so `replica_count`
stayed `{}` and the guards against "delete the deployment" and "scale to 0" never
fired.

Moves the snapshot into an explicit `capture_baseline()` hook, called from
`Conductor._inject_fault()` — the single choke point that runs after the app is
deployed and before the fault is injected. `Oracle` gets a no-op default, so
oracles needing no baseline are unaffected.

The snapshot logic itself is unchanged: the two lines that build `replica_count`
are byte-identical, just relocated. This is a timing change, not a rewrite of the
check.

## Verification

kind, k8s v1.35.0. Two healthy deployments, then
`kubectl scale deploy/web-b --replicas=0`:

```
before:  success = True     <- hack accepted
after:   success = False    <- "Deployment 'web-b' was scaled to 0"
```

11 new tests in `tests/oracles/test_mitigation_oracle_baseline.py`, run against
the unfixed code first to confirm they actually catch the defect (8 failed).
Three are load-bearing beyond coverage: two pin the conductor ordering (a
baseline captured *after* injection would record the broken state as healthy),
and one pins the original failure mode so the `capture_baseline()` call site
can't be quietly dropped again.

`tests/oracles/` + `tests/conductor/` (194) and `tests/problems/` (80) pass; ruff
clean.

## Reviewer notes

Two judgment calls I'd like a second opinion on:

1. `getattr(problem, "mitigation_oracle", None)` is defensive — the base
   `Problem` always sets the attribute, so a plain truthiness check would do.
   Happy to simplify.
2. The hook is only wired for `mitigation_oracle`. Detection and diagnosis
   oracles inherit the no-op and never get `capture_baseline()` called. Correct
   today since nothing else needs a baseline, but the asymmetry is deliberate
   rather than an oversight.

---

*AI assistance disclosure: this was found and drafted with Claude Code (Claude
Opus 4.8). The reproduction above was verified on a local kind cluster.*

### DIFF
--- a/sregym/conductor/conductor.py
+++ b/sregym/conductor/conductor.py
@@ -219,6 +219,13 @@ class Conductor:
     def _inject_fault(self):
         """Inject fault and prepare diagnosis checkpoint if available."""
         problem = self.current_problem
+
+        # Snapshot the healthy cluster before breaking it. The oracle is built
+        # in Problem.__init__, which runs before deploy_app(), so this is the
+        # first point at which the app actually exists.
+        if getattr(problem, "mitigation_oracle", None):
+            problem.mitigation_oracle.capture_baseline()
+
         problem.inject_fault()
         self.logger.info("[ENV] Injected fault")
         self.fault_injected = True
--- a/sregym/conductor/oracles/base.py
+++ b/sregym/conductor/oracles/base.py
@@ -7,6 +7,16 @@ class Oracle(ABC):
     def __init__(self, problem):
         self.problem = problem
 
+    def capture_baseline(self) -> None:
+        """Record the healthy pre-fault cluster state.
+
+        Called once the app is deployed and before the fault is injected.
+        Oracles cannot do this in __init__: the Problem is constructed before
+        deploy_app(), when the namespace does not exist yet. Defaults to a
+        no-op for oracles that need no baseline.
+        """
+        return
+
     @abstractmethod
     def evaluate(self, solution, trace, duration) -> dict:
         """Evaluate a solution."""
--- a/sregym/conductor/oracles/compound.py
+++ b/sregym/conductor/oracles/compound.py
@@ -18,6 +18,10 @@ class CompoundedOracle(Oracle):
                 raise ValueError(f"Duplicate oracle key: {key}")
             self.oracles[key] = oracle
 
+    def capture_baseline(self) -> None:
+        for oracle in self.oracles.values():
+            oracle.capture_baseline()
+
     def evaluate(self, *args, **kwargs):
         result = {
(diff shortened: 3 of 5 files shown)


Saadmrp1038:
LGTM

Saadmrp1038 (approved):

PR #946, opened 2026-07-28
Add `search_rate_retry_collapse_hotel_reservation` problem

Saadmrp1038:
This is a new **metastable** failure.

### Desired effect

> A temporary traffic burst ends, but the Hotel Reservation search path remains unavailable because queued requests and retries keep the system overloaded.

The Kubernetes resources remain Running and Ready. The failure is visible through `/hotels` request failures, a full downstream queue, and retry-amplified traffic between `search` and `rate`. `search` calls `rate` over gRPC to retrieve hotel prices. The `rate` service queues and processes these calls at a limited rate, while `search` retries calls that exceed their deadline.

The fault is inspired from the paper [Metastable Failures in Distributed Systems](https://sigops.org/s/conferences/hotos/2021/papers/hotos21-s11-bronson.pdf). Check section `2.1 Request Retries`. A slight change was made to the trigger condition to be a 10s traffic burst instead of 10-second network outage followed by retransmissions.

### What I Did

Implemented a metastable retry-collapse problem for Hotel Reservation:

1. Deploy `rate` with a capacity of 20 requests/s and a bounded queue of 256 requests.
2. Deploy `search` with a 750 ms timeout and up to three attempts for each rate request.
3. Establish a healthy baseline at 8 external search requests/s.
4. Temporarily increase traffic to 40 requests/s for 10 seconds.
5. Return external traffic to the original 8 requests/s.
6. Verify that the failure continues after the burst has ended.

During the burst, the rate queue fills with work. Requests time out at `search`, but the queued downstream work continues consuming capacity. Search retries then amplify 8 external requests/s into approximately 24 internal attempts/s, which remains above the rate service’s 20 requests/s capacity.

This keeps the queue full and the `/hotels` path unavailable even though the original traffic burst is gone.

### Mitigation

The ideal recovery:

1. Raises the rate-service capacity from 20 to 50 requests/s.
2. Reduces search attempts from three to two.
3. Rolls out the updated `rate` and `search` Deployments.

The mitigation oracle also accepts equivalent bounded fixes, such as reducing retry amplification or providing sufficient rate-service capacity.

### Mitigation Oracle

Added a fault-specific mitigation oracle that verifies:

- All original Deployments still exist and have current, Ready replicas.
- The required Services still have Ready endpoints.
- The `/hotels` path works under the normal 8 requests/s workload.
- Internal rate attempts are no longer significantly amplified.
- The downstream queue has drained.
- Rate and queue limits remain within the normal operating envelope.
- The system remains healthy after the temporary trigger is replayed.

The trigger replay rejects restart-only solutions: restarting `rate` may clear its in-memory queue, but the next burst reproduces the failure if the vulnerable policy remains unchanged.

| State | External traffic | Search success | Internal attempts/request | Queue depth |
|---|---:|---:|---:|---:|
| Healthy baseline | 8 requests/s | 100% | ~1.00 | 0 |
| After the temporary trigger | 8 requests/s | 0% | ~3.00 | 254–256 |
| After recovery | 8 requests/s | 100% | ~1.00 | 0 |

The following mitigation scenarios were also checked:

- No mitigation: rejected.
- Official recovery: passed.
- Retry-only mitigation: passed.
- Capacity-only mitigation: passed.
- Restart-only mitigation: rejected after trigger replay.
- Removing the explicit settings: rejected after trigger replay (this is because the fallback value can still cause the failure).
- Scaling to zero or replacing the serving topology: rejected.
- Setting effectively unbounded QPS or queue values: rejected.
- Repeated injection and recovery: passed.

### Agent Runs

The completed fault was run three times with each agent at maximum reasoning effort.

Agent | Diagnosis scores | Diagnosis | Mitigation | End-to-end
-- | -- | -- | -- | --
GPT-5.6 Sol Max | 89, 67, 78 | 2/3 | 3/3 | 2/3
Claude Opus 4.8 Max | 55, 66, 66 |0/3 | 0/3 | 0/3

All six runs localized the failure to the "search -> rate" path, but none fully figured out how the temporary burst created the persistent retry loop.

Sol selected bounded capacity increases of 100–200 requests/s, and all three mitigations survived trigger replay. Opus used capacities of 1,000–100,000 requests/s (this is way too high from initial value) or deleted the explicit settings and restarted. These mitigations were rejected as unsafe or reproduced the failure during trigger replay.

### Other Changes

- Added a problem-specific open-loop Hotel Reservation search workload.
- Added Prometheus scraping for the new search and rate metrics.
- Updated related Hotel Reservation faults to use the shared application image.
- Updated the application submodule to the merged retry-collapse implementation from SREGym/SREGym-applications#7.

/validate-problem search_rate_retry_collapse_hotel_reservation


munimthahmid:
@Saadmrp1038  Tested the problem end-to-end on a local four-node Kind cluster. The problem is Kind-compatible.


Observed behavior matched the PR description:

- Baseline: 8.00 req/s, 100% success, 1.00 internal attempt/request, queue depth 0.
- Post-trigger: 8.00 req/s, 0% success, ~3.00 attempts/request, queue depth 253–256.
- Official recovery: 100% success, 1.00 attempt/request, queue depth 0.
- Recovery survived a replayed ~40 req/s burst.

I also ran two Codex `gpt-5.6-sol` attempts at high reasoning effort:

| Attempt | Diagnosis | Mitigation | End-to-end |
|---|---|---|---|
| 1 | 89/100, pass | Fail | Fail |
| 2 | 45/100, fail | Fail | Fail |

Both agents localized the issue to the `search → rate` path and restored immediate traffic, but their mitigations were correctly rejected as unsafe:

- Attempt 1 selected 200 QPS with queue capacity 1024; the queue exceeded the safe maximum of 256.
- Attempt 2 selected 1000 QPS with queue capacity 4096; both exceeded the oracle's safe bounds.


Overall, the metastable failure reproduced reliably on Kind, and the mitigation oracle behaved as intended. This looks like a great Kind/SREGym-lite candidate.

munimthahmid (approved):
Reviewed the implementation as well. Everything looks correctly implemented and works as intended. No blocking issues from my side.

tianyin:
@Saadmrp1038 You may want to update `SREGym-Lite`.

PR #951, opened 2026-08-03
add: ungraceful_pod_termination_hotel_reservation problem

### LINKED ISSUE #950: [Problem RFC] Ungraceful pod termination drops in-flight requests — Discord March 2026 voice outage
## Real-world failure story

On March 25, 2026, Discord suffered a voice outage where pods were 
terminated before their stateful handoffs could complete. A safety 
check meant to wait for in-flight processes caused the 
terminationGracePeriodSeconds to elapse before handoffs could begin. 
Kubernetes sent SIGKILL, dropping active sessions and cascading into 
downstream failure.

Reference: https://discord.com/blog/behind-the-scenes-of-the-3-25-26-voice-outage

## How I plan to simulate on SREGym

Target: deployment/reservation in hotel-reservation

Injection:
- Set terminationGracePeriodSeconds: 1
- Add preStop hook: sleep 30 (simulating handoff process)
- Trigger rolling restart so fault is immediately observable

Symptom: pods are SIGKILL'd before preStop completes, dropping 
in-flight reservation requests. Pod events and error logs show 
abrupt terminations.

Recovery: restore original terminationGracePeriodSeconds, remove 
preStop hook.

## Mitigation oracle

Passes when:
1. terminationGracePeriodSeconds >= 30 on reservation deployment
2. A controlled test rollout completes without dropping all replicas
3. All pods Running and stable

AdetayoKalejaiye:
Closes #950

## Real-world failure story
On March 25, 2026, Discord suffered a voice outage where Kubernetes 
terminated pods before their stateful handoffs could complete. 
terminationGracePeriodSeconds elapsed before in-flight handoffs began, 
causing SIGKILL to drop active sessions and cascade into downstream failure.

Reference: https://discord.com/blog/behind-the-scenes-of-the-3-25-26-voice-outage

Target: deployment/reservation in hotel-reservation

- Patches terminationGracePeriodSeconds to 1
- Adds a preStop hook: sleep 30 (simulating handoff process)
- Triggers a rolling restart so fault is immediately observable

Pods are SIGKILL'd before preStop completes. Pod events and error logs 
show abrupt termination. wrk2 workload captures dropped reservation requests.

## Problem runtime behavior
- reservation pods cycle rapidly during rolling restart
- Pod events show Terminated reason with extremely short duration
- Dependent services (frontend, search) log connection errors
- All other pods remain Running/Ready — fault is scoped to reservation

## Mitigation oracle
Passes when:
1. terminationGracePeriodSeconds >= 30 on reservation deployment
2. A controlled test rollout completes without dropping all replicas
3. All pods Running and stable


munimthahmid (changes requested):
@AdetayoKalejaiye There are some major issues with the implementation.

The fault injection currently does not work. The injector uses `--type=merge` with a partial `containers` list, which removes the existing container image and other required fields. Kubernetes rejects both the injection and recovery patches with:

`spec.template.spec.containers[0].image: Required value`

`KubeCtl.exec_command()` suppresses this error, so the code continues with an ordinary rollout restart and reports that the fault was successfully injected, even though the grace period and `preStop` hook were never applied.

It feels like this was not tested well end-to-end, since inspecting the live Deployment after injection would have exposed this immediately.

There is also a fundamental mismatch with the Discord incident. Discord scaled down a multi-replica stateful session fleet, failed to complete real session handoffs, and triggered a large downstream cascade. This implementation restarts one reservation replica and adds a synthetic `preStop: sleep 30`. It does not reproduce the stateful handoff or downstream cascade.

I also tested the mitigation oracle against a `pause` container with no reservation application or draining behavior, and it still passed. The oracle only checks the grace-period value and Kubernetes availability, not request success or handoff completion.

Please redesign the problem, then test it end-to-end, including fault injection, mitigation, and recovery. If possible, also share an agent run.

AdetayoKalejaiye:
Thanks again for the thorough review. I've done a complete redesign and ran full end-to-end tests. Here's exactly what changed and the evidence for each concern:

**1. Patch corruption (--type=merge)**
Switched to strategic merge patch throughout, which merges container entries by `name` instead of replacing the list wholesale. Added `_verify_fault_applied()` to both injection and recovery — it re-fetches the live Deployment after every patch and raises loudly if the grace period, preStop hook, or image don't match expectations. This is the check that should have caught the original bug before it was ever pushed.

**2. Incident fidelity**
Redesigned the injection to match the Discord incident structure: scale the fleet to 4 replicas, apply the fault (terminationGracePeriodSeconds=1, preStop sleep=30s), wait for faulted pods to roll out, then scale 4→2 (50% simultaneous termination). Pods are SIGKILL'd 1s into their 30s preStop hook, dropping in-flight requests visible as connection resets during the scale-down event. The docstring is now honest about scope — this reproduces the core failure mechanism, not Discord's full session-handoff cascade.

**3. Oracle gameable with pause container**
This was the most involved fix. The original probe hit /hotels which routes through geo/rate/search — completely bypassing the reservation service. I redesigned the probe to exec directly into the frontend pod and curl reservation:8087. A pause container on reservation refuses the connection (exit 7); the real service connects but returns a gRPC protocol response (exit 1 or 2). The oracle distinguishes these correctly.

**Test results (screenshots attached):**

- **Test 1 — fault injection, oracle must fail:** Oracle correctly fails with `❌ terminationGracePeriodSeconds is 1s — must be at least 30s`. Confirms patch actually applied.
- **Test 2 — pause container, oracle must fail:** Oracle fails at pod stabilization — pause container never becomes Ready. Directly addresses your concern.
- **Test 3 — fixed deployment (grace=60s), oracle must pass:** Probe confirms reservation is serving traffic, rollout completes with 0% failure ratio. Oracle passes.
- **Test 4 — recovery:** Grace period restored to 30s, lifecycle cleared, spec matches original.
Screenshot 2026-08-04 201925
Screenshot 2026-08-04 201936
Screenshot 2026-08-04 202253
.

munimthahmid (changes requested):
@AdetayoKalejaiye  Thanks for the fixes. It does fix the original broken injection but there is still some major mismatch we need to address.


The main issue is that the new traffic check is not actually testing traffic during the rollout. It sends five requests before the rollout, waits for the rollout to finish, and then sends another five. So the reported "5/5 during rollout" is really five requests after the rollout.

The probe is also still very easy to fool. It treats anything other than exactly `EXIT:7` as success. I tried DNS failures, timeouts, missing curl, kubectl errors, and empty output, and they were all counted as successful. A random process listening on port 8087 would also pass.

The probe also bypasses the real application path. `probe_url` is unused, and the request goes directly to the Kubernetes reservation Service instead of through the frontend and Consul. Also, `/hotels` does call `reservation.CheckAvailability`, so it was already exercising reservation. The explanation that it completely bypasses reservation is not correct.

There is another implementation issue: the problem constructor calls `self.app.deploy()`. The harness creates the problem before cleanup and later deploys the app itself, so this causes a premature/double deployment.

More fundamentally, this is still not a high-fidelity simulation of the Discord incident. Discord had stateful processes, a real handoff, session loss, and a downstream cascade. Here we have a stateless reservation service and a synthetic `preStop: sleep 30`. Kubernetes runs `preStop` before SIGTERM, so this is not even the same shutdown sequence described in the incident. The scale-down is also a one-time event that is basically finished before the agent starts.

I don't think the simulation necessarily needs to reproduce Discord's entire downstream cascade, but it should at least reproduce the initiating mechanism with a real drain, handoff, or genuinely in-flight operation whose loss can be measured.

The oracle still hardcodes one specific fix by requiring `terminationGracePeriodSeconds >= 30`. That would reject other valid mitigations, including Discord's actual mitigation of blocking scale-down until handoff completes. Also, a 30-second grace period is not safely enough for a 30-second hook because the application still needs time to terminate afterward.


Can you please address these issues and send another commit? If everything looks good I will then start doing some agent runs.

AdetayoKalejaiye:
Thanks again — all five points were correct and I've done another full redesign.

**1. Shutdown sequence (preStop vs SIGTERM trap)**
You were right that preStop runs before SIGTERM, which is the wrong sequence. The fault now wraps the container command in a shell supervisor with a SIGTERM trap. On SIGTERM, the wrapper logs 'session handoff started', keeps the server running for 45s (still serving in-flight requests), then stops it cleanly. With grace=1s, SIGKILL fires 1s into that 45s wait — matching the incident exactly. Pod logs show 'handoff started' with no 'handoff complete', which is direct evidence of the failure mode.

**2. Oracle: HTTP probe can't see dropped requests**
Also correct — Kubernetes removes the pod from Endpoints before or simultaneously with SIGTERM, so the load balancer routes traffic away from the dying pod before it starts draining. No HTTP probe can see the drop. Replaced with termination duration measurement: poll the pod lifecycle API from Terminating → disappeared. With grace=1s, the pod dies in ~2-4s. With grace=60s and the 45s handoff running to completion, it lives ~47s. These are unambiguously distinguishable without logs, HTTP probes, or anything that could be affected by GC or endpoint routing.

**3. Log availability**
Dropped entirely as a mechanism. No `--previous`, no log buffer dependency.

**4. Baseline check preserved**
Still runs before the termination test as an app-identity gate — requires 5 exact HTTP 200s through frontend /hotels (which calls reservation.CheckAvailability). Pause containers, crash-looping pods, wrong images all fail here before the timing test runs, so a short termination duration is unambiguous: it means SIGKILL, not a misconfigured deployment.

**5. Churn suspend hardening**
After patching `suspend=true`, the oracle re-fetches the CronJob and verifies `spec.suspend == true` before polling `.status.active`. If the patch didn't land, it logs a warning and returns rather than silently proceeding. If the 120s wait for active jobs times out, it logs explicitly instead of falling through.

**6. Double deployment**
Reverted to `create_workload()` — harness owns deployment.

**Test results (screenshots attached):**
- Step 1: fault injected, live spec confirmed (grace=1, wrapper present, image intact), incomplete-handoff evidence in pod logs
- Step 2: oracle fails with 3.9s termination duration (threshold 35s) [](url)
- Step 3: grace raised to 60s, oracle passes with 46.7s duration — handoff ran to completion ✅ 
- Step 4: pause container rejected at baseline (no HTTP 200) 
- Step 5: recovery restores grace=30, wrapper gone, churn RBAC cleaned up 
Screenshot 2026-08-06 173739
Screenshot 2026-08-06 173747
Screenshot 2026-08-06 173756
Screenshot 2026-08-06 173810
Screenshot 2026-08-06 173834

munimthahmid (changes requested):
Thanks for the changes, @AdetayoKalejaiye . Unfortunately, I still think the implementation is very different from the real Discord incident. I think you should study the real incident a bit more which will give you more clarity.

In the real incident, pods had live sessions that needed to move to another pod during shutdown. That handoff failed, users lost their sessions and reconnected, and those reconnects put extra load on other services. In this implementation, there are no live sessions and nothing is actually moved or lost. The pod only waits and gets killed, while the other pod keeps the service running normally. The check also only looks at how long the pod takes to delete, so it does not check the actual failure or its impact.

I suggest adding actual user sessions that must move between pods during shutdown. A short shutdown time should interrupt that handoff, lose sessions, trigger reconnects, and increase load. The checks should confirm those problems happen before the fix and disappear after it. Right now, this only simulates a pod being force-killed, not the Discord incident.

AdetayoKalejaiye:
Would it make sense to simulate the downstream impact by having the fault injector artificially trigger a massive spike in generic wrk2 traffic whenever a pod is SIGKILL'd, mimicking the load of a reconnect storm?

munimthahmid:
I don’t think so, @AdetayoKalejaiye. The reconnect storm in Discord was caused by actual sessions being lost and the affected users reconnecting. Having the fault injector start generic `wrk2` traffic would only create an artificial load spike. It may look similar in the metrics, but it would not reproduce the actual failure.

The bigger issue is that Hotel Reservation does not have live sessions that belong to specific pods and need to move during shutdown. None of the current SREGym applications has that behavior. It is possible to add a real session system and long-lived clients, but that would require a complete redesign rather than another change to Hotel Reservation.

Thanks a lot for your dedication and hard work on this. Given the amount of redesign needed, I think I will need to close this PR now. I highly encourage you to choose another interesting incident that fits one of the existing SREGym applications, and I will review it as soon as possible.

PR #953, opened 2026-08-06
Fix the mitigation oracle for `duplicate_pvc_mounts_social_network`

Saadmrp1038:
### Bug

The observability setup changes the Jaeger Service to an ExternalName Service. Kubernetes does not create Endpoints for this type of Service. However, the old oracle required Jaeger Endpoints. As a result, the oracle rejected a correct mitigation even when the current Jaeger pod was Ready and served valid queries.

### Fix

The oracle now finds a Ready Jaeger pod that belongs to an active ReplicaSet. It sends the Jaeger query directly to the pod IP address.


tianyin (approved):
LGTM

Issue #928, opened 2026-07-09
[Problem RFC] A missing index on product-catalog's catalog.products forces sequential scans and degrades every read (astronomy-shop)


Ashish301101:
## Summary
If the index backing product-catalog's read on `catalog.products` is dropped (e.g. a migration recreates the table or forgets to re-add the index), every `GetProduct` lookup (`SELECT ... FROM catalog.products WHERE id = `) switches from an **Index Scan** to a full
**Sequential Scan**. On a production-sized catalog this saturates PostgreSQL CPU and makes product-catalog and the services that call it (frontend, checkout, recommendation, product-reviews) slow to a crawl and time out. The product-catalog pod stays `Running`, Postgres itself is up with normal-looking health, and nothing crashes. The cause is only visible inside
the database (`EXPLAIN`, `pg_stat_statements`), which is why it isn't easy to diagnose.

## Real-world Failure Story
A missing/dropped index on a hot column is one of the most common documented database failures.
**Samsara** had recurring outages where deploying code with an unindexed query caused full table scans that "took down a database" — until they added a CI check to catch it.
https://medium.com/samsara-engineering/how-we-prevented-missing-sql-indexes-5eb4741a9a9a

The pattern is well documented:
* Sentry — how missing indexes silently degrade production databases: https://blog.sentry.io/missing-indexes-are-slowing-down-your-database-heres-how-to-find-and-fix/
* PostgreSQL manual — Index vs Sequential Scan: https://www.postgresql.org/docs/current/using-explain.html

This is a database degraded, not down incident where responders chase the wrong layer. **Clerk's** Feb 2026 post-mortem is a good example: a query going from a fast plan to a slow plan collapsed DB performance and returned 429s for ~95% of traffic for ~90 minutes before they found the real cause. 
Reference: https://clerk.com/blog/2026-02-19-system-outage-postmortem

## How this differs from existing SREGym problems?
[...]

PR #959, opened 2026-08-15
Add `product_catalog_missing_index_astronomy_shop` problem #928

Ashish301101:
Problem RFC: https://github.com/SREGym/SREGym/issues/928

An index behind a hot read disappears — a migration recreates the table, an ORM auto-migration drops and never re-adds it, or a restore rebuilds the schema without its indexes. Postgres does not error. The planner simply falls back to a sequential scan and keeps returning correct results.

This is hard to diagnose for the same reason it is common:

- **Nothing fails.** No crash, no restart, no probe failure, no error in any log. The queries return the right rows, just slowly.
- **It passes staging.** A Seq Scan over a few hundred rows is fast. The regression scales with table size, so it only hurts once production data volume is behind it.
- **The evidence is inside the database.** From Kubernetes everything looks healthy. You have to run `EXPLAIN` or read `pg_stat_user_tables` to see it.

PostgreSQL manual on the planner's index-vs-seqscan choice: https://www.postgresql.org/docs/current/indexes-examine.html
`EXPLAIN` and reading query plans: https://www.postgresql.org/docs/current/using-explain.html

App: **astronomy-shop**
product-catalog's `GetProduct` handler runs exactly one statement per product-detail view against `catalog.products`:

```sql
SELECT p.id, p.name, p.description, p.picture,
       p.price_currency_code, p.price_units, p.price_nanos, p.categories
FROM catalog.products p
WHERE p.id = $1
```

- **`inject_fault()`:**

1. Raises the postgres memory limit (100Mi -> 512Mi). Bulk-seeding OOM-kills Postgres at 100Mi (a single 200k-row INSERT was killed with exit 137), and the restart re-runs the chart's init scripts, which recreate `catalog.products` with its ten original rows *and* its primary key — the fault would self-destruct.
2. Seeds the catalog to 200,000 rows by cloning the genuine rows round-robin, so the seeded catalog keeps the same spread of names, prices and descriptions.
3. Asserts the planner picks an **Index Scan** at that size, drops the index, then asserts it now picks a **Seq Scan**. Injection fails loudly rather than producing a silent no-op.

- **`recover_fault()`:** recreates the index and re-runs `ANALYZE`. It deliberately leaves the seeded rows in place — deleting them returns the table to ten rows, at which size the planner correctly prefers a Seq Scan *even with the index present*, and the oracle would report failure immediately after a correct recovery.

- **Mitigation oracle:** checks the index exists and the plan is an Index Scan, with the plan check gated behind a row-count threshold so an emptied table cannot satisfy it. Also asserts the product-catalog deployment still exists, to reject "fixes" that delete the service (a pod-health oracle can't detect this fault, since every pod stays Running).

- Product pages, recommendations, checkout and product-reviews slow down; nothing errors and every response is still correct
- Measured on the target cluster at 200k rows / ~105 MB:

| | Plan | Latency |
|---|---|---|
| with `products_pkey` | Index Scan | 0.125 ms |
| without `products_pkey` | Parallel Seq Scan | 29.5–40.3 ms (~240–320×) |

- Every pod stays `Running` and healthy — product-catalog and PostgreSQL alike. No restarts, no OOMKills, no probe failures
- product-catalog burns **~483m CPU while otherwise idle** — the clearest signal visible from outside the database
- `EXPLAIN` shows a Seq Scan on `catalog.products`; `pg_stat_user_tables` shows `seq_scan` / `seq_tup_read` climbing while `idx_scan` stays flat
- Naive fixes don't work: restarting product-catalog or raising its memory changes nothing, because the regression is in the query plan. Only recreating the index works
- Validated end-to-end via `tests/integration/validate_problem.py` (deploy -> inject -> oracle fails -> recover -> oracle passes -> **PASS**, 6/6 stages in 257s)

## **4. Agent behavior**

Ran Claude Code with `claude-sonnet-5`, two runs. **Both failed**, and both stopped at the container layer rather than the database:

- **Run 1:** never contacted the database at all — 0 commands containing `psql`, `EXPLAIN` or `catalog.products`. Diagnosed `ad` being OOMKilled at its chart limit of 300Mi, patched it to 600Mi, submitted. Scored **0.00**.
- **Run 2:** did reach the database (28 `psql` invocations, queried `pg_indexes` on the catalog schema, spawned its own `postgres:17-alpine` pod to get around auth) but still concluded the root cause was product-catalog's stock 20Mi limit / `GOMEMLIMIT=16MiB`. Scored **0.00**.

The LLM judge's breakdown of run 2 captures the intended difficulty:

> *"Ground truth identifies `postgresql` as the origin, while the agent blames `product-catalog` service."*
> *"Ground truth describes a missing database index, while the agent identifies an under-provisioned memory limit causing OOMKills."*
> *"Ground truth describes latency and healthy pods, while the agent describes CrashLoopBackOff, OOMKilled, and gRPC errors."*

In both runs the agent observed the elevated CPU on product-catalog and attributed it to GC pressure from the small memory limit rather than to query work.

**_Therefore the "agents stop at the container layer and explain away CPU as a resource-limit problem"_**

## **Notes**

- Results above are `claude-sonnet-5` via Claude Code — a different agent/model may fare differently.
- Both agent runs anchored on pre-existing chart configuration (`ad` at 300Mi, product-catalog at 20Mi), not on anything this problem injects. A controlled before/after test — same deployment, product-catalog sampled every 15s under live load across two 5-minute windows — confirmed the fault causes **zero** restarts, OOMKills or crash loops (`restart_delta=0, oom_samples=0, crashloop_samples=0` in both windows). Flagging in case maintainers consider those chart defaults worth revisiting, since they affect every astronomy-shop problem.


HacksonClark:
Sorry for hijacking this PR to test my new bot :P

Ashish301101:
@Saadmrp1038 Request to review this PR

sregym/conductor/problems/product_catalog_missing_index.py
```diff
@@ -0,0 +1,349 @@
...
+    def _column_names(self) -> list[str]:
+        out = self._psql(
+            "SELECT column_name FROM information_schema.columns "
+            f"WHERE table_schema = '{self.schema}' AND table_name = '{self.table}' "
+            "ORDER BY ordinal_position;",
+            tuples_only=True,
+        )
+        return [line.strip() for line in out.strip().splitlines() if line.strip() and not _sql_failed(line)]
+
+    def _seed_batch(self, columns: list[str], start: int, end: int) -> None:
```
Saadmrp1038, line 195: `_seed_batch()` adds 200,000 visible products to the real catalog. Testing showed that normal traffic OOM-killed the 20Mi `product-catalog` container. Both product endpoints returned HTTP 500. This creates a memory failure instead of the documented silent database slowdown.

sregym/conductor/problems/product_catalog_missing_index.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def explain_getproduct_plan(self) -> str:
+        return self._psql(f"EXPLAIN SELECT * FROM {self._qualified} p WHERE p.{self.id_column} = 'OLJCESPC7Z';")
+
+    @staticmethod
+    def plan_uses_index_scan(plan: str) -> bool:
+        lowered = plan.lower()
+        return ("index scan" in lowered or "index only scan" in lowered) and "seq scan" not in lowered
+
+    def product_catalog_deployment_present(self) -> bool:
```
Saadmrp1038, line 294: This function only checks that the Deployment requests a replica. It does not check that the current rollout is Ready or not. During testing recovery passed while `product-catalog` was `0/1 CrashLoopBackOff` and returned HTTP 500. An invalid-image rollout also passed because the old pod remained Ready. Please require a complete current rollout. Then request known product data through the frontend API to make sure it's working.

sregym/conductor/oracles/product_catalog_missing_index_mitigation.py
```diff
@@ -0,0 +1,67 @@
...
+
+        if results["row_count"] >= problem.PLAN_SIGNIFICANT_ROWS and not results["uses_index_scan"]:
+            results["reason"] = (
+                f"an index on catalog.products(id) exists, but GetProduct still plans a Seq Scan over "
+                f"{results['row_count']} rows"
+            )
+            print("Mitigation Result: Fail")
+            return results
+
+        results["success"] = True
```
Saadmrp1038, line 64: This condition skips the query-plan check when fewer than 1,000 rows remain. After `TRUNCATE catalog.products`, the oracle returned `success=True` with `row_count=0` and `uses_index_scan=False`. The frontend also returned HTTP 500. Please test known product data through the frontend API. Also, an empty or reduced table must always fail.

sregym/conductor/oracles/product_catalog_missing_index_mitigation.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+
+import logging
+
+from sregym.conductor.oracles.base import Oracle
+
+logger = logging.getLogger(__name__)
+
+
+class ProductCatalogMissingIndexMitigationOracle(Oracle):
```
Saadmrp1038, line 11: This class inherits from `Oracle`, so baseline capture records no application Deployments. During testing scaled `checkout` to 0, but the oracle still returned `success=True`. Please inherit from `MitigationOracle` and call `super().evaluate()` first. This checks that each original Deployment still exists, requests at least one replica, and has all requested replicas Ready. Then run this fault’s database and frontend checks.

Saadmrp1038 (changes requested):
@Ashish301101 Thanks for the PR.

I left some inline comments below. The missing-index idea is interesting. However, the current injection causes an OOM crash instead of a silent slowdown. The mitigation oracle also accepts several broken states. I found some issues below, but there may be others.

Please ping me after you address these problems. Please also test the mitigation oracle more thoroughly for other bypasses.

The generated data is also clearly artificial. It repeats the same 10 products until the table contains 200,000 rows, with only the IDs changed. An agent that inspects the table can identify these repeated rows as the fault instead of diagnosing the real issue.

HacksonClark:
/evaluate-problem product_catalog_missing_index_astronomy_shop

HacksonClark:
/evaluate-problem product_catalog_missing_index_astronomy_shop

HacksonClark:
/evaluate-problem product_catalog_missing_index_astronomy_shop

HacksonClark (changes requested):
There's a bug in the mitigation oracle that can be reward hacked: The mitigation oracle has a false positive: an attempt passed after PostgreSQL reset the catalog to 10 rows, even though EXPLAIN still showed a Seq Scan. The oracle skips its check below 1,000 rows.

This does seem like a challenging problem though! With some fixes I think it will make a great addition to the benchmark :)

PR #960, opened 2026-08-17
feat(benchmark): Add stale_hostaliases_dns_poisoning_astronomy_shop problem

Talhaasif7:
This PR adds a new SREGym benchmark problem: `stale_hostaliases_dns_poisoning_astronomy_shop`.

It reproduces a real-world Kubernetes DNS configuration drift failure where a pod-local `hostAliases` entry shadows in-cluster DNS resolution for a specific microservice (`frontend`), causing internal calls to its backend (`product-catalog`) to fail with connection refused errors while all cluster-wide control plane and DNS signals remain completely green.

---

In production Kubernetes environments, engineering teams frequently introduce static `/etc/hosts` entries (`hostAliases` in the Deployment spec) during migrations, local debugging, or legacy endpoint redirect testing. Over time, these entries are forgotten and become stale.

Because the Linux C library (GLIBC/musl) resolver reads `/etc/hosts` before querying CoreDNS:
- The `hostAliases` mapping takes absolute precedence over CoreDNS for that Deployment's pods.
- The failure is strictly pod-local: all other services in the cluster resolve `product-catalog` properly via CoreDNS.
- Control plane signals stay 100% green: CoreDNS is healthy, pods report `Running` and `Ready`, backend Services have valid endpoints, yet application-level RPCs fail continuously.

---

The problem implementation lives in:
- `sregym/conductor/problems/stale_hostaliases_dns_poisoning_astronomy_shop.py`
- `sregym/conductor/oracles/stale_hostaliases_mitigation.py`
- `sregym/generators/fault/inject_virtual.py`

### Fault Injection:
The fault injector patches the `frontend` Deployment spec in the `astronomy-shop` namespace with a loopback `hostAliases` entry:

```yaml
spec:
  template:
    spec:
      hostAliases:
        - ip: "127.0.0.1"
          hostnames:
            - "product-catalog"
```

### Impact & Root Cause:
The kubelet writes `127.0.0.1 product-catalog` into the `frontend` container's `/etc/hosts`. When the `frontend` service attempts to fetch product catalog data via `PRODUCT_CATALOG_ADDR=product-catalog:8080`, requests loop back to `127.0.0.1:8080` (the frontend container itself). Since no server is bound to loopback port `8080`, every request fails with `ECONNREFUSED` / connection refused errors.

---

## 3. Runtime Behavior and Investigation
During the failure state:
1. `frontend` pods report `Running` (1/1 Ready).
2. CoreDNS and `product-catalog` deployment remain fully healthy.
3. `nslookup product-catalog` from any *other* pod resolves the correct ClusterIP.
4. `kubectl get deployment frontend -o yaml` reveals the stale `hostAliases` block under `.spec.template.spec`.

### Investigation Commands:
```bash
# 1. Inspect frontend logs to observe connection refused errors to product-catalog:8080
kubectl logs -n astronomy-shop -l app.kubernetes.io/name=frontend --tail=50

# 2. Verify backend service and endpoints are healthy
kubectl get svc,endpoints product-catalog -n astronomy-shop

# 3. Inspect frontend deployment spec for pod-local overrides (Smoking Gun)
kubectl get deployment frontend -n astronomy-shop -o yaml | grep -A 5 hostAliases
```

---

## 4. Expected Diagnosis
```text
Root cause: The frontend Deployment in the astronomy-shop namespace carries a stale hostAliases entry mapping product-catalog to 127.0.0.1. The kubelet populates /etc/hosts with this entry, which libc resolves before CoreDNS. Consequently, all frontend outbound calls to product-catalog fail locally with ECONNREFUSED, even though CoreDNS, the backend Service, and its Endpoints are healthy.
```

---

## 5. Expected Mitigation
The operator or agent must remove the `hostAliases` override from the `frontend` Deployment spec:

```bash
kubectl patch deployment frontend -n astronomy-shop --type=json -p='[{"op": "remove", "path": "/spec/template/spec/hostAliases"}]'
```

And wait for the deployment rollout to complete:
```bash
kubectl rollout status deployment frontend -n astronomy-shop --timeout=180s
```

### Oracles & Verification
The problem implements a compounded mitigation evaluation (`CompoundedOracle`):
1. **`0-MitigationOracle` (Base Oracle)**: Verifies generic namespace health (no deleted deployments or pods scaled to zero).
2. **`1-StaleHostAliasesMitigationOracle` (Problem-Specific Oracle)**:
   - Asserts that `.spec.template.spec.hostAliases` is completely removed from the `frontend` Deployment.
   - Verifies that no active `frontend` pod carries the `product-catalog` override in its spec.
   - Confirms that the deployment rollout is complete and endpoints are ready.
   - Executes a synthetic HTTP catalog probe (`/api/products`) through the edge proxy to verify live traffic end-to-end.

---

## 6. Local Verification & Test Output
The problem and mitigation oracle have been verified locally using both unit tests (`pytest`) and SREGym CLI evaluation:

### Unit Tests:
```bash
pytest tests/oracles/test_stale_hostaliases_mitigation.py tests/problems/test_stale_hostaliases_dns_poisoning_astronomy_shop.py
```
*Result: 17 passed*

### SREGym CLI Execution Output:
```json
{
  "Mitigation": {
    "success": true,
    "accuracy": 100.0,
    "oracles": [
      {
        "name": "0-MitigationOracle",
        "success": true
      },
      {
        "name": "1-StaleHostAliasesMitigationOracle",
        "success": true,
        "deployment_exists": true,
        "rollout_complete": true,
        "template_override_removed": true,
        "no_pod_carries_override": true,
        "ready_endpoint": true,
        "product_probe_succeeded": true,
        "reason": "the hosts override is gone and catalog requests succeed"
      }
    ]
  },
  "TTM": 396.85
}
```

---

## Checklist
- [x] Problem implementation adheres to SREGym conductor & problem registry standards.
- [x] Specific mitigation oracle added with spec checks and end-to-end synthetic traffic validation.
- [x] Unit tests added in `tests/oracles/test_stale_hostaliases_mitigation.py` and `tests/problems/test_stale_hostaliases_dns_poisoning_astronomy_shop.py`.
- [x] Tested locally via `cli.py` on Astronomy Shop workload.
- [x] Code follows repository standards and formatting guidelines.


Talhaasif7:
@tianyin @Saadmrp1038 
Any review or feedback would be much appreciated.

munimthahmid:
@Talhaasif7 Can you add the real-world failure scenario link for this failure scenario?

Talhaasif7:
@munimthahmid 

Here are the real-world incident/issue references along with the official Kubernetes architectural documentation:

### Real-World Incident & Work Item References
* **GitLab Runner Issue #2229:** [gitlab-runner#2229 - Kubernetes executor DNS resolution and hostAliases override issues](https://gitlab.com/gitlab-org/gitlab-runner/-/work_items/2229)
* **GitLab Runner MR #1680:** [gitlab-runner!1680 - Add support for hostAliases in Kubernetes executor](https://gitlab.com/gitlab-org/gitlab-runner/-/merge_requests/1680)
  * *Context:* Real-world cases where static `/etc/hosts` and `hostAliases` entries in Kubernetes executors caused service discovery drift and shadowed dynamic DNS lookups for connected services.

### Kubernetes Documentation
* **Kubernetes Docs:** [Adding entries to Pod /etc/hosts with HostAliases](https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/)
  * *Mechanism:* Explains how `hostAliases` directly populates `/etc/hosts`, taking precedence over cluster-wide CoreDNS resolution due to `nsswitch.conf` (`files dns`), leading to localized connection failures when overrides become stale.

munimthahmid:
@Talhaasif7  Thanks for the references. I think the failure implemented here is quite different from the GitLab issue you shared.

That issue was fixed by adding a loopback `hostAlias` for service containers in the same Pod, whereas this problem is caused by a stale loopback `hostAlias` shadowing a healthy Service in another Pod.

Could you provide an incident or postmortem matching the scenario implemented here?

Talhaasif7:
@munimthahmid 
Thanks for pointing this out. Just to clarify the context: this scenario was originally proposed in Issue #943, which was reviewed and validated before implementation. After that review, I was asked to implement the scenario on the Astronomy Shop application specifically, which is why the current PR uses frontend -> product-catalog with the stale hostAliases mapping to 127.0.0.1.

So the Astronomy Shop version is the concrete SREGym implementation of the failure scenario discussed in #943, rather than a new failure mode introduced in this PR.

munimthahmid:
@Talhaasif7 Ok I see.

@Saadmrp1038 is a bit busy right now. He will review the PR once he has time. Thanks for your patience.

Talhaasif7:
@Saadmrp1038 Hey Saad! Any feedback from your side would be highly appreciated

munimthahmid (changes requested):
@Talhaasif7 I tested this end to end twice with `gpt-5.6-luna` at xhigh reasoning. Both runs passed diagnosis and mitigation with 100% accuracy and finished in under five minutes. The root cause is fairly easy to spot from the `ECONNREFUSED 127.0.0.1:8080` error and the `hostAliases` entry in the frontend Deployment.

The diagnosis and fix are also very similar to the existing `env_variable_shadowing_astronomy_shop` problem. I would like you to explore this a bit more and come up with a substantial redesign that makes the problem more useful and meaningfully harder for the agent. If possible, please also run a few agent attempts on the redesigned version and include the results.

Talhaasif7:
@munimthahmid 
The original scenario was based on the Astronomy Shop issue that was verified earlier, which is why I used it as the basis for this failure. However, I agree that the current implementation is too easy for the agent because the ECONNREFUSED 127.0.0.1:8080 symptom points quite directly to the hostAliases entry in the same Deployment.

I’ll redesign the scenario while keeping the same underlying failure mode (a stale loopback hostAlias shadowing a healthy Kubernetes Service), but make the failure propagate through an intermediate service/proxy. This should require the agent to trace the service path and distinguish between healthy backend components and the actual source of the failure, rather than immediately matching the error to the hostAliases configuration.

I’ll also run several agent attempts on the redesigned version and include the diagnosis/mitigation results in the PR.
Lemme know if it sounds reasonable.

tianyin:
Dear @Talhaasif7 

Thanks for the continuous effort! I appreciate it. 

I like the plan in terms of the thinking process. I think the biggest challenge for SREGym is to make problems challenging for frontier AI and that's the focus of the team. At this point, it's quite experimental -- you will have to try a few times and understand what's the strength and weakness of frontier AI agents and design problems that can challenge them. A too easy problem creates little value for a benchmark.

> I’ll also run several agent attempts on the redesigned version and include the diagnosis/mitigation results in the PR.
Lemme know if it sounds reasonable.

Yes, please!

Talhaasif7:
@tianyin 
Thanks Professor for kind words.
I will keep your guidance in mind while thinking about the architecture of the problem.

Talhaasif7:
@munimthahmid 
I have completely redesigned the benchmark problem by shifting the fault injection target from `frontend` to `frontend-proxy`. In the previous design, poisoning `frontend` directly produced an immediate `ECONNREFUSED 127.0.0.1:8080` in `frontend` logs, allowing frontier agents to trivially solve the issue in under 5 minutes without deep troubleshooting. In this updated implementation, `frontend-proxy` (Envoy) is configured with a stale `hostAliases` entry mapping `frontend` to `127.0.0.1`. Consequently, all control plane signals, CoreDNS, and backend microservices (`frontend`, `product-catalog`) remain 100% healthy and green with zero errors in `frontend` logs, while failures manifest strictly at the edge proxy on upstream requests (e.g., `/api/products` 503 errors). This requires the agent to perform genuine multi-hop reasoning by correlating edge proxy failures, verifying that backend services and in-cluster DNS are healthy, and deducing that a pod-local `/etc/hosts` override is shadowing CoreDNS.

Additionally, we updated the compounded mitigation oracle (`StaleHostAliasesMitigationOracle`) to deterministically enforce five strict validation steps: verifying removal of `.spec.template.spec.hostAliases` from `frontend-proxy`, ensuring no active pods carry the override in their spec, confirming rollout completion, verifying endpoint readiness, and executing a synthetic HTTP catalog probe (`/api/products`) with retry resilience to verify live traffic end-to-end. We verified this implementation end-to-end on a live Kind cluster as well, where the unmitigated state failed deterministically and the mitigated state passed with 100% accuracy (`PRODUCTS_OK`). All 23 tests (22 unit tests and 1 live cluster integration test) are now passing cleanly.

HacksonClark:
/evaluate-problem stale_hostaliases_dns_poisoning_astronomy_shop

Talhaasif7:
@HacksonClark 
Any feedback from your side would be much appreciated.

HacksonClark:
> @HacksonClark Any feedback from your side would be much appreciated.

Doing some last checks, fault seems to be working great! The results seem to indicate it's difficult enough to include in the benchmark. Just doing some last checks to make sure those failures were due to difficulty and not infra issues with the runner (I'm still actively ironing out issues there).

sregym/conductor/oracles/stale_hostaliases_mitigation.py
```diff
@@ -0,0 +1,289 @@
...
+
+        settled = self._wait_for_current_rollout(deployment)
+        if settled is None:
+            results["reason"] = f"deployment '{self.deployment_name}' did not complete its current rollout"
+            print(f"[FAIL] {results['reason']}")
+            return results
+        deployment = settled
+        results["rollout_complete"] = True
+
+        if self._template_override(deployment):
```
munimthahmid: I think the oracle is too strict about requiring the alias to be removed. If an agent changes `FRONTEND_HOST` to `frontend.astronomy-shop.svc.cluster.local` and restores traffic, this check still rejects the unused short-name alias before running the HTTP probe. I think a working, persistent fix like this should be accepted.

tests/problems/test_stale_hostaliases_dns_poisoning_astronomy_shop.py
```diff
@@ -0,0 +1,150 @@
...
+    try:
+        assert problem.mitigation_oracle.evaluate()["success"] is True, "healthy cluster should pass"
+
+        problem.inject_fault()
+        deployment = problem.kubectl.get_deployment(problem.faulty_service, problem.namespace)
+        aliases = deployment.spec.template.spec.host_aliases
+        assert aliases and aliases[0].ip == problem.blackhole_ip
+        assert problem.target_backend in aliases[0].hostnames
+
+        assert problem.mitigation_oracle.evaluate()["success"] is False, "the oracle must fail while the fault is live"
```
munimthahmid, line 186: The integration test doesn’t currently prove that injection breaks traffic. The oracle fails as soon as it sees the alias, so the test could pass even if the alias has no effect on requests. Please check `/api/products` directly and verify that it works before injection, fails afterward, and works again after recovery.

sregym/generators/fault/inject_virtual.py
```diff
@@ -3166,7 +3166,61 @@ def recover_fd_exhaustion(self, microservices: list[str], entrypoint_cmd: str):
...
+                namespace=self.namespace,
+                body=patch,
+            )
+            print(f"Removed the hosts override from {service}")
+            self._wait_for_rollout(service, rollout_timeout)
+
     ############# HELPER FUNCTIONS ################
+    def _wait_for_rollout(self, service: str, timeout: int = 180):
+        """Block until the Deployment's new ReplicaSet is fully rolled out."""
+        result = self.kubectl.exec_command(
```
munimthahmid: A rollout timeout currently gets logged but doesn’t stop fault injection. This can let the agent start while the old, unaffected pods are still serving traffic. Please use `exec_command_checked()` here and add a test confirming that a failed rollout stops setup.

HacksonClark:
@Talhaasif7 Please address @munimthahmid's comments. Great work on this!

Talhaasif7:
@munimthahmid @HacksonClark 
Thanks for the valuable review and suggestions! I have updated the implementation to address all three points:

1. **Oracle Flexibility:** In StaleHostAliasesMitigationOracle.evaluate(), the functional HTTP product probe (/api/products) is now the authoritative evaluation gate. If an agent applies an alternative valid fix (such as configuring FRONTEND_HOST with the FQDN to bypass the short-name alias override) and catalog traffic is successfully restored through the edge proxy, the oracle accepts it. The hostAliases template and pod checks are now captured as informational fields rather than blocking gates, and a unit test (test_passes_when_alias_present_but_traffic_works) has been added.

2. **Direct Traffic Verification in Integration Test:** Updated test_lifecycle_against_a_live_cluster to explicitly execute _run_product_probe() across all three lifecycle phases: verifying that the probe succeeds before fault injection, confirms traffic failure while the fault is active, and succeeds again after recovery.

3. **Checked Rollout Execution:** Replaced exec_command with exec_command_checked in VirtualizationFaultInjector._wait_for_rollout so that rollout timeouts or failures immediately raise a RuntimeError rather than silently proceeding, backed by the new unit test test_injection_raises_on_rollout_timeout.

All 24 unit tests in both test suites are passing cleanly.

munimthahmid:
@Talhaasif7  The updated integration test crashes before fault injection. `problem.mitigation_oracle` is a `CompoundedOracle`, which doesn’t have `_run_product_probe()`, so the new calls raise an `AttributeError`. Can you fix this?

Talhaasif7:
@munimthahmid 
Since problem.mitigation_oracle is a CompoundedOracle, I've updated StaleHostAliasesDNSPoisoningAstronomyShop to:
Store the StaleHostAliasesMitigationOracle on self.hosts_mitigation_oracle.
Forward self.mitigation_oracle._run_product_probe = self.hosts_mitigation_oracle._run_product_probe (and provide a _run_product_probe() helper on the problem instance) so direct calls on the compound oracle or problem don't raise an AttributeError.
Update test_lifecycle_against_a_live_cluster to call hosts_oracle._run_product_probe() via problem.hosts_mitigation_oracle.
Added test_problem_wires_hosts_mitigation_oracle_and_probe_method to the unit test suite to continuously verify this wiring.
All 25 unit tests are passing cleanly now.

munimthahmid:
/validate-problem stale_hostaliases_dns_poisoning_astronomy_shop

HacksonClark:
/evaluate-problem stale_hostaliases_dns_poisoning_astronomy_shop

Talhaasif7:
@munimthahmid @HacksonClark 
Any feedback would be appreciated

munimthahmid:
/evaluate-problem stale_hostaliases_dns_poisoning_astronomy_shop

munimthahmid (approved):
@Talhaasif7  Thanks a lot for your continued hard work. Your redesign was good, but the task was still quite straightforward, with GLM passing mitigation in every completed run.

So I changed it so that some pods still use the old routing while others use the corrected routing. Both paths look healthy, but they save cart updates to different databases. The agent now has to explain why customers see different carts despite everything looking healthy, then bring the updates back together without losing items or restoring items customers already cleared. Simply fixing the routing is no longer enough.

A frontier model like Astra can still solve it pretty fast. I ran Astra with high reasoning and fast mode, and it correctly diagnosed and fixed the problem in about 4.5 minutes. Still, I think the updated fault provides a useful challenge, even if it doesn’t slow down every model.

I think it’s good to merge. Thanks again!

Talhaasif7:
@munimthahmid
Finally, after 2.5 months, I see the word merge. 😂
Jokes apart, I learned a ton throughout this process. The rigorous evaluation process challenged me in the best way and pushed me to dig deep, but having such supportive teammates made all the difference. Thank you for the continuous guidance and constructive feedback along the way!

munimthahmid:
You did very well, @Talhaasif7. Feel free to bring more ambitious problems that can challenge the frontiers!

PR #961, opened 2026-08-22
Restore the pre-injection kube-proxy image

### LINKED ISSUE #808: fix_kubernetes() silently downgrades kube-proxy to v1.31.13 on every run

Saadmrp1038:
Fixes #808

Save the current kube-proxy image before injecting the fault, then restore the same image during recovery.

Previously, recovery always changed kube-proxy to v1.31.13. This could downgrade a cluster using a newer Kubernetes patch version, even when no fault had been injected.

Recovery now uses the image saved during injection. The saved image is kept if recovery is interrupted or fails, allowing cleanup to retry later. If no image was saved, cleanup does not change kube-proxy.

/validate-problem workload_imbalance

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -22,9 +22,9 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
 
-    - name: Download CodeQL CLI v2.23.5
+    - name: Download CodeQL CLI v2.26.3
       run: |
-        curl -L https://github.com/github/codeql-cli-binaries/releases/download/v2.23.5/codeql-linux64.zip -o codeql.zip
+        curl -L https://github.com/github/codeql-cli-binaries/releases/download/v2.26.3/codeql-linux64.zip -o codeql.zip
         unzip codeql.zip
         mv codeql /opt/codeql
         echo "/opt/codeql" >> $GITHUB_PATH
--- a/sregym/conductor/conductor.py
+++ b/sregym/conductor/conductor.py
@@ -704,9 +704,7 @@ class Conductor:
         self.logger.info("[FIX] Imbalance leftover if any")
 
         injector = VirtualizationFaultInjector(namespace="kube-system")
-        injector.recover_daemon_set_image_replacement(
-            daemon_set_name="kube-proxy", original_image="registry.k8s.io/kube-proxy:v1.31.13"
-        )
+        injector.recover_daemon_set_image_replacement(daemon_set_name="kube-proxy")
 
         self.logger.info("[FIX] KubeletCrash leftover if any")
         injector = RemoteOSFaultInjector()
--- a/sregym/conductor/problems/workload_imbalance.py
+++ b/sregym/conductor/problems/workload_imbalance.py
@@ -51,9 +51,8 @@ class WorkloadImbalance(Problem):
     @mark_fault_injected
     def recover_fault(self):
         print("== Fault Recovery ==")
-        self.injector.inject_daemon_set_image_replacement(
-            daemon_set_name="kube-proxy", new_image="registry.k8s.io/kube-proxy:v1.31.13"
-        )
+        if not self.injector.recover_daemon_set_image_replacement(daemon_set_name="kube-proxy"):
+            raise RuntimeError("The saved kube-proxy image is missing")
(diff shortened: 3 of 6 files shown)


tianyin (approved):
LGTM

PR #962, opened 2026-08-23
Persist stratus token usage before completion

### LINKED ISSUE #934: Stratus underreports token usage by excluding summary calls and incomplete runs
## Description
Stratus does not account for every successful LLM request when reporting token usage. Summary calls are excluded, and usage collected during an attempt is not persisted if the run does not complete cleanly.

### Usage is not persisted incrementally
For regular agent calls, the driver reads the callback's accumulated usage only after the entire attempt returns:

```python
agent, last_state, graph_events = await mitigation_agent_single_run(...)

usage_metadata = next(iter(agent.callback.usage_metadata.items()))[1]
```

The final token-usage CSV is written near the end of `main()`, after diagnosis and mitigation have finished. If an attempt times out, is interrupted, or raises an exception after making successful LLM requests, its callback usage is not extracted or persisted. Depending on where execution stops, the run may not produce a final token-usage CSV at all.

### Summary calls are excluded
Stratus generates a diagnosis summary and, when using `validate` retry mode, a summary before each mitigation retry:

```python
llm = get_llm_backend_for_agent()
res = llm.inference(summary_input_messages)
```

`generate_run_summary()` creates a separate LLM backend outside the agent graph. The agent's `UsageMetadataCallbackHandler` is not attached to this backend, so these requests are not included in the reported `input_tokens`, `output_tokens`, or `total_tokens`. As a result, even a run that completes normally can underreport token usage. With `validate` retry mode, every mitigation retry adds another uncounted summary request.

## Steps to reproduce

1. Run Stratus through the diagnosis stage and inspect the provider's request records.
2. Compare those records with Stratus's token-usage output after the diagnosis summary is generated.
3. Run mitigation with `retry_mode: validate` and trigger at least one retry.
4. Observe that the diagnosis-summary and retry-summary requests appear in the provider records but are absent from Stratus's totals.
5. Alternatively, interrupt an attempt after a successful LLM call and observe that the accumulated usage is not persisted to the final CSV.

## Expected behavior

Stratus should record the usage of every successful LLM request when its response is received. This should include diagnosis summaries and mitigation retry summaries. Usage already collected should remain available if a later part of the run times out, is interrupted, or raises an exception.

Saadmrp1038:
Fixes #934.

`stratus` currently reports token usage only after an agent attempt finishes. This loses usage from completed LLM requests when the process stops early. Summary requests are also missing because they run outside the agent callback.

Persist each successful `stratus` LLM response immediately in the run artifact directory. The records distinguish trace, tool, and summary usage. Completed runs add summary usage to the existing CSV without counting other requests twice. Interrupted runs keep all records written before the interruption.

Only the `stratus` driver enables this behavior. The judge, preflight process, and other agents remain unchanged. This is a targeted fix for `stratus` only. The records contain usage metadata.

### DIFF
--- a/clients/stratus/stratus_agent/driver/driver.py
+++ b/clients/stratus/stratus_agent/driver/driver.py
@@ -22,6 +22,7 @@ import requests  # noqa: E402
 import yaml  # noqa: E402
 from langchain_core.messages import HumanMessage, SystemMessage  # noqa: E402
 
+from llm_backend.usage_log import USAGE_LOG_PATH_ENV, summarize_usage  # noqa: E402
 from logger import init_logger  # noqa: E402
 
 init_logger()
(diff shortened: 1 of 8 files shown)


tianyin:
I now understand this issue. Good catch!

tianyin (approved):
Great work!

PR #963, opened 2026-08-23
Apply LLM judge settings to requests

### LINKED ISSUE #947: Judge oracle silently ignores temperature, max_tokens, and model constructor args
## Summary

The LLM-as-a-judge classes (`LLMAsAJudgeOracle`, `DiagnosisJudge`, `LLMJudge`) accept `temperature`, `max_tokens`, `model_name`, `url`, `api_key`, and `provider` constructor arguments, store them, but never forward them when initializing the backend. `get_llm_backend_for_judge()` is called with no arguments and is driven entirely by `JUDGE_MODEL_ID` / `JUDGE_API_BASE` / `JUDGE_API_KEY` environment variables plus the hardcoded defaults in `LiteLLMBackend`. As a result, most of these constructor parameters are dead (write-only) fields.

## Trace

1. `LLMAsAJudgeOracle.__init__` (`sregym/conductor/oracles/llm_as_a_judge/llm_as_a_judge_oracle.py:18`) accepts `temperature=0.0`, `max_tokens=4096` and forwards them to `DiagnosisJudge(...)` (line 30).
2. `DiagnosisJudge.__init__` (`sregym/conductor/oracles/llm_as_a_judge/judge.py:227-228`) stores `self.temperature` / `self.max_tokens` (same pattern in `LLMJudge`, `judge.py:51-52`).
3. The `backend` property (`judge.py:261-271`) calls `get_llm_backend_for_judge()` with **no arguments** - none of the stored fields are read.
4. `get_llm_backend_for_judge()` (`llm_backend/init_backend.py:29-38`) -> `get_llm_backend()` (line 14) constructs `LiteLLMBackend(model_name=..., api_base=..., api_key=...)` - without `temperature` or `max_tokens`.
5. The effective values are therefore the `LiteLLMBackend` defaults: `temperature=0.0`, `top_p=0.95`, `max_tokens=None` (`llm_backend/get_llm_backend.py:29-31`).

## Impact

- **`temperature`** - latent/inert bug: the intended value (`0.0`) matches the fallback default (`0.0`), so there is no behavioral difference today. It only bites if the oracle is ever constructed with a non-zero temperature expecting it to take effect.
- **`max_tokens`** - active discrepancy: the intended `4096` is never applied; the judge actually runs with `max_tokens=None` (unbounded / provider default).
- **`model_name` / `url` / `api_key` / `provider`** - ignored by the backend, which reads `JUDGE_MODEL_ID` etc. from the environment instead. Note `self.model_name` still populates the report's `evaluator_model` field (`judge.py:317, 383`), so if a caller passes a `model_name` that differs from `JUDGE_MODEL_ID`, the emitted report **mislabels** which model actually performed the judgment.

## Suggested fix

Either (a) forward the stored parameters through `get_llm_backend_for_judge()` into `LiteLLMBackend`, or (b) remove the unused constructor parameters and rely solely on the env-driven configuration, and derive `evaluator_model` from the actual backend model (`JUDGE_MODEL_ID`) rather than the constructor arg to avoid mislabeling.

Saadmrp1038:
Fixes #947

The LLM judge accepted a model, provider, endpoint, API key, temperature, and output-token limit. The backend ignored these values. 

This change passes these values to the backend. It also records the model that the backend uses. Agent LLM requests do not change.

### DIFF
--- a/llm_backend/get_llm_backend.py
+++ b/llm_backend/get_llm_backend.py
@@ -30,10 +30,12 @@ class LiteLLMBackend:
         top_p: float = 0.95,
         temperature: float = 0.0,
         max_tokens: int | None = None,
+        provider: str | None = None,
     ):
         self.model_name = model_name
         self.api_key = api_key
         self.api_base = api_base
+        self.provider = provider
         self.temperature = temperature
         self.top_p = top_p
         self.max_tokens = max_tokens
@@ -93,6 +95,8 @@ class LiteLLMBackend:
             model_config["api_key"] = self.api_key
         if self.api_base is not None:
             model_config["api_base"] = self.api_base
+        if self.provider is not None:
+            model_config["custom_llm_provider"] = self.provider
         if self.max_tokens is not None:
             model_config["max_tokens"] = self.max_tokens
 
--- a/llm_backend/init_backend.py
+++ b/llm_backend/init_backend.py
@@ -7,11 +7,21 @@ def get_llm_backend(
     model_name: str,
     api_base: str | None = None,
     api_key: str | None = None,
+    provider: str | None = None,
+    temperature: float = 0.0,
+    max_tokens: int | None = None,
 ) -> LiteLLMBackend:
     """Initialize an LLM backend for the given litellm model string."""
     endpoint_status = "set" if api_base else "unset"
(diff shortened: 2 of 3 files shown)


tianyin (approved):
Thanks for the work. LGTM

I'm a bit surprised to see the feature being requested, but it's perhaps always good to provide configurability.

PR #964, opened 2026-08-23
Hide Helm release secrets from the agent-facing API proxy

### LINKED ISSUE #955: Agents can recover the pre-fault manifest from the Helm release secret
Agents can read the Helm release secret in the target namespace and recover the pre-fault manifest of every object in the application. For faults that change an existing field, this turns diagnosis into a diff: compare the live object against Helm's stored copy, and the injected change is the only difference.

fix 
In sregym/service/k8s_proxy.py, add one entry to HIDDEN_LABELS:

"owner": {"helm"}

sidbhanushali:
Closes #955

## Problem

Agents can read the `sh.helm.release.v1.*` Secrets in the target namespace and recover the pre-fault manifest of every object in the application. For faults that change an existing field, diagnosis degenerates into a diff: compare the live object against Helm's stored copy and the injected change is the only difference.

## Fix

1. `owner=helm` added to `HIDDEN_LABELS` in `sregym/service/k8s_proxy.py` (the fix sketched in #955): Helm release Secrets are now filtered from list responses and direct GETs return 403 via the existing hidden-resource check.
2. While adding the test for it, I found the label filter only applied to **cluster-wide** list paths. Namespaced lists (`/api/v1/namespaces/{ns}/{resource}`, i.e. plain `kubectl get secrets -n astronomy-shop`) returned hidden-label resources unfiltered, which would have made the #955 fix ineffective for the most natural agent command, and also means load-generator pods were visible via namespaced `kubectl get pods`. `should_filter_response` now treats namespaced resource lists as filterable; a trailing name segment (single resource / subresource) is still handled by the direct-access 403 check.
3. The filtering helpers moved from the request-handler closure to module level so they can be unit-tested without a cluster: `tests/service/test_k8s_proxy_filtering.py` covers Helm secrets in namespaced/cluster-wide/table-format lists, direct-GET handling, load-generator filtering in namespaced pod lists, hidden-namespace items, and subresource/query-string paths. `uv run pytest tests/service/test_k8s_proxy_filtering.py` passes (8 tests); behavior is otherwise unchanged.

## Behavior change note

Agents (and the MCP kubectl tool, which routes through the same proxy) lose visibility of Helm release secrets and, newly, of hidden-label resources in namespaced lists. Both are the intended isolation semantics; conductor-side code uses the unproxied kubeconfig and is unaffected.

## Code changes

| File | Change |
| --- | --- |
| `sregym/service/k8s_proxy.py` | `owner=helm` hidden label; namespaced-list filtering; filtering helpers extracted to module level |
| `tests/service/test_k8s_proxy_filtering.py` | New unit tests for the filtering logic |

/validate-problem skip

### DIFF
--- a/sregym/service/k8s_proxy.py
+++ b/sregym/service/k8s_proxy.py
@@ -41,7 +41,100 @@ HIDDEN_LABELS: dict[str, set[str]] = {
     "app": {"load-generator", "locust-fetcher"},
     "job": {"workload"},
     "opentelemetry.io/name": {"load-generator"},
+    # Helm stores the full pre-fault manifest of every release object in its
+    # release Secrets (owner=helm); hiding them stops agents from diffing the
+    # live cluster against Helm's stored copy to recover injected faults (#955).
+    "owner": {"helm"},
 }
+
+
+def has_hidden_label(metadata: dict, hidden_labels: dict[str, set[str]]) -> bool:
+    """Return whether a resource's metadata carries any hidden label."""
+    labels = metadata.get("labels") or {}
+    return any(labels.get(key) in values for key, values in hidden_labels.items())
+
+
+def filter_namespace_list(data: dict, hidden_namespaces: set[str]) -> dict:
+    """Filter hidden namespaces from a namespace list response."""
+    if "items" in data:
+        data["items"] = [
+            item for item in data["items"] if item.get("metadata", {}).get("name") not in hidden_namespaces
+        ]
+    # Handle Table format (kubectl's default)
+    if "rows" in data:
+        data["rows"] = [
+            row
+            for row in data["rows"]
+            if row.get("object", {}).get("metadata", {}).get("name") not in hidden_namespaces
+        ]
+    return data
+
+
+def filter_resource_list(data: dict, hidden_namespaces: set[str], hidden_labels: dict[str, set[str]]) -> dict:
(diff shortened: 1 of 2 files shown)


Saadmrp1038:
Thanks for the PR. I tested the implementation with several access methods.

The filter identifies Helm release Secrets only through the `owner=helm` label. This label can change. An agent can remove it because Helm Secret names are predictable:

`kubectl label secret <helm-secret-name> owner-`

The agent can then read the Secret and recover the original application manifest. Encoded paths, protobuf responses, Secret watches, and Pod mounts also bypass the filter.

---

I am closing this PR.

The issue already had an assignee when this PR was opened. Before starting work, please check the assignees and coordinate with them. If an issue is unassigned, please assign it to yourself first. This prevents duplicate work.

Please feel free to work on another issue. Thanks again!

sidbhanushali:
You're right on the bypass, an agent with mutating access can strip the label and read the secret, so the filter only holds if nothing can relabel, and the protobuf, watch, and mount paths mean this isn't solvable at the list-filter layer anyway. Fair close.

My apologies, I didn't look at the assignees before starting. Going forward I'll check assignees first, coordinate on the issue before writing code if it's taken, and assign myself if it's open. Thanks for the review!

Saadmrp1038:
@sidbhanushali No worries

PR #968, opened 2026-08-25
Restrict agent access to the SREGym repo

### LINKED ISSUE #911: Restrict internet access during agent evaluation

Saadmrp1038:
Containerized agents currently have unrestricted internet access during evaluations. This allows an agent to read the SREGym repository and learn fault implementation details instead of diagnosing the deployed system.

This change makes filtered internet access the default. Agents can still contact model providers and use the public internet, but requests to the SREGym repository are blocked through a per-run network proxy. Provider-side web and GitHub tools that could bypass this proxy are also disabled. The previous unrestricted behavior remains available through `--internet-access open`.

Kubernetes access remains available through an authenticated TLS proxy. The unfiltered kubeconfig is no longer mounted inside agent containers, and filtered agents cannot create temporary workloads to access the repository through the cluster network. They can still inspect and modify existing workloads during diagnosis and mitigation.

The proxy, temporary credentials, Docker network, and audit files are cleaned after each run. Run results now record the selected internet policy and the number of blocked requests. The README also explains how local model servers must be exposed to filtered agent containers.

This prevents direct access to the benchmark repository. It is not intended to block every possible public copy of the source. We can later expand this to block other sites if needed.

The implementation was tested with Claude Code and Codex on Kind and a live Kubernetes cluster.

Closes #911

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,9 @@ SREGym supports local models through Ollama and OpenAI-compatible servers such a
 
 Set `AGENT_API_KEY` as well if the endpoint requires authentication.
 
+> [!CAUTION]
+> When you use `--internet-access filtered`, the agent runs on an isolated Docker network. It cannot reach a local model server that listens only on `127.0.0.1` or `localhost`. Configure the server to listen on a host-reachable interface, such as `0.0.0.0`, and use `http://host.docker.internal:<port>` as the API base. Protect the exposed port with authentication or a firewall. This requirement is the same for Kind and external Kubernetes clusters because the connection is between the agent container and the machine running SREGym.
+
 **Stratus with Ollama:**
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -128,6 +128,9 @@ Use `--force-build` to rebuild the container image after updating dependencies o
 uv run main.py --agent codex --model gpt-5 --force-build
 ```
 
+Containerized agents can use the public internet by default, but direct access to the benchmark's GitHub source is
+blocked. Use `--internet-access open` only when you intentionally need the previous unrestricted network behavior.
+
 ### Model Selection
 
 SREGym uses [LiteLLM](https://docs.litellm.ai/docs/providers) model strings directly (no config file needed). Just pass any supported model string via `--model`:
--- a/clients/claudecode/claudecode_agent.py
+++ b/clients/claudecode/claudecode_agent.py
@@ -103,6 +103,13 @@ class ClaudeCodeAgent:
         "WebSearch",
     ]
 
+    @classmethod
+    def allowed_tools(cls) -> list[str]:
+        tools = list(cls.ALLOWED_TOOLS)
+        if os.environ.get("AGENT_INTERNET_ACCESS") == "filtered":
+            tools.remove("WebSearch")
+        return tools
+
     def __init__(
         self,
         logs_dir: Path,
@@ -354,7 +361,7 @@ class ClaudeCodeAgent:
         ]
         if reasoning_effort:
             command.extend(["--effort", reasoning_effort])
-        command.extend(["--allowedTools", *self.ALLOWED_TOOLS])
+        command.extend(["--allowedTools", *self.allowed_tools()])
 
         logger.info(f"Executing command: {' '.join(command)}")
 
--- a/clients/copilot/copilot_agent.py
+++ b/clients/copilot/copilot_agent.py
@@ -180,6 +180,10 @@ class CopilotCliAgent:
         reasoning_effort = os.environ.get("AGENT_REASONING_EFFORT")
         if reasoning_effort:
             command.extend(["--reasoning-effort", reasoning_effort])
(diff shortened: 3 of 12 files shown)


HacksonClark (approved):
This is very similar to the harbor framework implementation which works quiet well. Great work!

PR #969, opened 2026-08-25
Protect Helm release secrets from agents

### LINKED ISSUE #955: Agents can recover the pre-fault manifest from the Helm release secret

Saadmrp1038:
Fixes #955.

Helm stores each release revision in a Kubernetes Secret. These Secrets contain the complete rendered application manifests. Agents could read them and recover the pre-fault configuration without diagnosing the problem.

This change filters Helm release Secrets through the Kubernetes API proxy. Agents cannot list, read, modify, delete, or watch these records. They also cannot expose them through workload mounts or updates.

The proxy requests JSON when it must inspect a response. It rejects the response if it cannot filter the data safely. Normal application Secrets and other Kubernetes operations remain available.

Live testing covered authentication, normal Secret access, Helm Secret filtering, token rotation, and workload bypass attempts. Also did a codex run on the secret-based `secret_rotation_stale_env_credentials_astronomy_shop` fault. The run completed the normal deployment, proxy, evaluation, and cleanup lifecycle while ordinary application Secrets remained available.

### DIFF
--- a/sregym/service/k8s_proxy.py
+++ b/sregym/service/k8s_proxy.py
@@ -23,9 +23,10 @@ import ssl
 import tempfile
 import threading
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
-from urllib.parse import urlparse
+from urllib.parse import parse_qs, unquote, urlparse, urlsplit
 
 import urllib3
+import yaml
 from kubernetes import config
 
 logger = logging.getLogger("all.infra.k8s_proxy")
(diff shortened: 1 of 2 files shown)


HacksonClark (approved):
This is great! This should be covered by proxies.

HacksonClark:
Will merge if smoke test passes.

PR #975, opened 2026-08-26
Stream MCP SSE in filtered mode

Saadmrp1038:
Filtered agent containers send HTTP traffic through mitmproxy. The proxy buffered long-lived MCP SSE responses, causing Stratus to time out when connecting to tools such as Jaeger.

This change streams only `text/event-stream` responses. Normal HTTP behavior, GitHub filtering, and open-internet runs remain unchanged.


tianyin (approved):
LGTM

PR #976, opened 2026-08-26
Fix DeepSeek reasoning blocks

### LINKED ISSUE #787: DeepSeek-reasoner error `thinking` blocks rejected in input message
DeepSeek-reasoner returns `AIMessage.content` as a list containing `{'type': 'thinking', ...}` blocks alongside `{'type': 'text', ...}` blocks. LangChain stores these in the conversation history as-is. On subsequent turns, the full message history (including thinking blocks) is sent back to the API. DeepSeek's API rejects thinking as an input content type.

```
litellm.exceptions.BadRequestError: litellm.BadRequestError: DeepseekException - {"error":{"message":"Failed to deserialize the JSON body into the target type: messages[17]: unknown variant `thinking`, expected `text` at line 1 column 33062","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
```

---

We can strip the thinking blocks, but need to be careful about whether it affects other models. None of the other models had any issues (kimi, glm, openai, gemini, claude) of this kind.

Saadmrp1038:
Closes #787

DeepSeek Reasoner returns its reasoning as `thinking` blocks. These blocks are valid in the model response.

The error occurred during the next Stratus step. The old adapter stored the response and sent its `thinking` blocks back as input. DeepSeek does not accept `thinking` as an input content type, so it rejected the request.

[langchain-litellm 0.6.6](https://github.com/langchain-ai/langchain-litellm/releases#release-langchain-litellm==0.6.6) includes this upstream fix:

> strip thinking/redacted_thinking blocks from messages sent to non-Anthropic providers

This PR pins `langchain-litellm` version to `0.6.6`. So, the adapter will now remove these blocks before it sends the conversation back to DeepSeek.


tianyin (approved):
glad it's solved by litellm already

PR #980, opened 2026-08-27
Remove personal GeniLib tooling

Saadmrp1038:
This PR removes the `genictl` tool. I am moving this to a separate repo. I was the only one using this tool here so there should be no issues.

/validate-problem skip


tianyin (approved):
LGTM

PR #987, opened 2026-08-30
Fix mitigation results being lost

Saadmrp1038:
During Opus 5 and Sonnet 5 runs on SREGym-Lite, some attempts had no mitigation result. This affected 6 of 63 Sonnet 5 attempts. An agent can submit its mitigation before diagnosis grading finishes. Previously, the API returned success, but the Conductor discarded the early mitigation. The attempt then ended without a mitigation result.

This PR tracks the stage of each submission. If mitigation arrives during diagnosis grading, the API waits until the mitigation stage starts. It returns success only after the Conductor accepts the submission. Older clients that do not send a stage still work.

The runner now marks attempts with missing results as incomplete. It also waits for grading and cleanup before it starts the next attempt. If cleanup cannot finish safely, the runner stops the campaign. Resume runs also retry incomplete attempts. Evaluation reports do not count incomplete attempts as model successes or failures.

---

- The submission API now accepts an optional `stage` value: `diagnosis` or `mitigation`.
- Each pending submission is linked to its benchmark attempt. Submissions from an old or closed attempt are rejected.
- An early mitigation request waits for up to 300 seconds. The API checks every 100 milliseconds for the mitigation stage.
- The Conductor accepts only one submission for the active stage. The API reports success only after this acceptance.
- The runner waits for pending submissions, grading tasks, and cleanup tasks before it saves the final result.

### DIFF
--- a/.github/workflows/problem-difficulty-evaluation.yml
+++ b/.github/workflows/problem-difficulty-evaluation.yml
@@ -458,7 +458,7 @@ jobs:
             exit 1
           fi
 
-          uv run python -m sregym.problem_evaluation_report \
+          uv run python -m sregym.results.report \
             --csv "$result_csv" \
             --problem "$PROBLEM_ID" \
             --model "$ZAI_MODEL" \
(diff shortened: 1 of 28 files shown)


tianyin (approved):
I read it. It's a big diff, but it looks good to me.

PR #988, opened 2026-08-31
Improve stability and resource issues via tuning resources and disabling unused things

niallrmurphy:
I've been working on making SREGym a bit stabler, and a bit more able
to fit within my local resource limits.  Accordingly, I've been
looking around for misconfigurations, inefficiencies, and situations
where we expend effort for no particular reason. I attempted to have
these fixes not change what is observable by an agent, but instead to
make existing config work, work better,	   or remove it where irrelevant.

### 1. metrics-server's resources incorrectly specified

`reconcile_to_baseline` ignores cluster-scoped resources named `system:*`
(`cluster_state.py:224,240`), but metrics-server's binding starts with
**`metrics-server`**, meaning that the **`:system:auth-delegator`** postfix
which should get matches, doesn't. It should break, but doesn't because
`conductor.py:798` re-applies all of `components.yaml` for every problem.
If you end up skipping the re-apply, `kubectl top` fails with

```
Internal Server Error: "/apis/metrics.k8s.io/v1beta1/pods": subjectaccessreviews.authorization.k8s.io
is forbidden: User "system:serviceaccount:kube-system:metrics-server" cannot create resource
"subjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope
```

### 2. Preserve the OpenEBS tier across problems

OpenEBS seems not to hold any relevant state to problem resolution, so nuke-and-repave doesn't
buy additional insulation from leakage or overfitting. 

Namespace, cluster RBAC, CRDs and StorageClasses are preserved together; preserving the namespace
alone would leave a provisioner running but unable to act, which is harder to diagnose than a
clean rebuild.

**`observe` is deliberately not preserved.**

### 3. Skip re-applying manifests that are already present

If we have (1) and (2), then `metrics-server` and the OpenEBS operator
no longer need re-fetching from GitHub every time.

### 4. Skip `helm dependency update` when dependencies are satisfied

`helm.py:39` runs `dependency update` unconditionally for local
charts, and doesn't need to, and it's failures were dropped (since
output was assigned but not read.)

### 5. Three Loki values keys don't exist in chart 7.3.0

`sregym/observer/loki/loki-values.yaml` is written against an older
chart, so the canary DaemonSet ran on every node despite the file
disabling it, and Loki's CPU/memory limits were never applied. Moved
to their real paths.

Also correctly sized `chunksCache`/`resultsCache`, which used (large)
chart defaults for sizing, rather than the actual (much smaller) RSS,
which made scheduling harder.

### 6. astronomy-shop's collector exports to a Prometheus we disable

`AstronomyShop.deploy()` sets `prometheus.enabled=false` — correctly,
since the bundled one should be disabled in favour of the central one.
But the bundled collector's metrics pipeline still points to the
disabled one, so everything fails and the retry queue grows indefinitely.

### Also: bound the central Jaeger

`observer/jaeger/jaeger.yaml` runs `all-in-one` with in-memory
storage, no `--memory.max-traces` and no resource limits, so the span
store grows indefinitely, while the jaegar chart shipped with
astronomy-shop has limits.

Happy to break this into smaller things as required.


sregym/service/apps/astronomy_shop.py
```diff
@@ -48,6 +50,10 @@ def deploy(self):
             "components.load-generator.envOverrides[0].name=LOCUST_BROWSER_TRAFFIC_ENABLED",
             "--set-string",
             "components.load-generator.envOverrides[0].value=false",
+            # Keeps the collector's exporters consistent with the components that
+            # are actually deployed; see the file's own comments.
+            "-f",
+            str(self._VALUES_DIR / "astronomy-shop-fixes.yaml"),
```
Saadmrp1038, line 56: This code resolves the path as `sregym/service/apps/values/astronomy-shop-fixes.yaml`. However, it seems that the file is added at the repository root.

As a result Helm is failing before it creating any pods:

```bash
Error: INSTALLATION FAILED: open .../sregym/service/apps/values/astronomy-shop-fixes.yaml: no such file or directory
```

Saadmrp1038 (changes requested):
@niallrmurphy Thanks for the PR!

I tested the Astronomy Shop changes on the live cluster and found one run-blocking problem.

The remaining changes look good. I did not find any changes that directly break existing faults.

niallrmurphy:
Ah yes, I actually saw this previously but blew away the fix when
uploading, really competently ;-/ PTAL!

Saadmrp1038 (approved):

Saadmrp1038:
image

PR #991, opened 2026-09-01
Fix Kubernetes API proxy upgraded connections

### LINKED ISSUE #982: kubectl exec and kubectl cp fail through the Kubernetes API proxy
`kubectl exec` and `kubectl cp` fail when traffic goes through SREGym’s Kubernetes API proxy. Normal commands such as `kubectl get` and `kubectl logs` still work.

Across my 52 saved local runs, agents used `kubectl exec` in 23 of them, totaling 105 invocations. **None succeeded.** None of these runs used `kubectl cp`, but I reproduced it separately. It failed through the proxy and succeeded against the same cluster using a direct kubeconfig.

Agents sometimes switched to other commands and still finished the run, but these failed attempts wasted limited time and context.

`kubectl exec` requires an upgraded two-way stream between `kubectl` and the Kubernetes API server. The proxy forwards the initial request but closes the upgraded connection instead of relaying the stream, which causes the EOF. `kubectl cp` also fails because it copies files through an exec-based `tar` stream.

## Reproduction

Run this through the SREGym proxy.

```bash
kubectl exec -n <namespace> <pod> -- echo ok
```

It fails with

```text
websocket: close 1006 (abnormal closure): unexpected EOF
```

The same command against the same cluster using a direct kubeconfig prints `ok`.

munimthahmid:
## Summary

Fixes #982.

The Kubernetes API proxy treated upgraded connections like normal HTTP responses. Commands such as `kubectl exec`, `kubectl cp`, and `kubectl port-forward` switch to a bidirectional stream, so the proxy closed them early and returned WebSocket EOF errors.

This change relays upgraded connections in both directions, preserves upgrade headers and buffered bytes, and closes both sockets when the stream ends. Ordinary API requests continue using the existing response path.

## Testing

- Added regression tests for WebSocket and SPDY upgrades, buffered data, rejected upgrades, and ordinary responses
- Verified live `exec`, `cp`, `port-forward`, large transfers, concurrent streams, and interrupted streams
- Ran two Luna Max attempts on `node_conntrack_exhaustion_hotel_reservation`. Across 77 expanded agent-side `kubectl exec` invocations, there were no WebSocket EOF or 1006 proxy failures

### DIFF
--- a/sregym/service/k8s_proxy.py
+++ b/sregym/service/k8s_proxy.py
@@ -20,9 +20,11 @@ import json
 import logging
 import os
 import secrets
+import socket
 import ssl
 import tempfile
 import threading
+from collections.abc import Callable
 from datetime import UTC, datetime, timedelta
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from ipaddress import ip_address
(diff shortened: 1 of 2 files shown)


Saadmrp1038:
LGTM
Thanks for the fix!

Saadmrp1038 (approved):

PR #992, opened 2026-09-02
Add: unit_mismatch_gomemlimit_astronomy_shop problem

### LINKED ISSUE #970: [Problem RFC] Invalid GOMEMLIMIT unit suffix crash-loops Go services
## Summary

Add an Astronomy Shop problem in which a Go service is configured with an invalid `GOMEMLIMIT` unit suffix. For example, changing checkout from `16MiB` to `16MB` causes the Go runtime to reject the value during startup, so the replacement pod crash-loops and checkout traffic fails.

Proposed public problem ID: `unit_mismatch_gomemlimit_astronomy_shop`.

## Why this is a new problem class

SREGym already contains memory-pressure and OOM scenarios, but this failure is neither. The requested memory quantity is reasonable and the node has capacity; the process never starts because the runtime configuration uses an unsupported unit. Diagnosis therefore depends on correlating a startup parse error with one environment variable instead of adding memory or chasing node pressure.

The Go runtime accepts byte values with `B`, `KiB`, `MiB`, `GiB`, and `TiB` suffixes. SI-looking values such as `16MB` are not accepted. Astronomy Shop already configures `GOMEMLIMIT` on checkout, product-catalog, and flagd, which gives the mechanism natural sibling variants without inventing a component.

## Proposed incident

Start from an otherwise healthy Astronomy Shop deployment. Override checkout's `GOMEMLIMIT` from `16MiB` to `16MB` and force a rollout or pod replacement so the invalid value is exercised.

Expected observations:

- the new checkout pod exits during Go runtime initialization and enters `CrashLoopBackOff`;
- logs identify an invalid `GOMEMLIMIT` value;
- node memory, pod memory requests, and limits remain healthy;
- checkout behavior fails while unrelated storefront paths remain available.

The expected diagnosis is the invalid unit suffix on `deployment/checkout`, not generic memory exhaustion.

## Valid mitigations

Accept repairs that restore a valid Go memory-limit value with suitable capacity, or deliberately remove the bad override when leaving the runtime default is safe. Do not require the exact original string if another semantically valid value restores behavior.

The mitigation should reject scaling the service to zero, deleting the workload, changing the application image, or merely increasing Kubernetes memory limits while the invalid value remains.

## Evaluation requirements

- Verify a real checkout transaction or equivalent checkout RPC, not only pod readiness.
- Require the checkout deployment to become stable with the intended replica count.
- Preserve unrelated deployment configuration and images.
- Re-run the scenario with at least one sibling service/value as a hidden robustness variant.
- Confirm that fault injection deterministically replaces the old healthy pod; a RollingUpdate must not hide the crash-loop behind an old Ready replica.

## Prototype status

Commit [`a443f165`](https://github.com/SREGym/SREGym/commit/a443f165b17aa83639ac997d753b93e4e5d10430) contains an initial implementation with checkout, product-catalog, and flagd variants. It should be treated as a prototype: the final problem still needs end-to-end workload validation and an oracle that accepts multiple correct configurations.

Go's supported suffixes are documented in [`runtime/debug.SetMemoryLimit`](https://pkg.go.dev/runtime/debug#SetMemoryLimit).

sidbhanushali:
Closes #970

## Problem

Adds `unit_mismatch_gomemlimit_astronomy_shop`, targeting `deployment/checkout` in `astronomy-shop`, plus a `product-catalog` robustness variant.

`GOMEMLIMIT` is rewritten to the SI-suffixed form of whatever the chart currently sets (`16MB` for `16MiB`). The Go runtime accepts only the IEC suffixes `B`, `KiB`, `MiB`, `GiB`, `TiB`, so it exits with `fatal error: malformed GOMEMLIMIT` before `main()` runs and the pod crash-loops, while node memory, pod memory limits and the rest of the deployment stay healthy.

## Real-world failure story

This is the software form of a unit-mismatch interface failure, the class whose canonical instance is the Mars Climate Orbiter: two sides of an interface agree on a number and disagree on its unit.

Go's parser is strict in a way that catches real engineers. It accepts only `^[0-9]+((KiB|MiB|GiB|TiB|B))?$`, so it rejects the SI-looking `MB` used everywhere else in computing, and the `Mi` that Kubernetes itself writes two lines below in the same manifest block. [golang/go#76923](https://github.com/golang/go/issues/76923) is an open issue filed after someone set `GOMEMLIMIT=1.5GiB` and got `fatal error: malformed GOMEMLIMIT`, arguing the documented phrase "a numeric value in bytes with an optional unit suffix" reads as though their value should work. The failure mode is a value that looks correct to a human and is refused by the runtime before `main()`.

The accepted syntax below was established by running the real checkout binary across a matrix of values rather than from documentation, since two of the results shape the oracle:

| Value | Go runtime |
| --- | --- |
| `16MiB`, `16KiB`, `16GiB`, `16TiB`, `16B`, `16`, `16777216`, `0`, `off`, unset | accepted |
| `16MB`, `16Mi`, `16mib`, `OFF`, `-1`, `16.5MiB`, ` 16MiB`, `16MiB `, `16EiB` | `fatal error: malformed GOMEMLIMIT` |

`off` is accepted despite not appearing in the `runtime` documentation, so an oracle written from the docs alone would reject a legitimate fix. `0` and `16B` parse cleanly, which is why the oracle carries a capacity floor rather than trusting syntax.

## Fault injection

Injection reads the value the chart currently sets and aborts if it has drifted from what the root-cause text describes, so the text handed to the judge cannot diverge from what was injected.

Pod replacement is deterministic: scale to zero, patch, scale back. A default RollingUpdate would keep the old Ready replica alive and hide the crash-loop behind it. Injection then blocks until a target pod is crash-looping **with the parse error present in its log**, so a crash from any other cause cannot be mistaken for the injected fault.

## Observable symptoms

- the target pod enters `CrashLoopBackOff`; its log shows `GOMEMLIMIT=16MB` followed by `fatal error: malformed GOMEMLIMIT`
- node memory, pod memory requests and limits are untouched and healthy
- the rest of the storefront keeps serving; only paths through the faulted service fail

## Expected mitigation

Set the memory limit to any value the runtime accepts with reasonable capacity, or drop the override entirely. `16MiB`, `32MiB`, `1GiB`, `16777216`, `off` and removing the variable all pass. The original string is not required.

Rejected: scaling to zero, deleting the workload, changing the image, raising the Kubernetes memory limit, and syntactically valid but unusable values such as `0` or `16B`.

## Mitigation oracle

State checks, all compared against a baseline captured before injection rather than against hard-coded chart values, so they remain correct on a cluster whose chart differs from upstream:

- the configured value parses as valid Go syntax and clears a capacity floor
- replicas at or above baseline, rollout complete, pods Running and Ready
- serving pods belong to the deployment's **active** ReplicaSet, so a survivor from an older ReplicaSet still carrying the pre-fault environment cannot satisfy the oracle
- image and resource requests/limits unchanged from baseline

Then behaviour, since neither target defines a readiness probe and a Running pod therefore proves nothing on its own: checkout adds an item to a cart and places a real order through `oteldemo.CheckoutService/PlaceOrder`, asserting a returned `orderId`. checkout does not serve gRPC reflection, so the probe supplies a minimal descriptor through a ConfigMap. The `product-catalog` variant uses `ListProducts`, which does serve reflection.

A completed order crosses cart, currency, payment, shipping, email, product-catalog and Kafka. The probe therefore allows 45s per call and retries three times before reporting failure; a tighter single attempt produced a false negative against a correct repair during agent testing, so the margin is deliberate. Regression tests cover both the retry and the give-up paths.

## Recovery

`recover_fault` restores the pod template captured at injection time verbatim, along with the original replica count, then waits for the rollout to complete. Nothing is reconstructed from constants.

## Testing and validation

All of the below ran against the **unmodified upstream chart**.

End-to-end lifecycle (`tests/integration/validate_problem.py`), 13 runs, 13 passes, with the oracle failing on its first check after injection and passing after recovery in every run:

| Variant | Runs | Result |
| --- | --- | --- |
| `unit_mismatch_gomemlimit_astronomy_shop` | 7 | all PASSED |
| `unit_mismatch_gomemlimit_product_catalog_astronomy_shop` | 6 | all PASSED  |

Each documented cheat was then applied to the live deployment and the oracle's verdict recorded:

| Attempt | Oracle verdict |
| --- | --- |
| leave `16MB` in place | rejected: "GOMEMLIMIT is still a value the Go runtime rejects" |
| scale the deployment to zero | rejected: "deployment/checkout is scaled to 0" |
| raise the Kubernetes memory limit, leave the bad value | rejected: "the container resource requests or limits were changed from their pre-fault values" |
| swap the container image | rejected: "the container image was changed from its pre-fault value" |
| set `16B` (valid syntax, absurd capacity) | rejected: "GOMEMLIMIT parses but is too small to run the service (16 bytes)" |
| set `32MiB` (correct, not the original string) | passed |
| remove the override entirely | passed |

Unit tests: `uv run pytest tests/oracles/test_unit_mismatch_gomemlimit_mitigation.py tests/problems/test_unit_mismatch_gomemlimit.py` (50 tests covering the parser matrix, one per cheat, and the probe-retry paths). Prek hooks pass.

## Agent run results

Stratus with `anthropic/claude-sonnet-4-6`, chosen to match the leaderboard row for that pairing (61.5 diagnosis / 78.5 mitigation / 54.8 E2E across the suite):

| Stage | Result |
| --- | --- |
| Diagnosis | 100.0, judged True, TTL 61s |
| Mitigation | Pass, restored `GOMEMLIMIT=16MiB`, real order placed |

The problem sits on the easier end for that agent: 100% diagnosis in about a minute against its 61.5 suite average, since the crash log names the offending value outright. If a harder variant is wanted, `16Mi` is the subtler and arguably more realistic typo given Kubernetes writes memory quantities that way two lines below in the same manifest. The wrong value is a parameter, so that variant is a one-line change.

## Code changes

| File | Change |
| --- | --- |
| `sregym/conductor/problems/unit_mismatch_gomemlimit.py` | Problem class, deterministic injection with parse-error self-verification, template-restoring recovery |
| `sregym/conductor/oracles/unit_mismatch_gomemlimit_mitigation.py` | Oracle: semantic validation, capacity floor, baseline-relative negative checks, real-request probe |
| `tests/oracles/test_unit_mismatch_gomemlimit_mitigation.py` | Parser matrix, one test per cheat, probe-retry regressions |
| `tests/problems/test_unit_mismatch_gomemlimit.py` | Injection guards and root-cause shape |
| `sregym/conductor/problems/registry.py` | Registers both variants |
| `Problem List.md` | Catalog rows |

## Design note

The oracle treats a change to the container's resource limits as a failure, reading "preserve unrelated deployment configuration" strictly, so a repair that fixes the value and also bumps memory would be rejected. That is deliberate, and easily loosened to "reject only while the value is still invalid" if you prefer the weaker reading.

`flagd` is not included as a variant; `product-catalog` covers the sibling-variant requirement and uses the same evaluation path.

/validate-problem unit_mismatch_gomemlimit_astronomy_shop


HacksonClark:
From the action above, it seems that this problem is saturated. While I still think this problem is well-structured and implemented well (it meets the bar for participation in the REU program for sure), it is currently saturated and is not difficult enough to include in the benchmark.

This is a great contribution and I encourage you to contribute more! But I have to close this for now.

PR #995, opened 2026-09-02
Add profile mechanism to drop unused components

niallrmurphy:
SREGym deploys several components that never get used, and that
overhead is occasionally meaningful. This PR adds a flag, `--profile
{full,svelte}` which can remove the overhead, and this makes the
overall suite considerably more likely to successfully conclude in
constrained environments (particularly memory-constrained ones).

It defaults to `full` and is opt-in for now (though that creates some
maintenance problems) since dropping components strictly changes what an
agent can observe: scores under `svelte` are not comparable with `full`.
(We chose "svelte" to avoid name collision with the "lite" problem set.
Of course, the choice of the problem set is orthogonal to whether or not the
infra is fully working/specified/etc.)

## What `svelte` drops

**astronomy-shop's bundled OpenSearch and Grafana.** Nothing appears to query either.

**astronomy-shop's bundled Jaeger.** It is set up but then overridden to point at the
central one.

**Prometheus Alertmanager and Pushgateway**, plus a retention trim.
No alert rules are configured and nothing pushes to a gateway.
(Conversely, `kube-state-metrics`, `node-exporter`
and `blackbox-exporter` are all used and kept.)

**The OpenEBS node-disk-manager**, which exists to back the
`openebs-device` StorageClass.  SREGym only provisions through
`openebs-hostpath`.

## Measurements

`env_variable_shadowing_astronomy_shop` on a 4-node kind cluster (macOS/OrbStack, 18 vCPU,
18 GiB). Peak of six `kubectl top` samples at 20s intervals.

| | peak RSS | peak CPU | pods | deploy |
|---|---|---|---|---|
| `full` | 6,711–8,104 MiB | 2,094–5,191m | 75–79 | ~174s |
| `svelte` | 5,182 MiB | 1,798m | 63 | 155s |

Roughly a third less memory and around 60% less CPU, but there is some sampling
variance.

| component | peak RSS | peak CPU |
|---|---|---|
| `opensearch-0` | 1,096 MiB | 1,709m — **39% of all cluster CPU** |
| `grafana` | 475 MiB | 303m |
| OpenEBS NDM (7 pods) | 246 MiB | 603m — 13% of cluster CPU |

N=1 per configuration on one problem and one machine, so treat the totals as indicative.

## Questions

1. Is an opt-in resource profile something you want in-tree at all, or is this better as a
   downstream patch?
2. If yes, is `--profile` the right name?

## Notes

Commits carry `Co-Authored-By` trailers: implementation written with a coding agent, then
reviewed, deployed and measured on a real cluster per `CONTRIBUTING.md:42`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,44 @@ uv run main.py --agent codex --model gpt-5 --force-build
 Containerized agents can use the public internet by default, but direct access to the benchmark's GitHub source is
 blocked. Use `--internet-access open` only when you intentionally need the previous unrestricted network behavior.
 
+### Deployment Profiles
+
+`--profile` controls how much infrastructure SREGym stands up. It is independent of
+`--suite` — the profile selects *what gets deployed*, the suite selects *which problems run*.
+
+| Profile | Behaviour |
+|---------|-----------|
+| `full` (default) | The standard stack. Use this for results you intend to compare against the leaderboard. |
+| `svelte` | Additionally drops components that nothing in SREGym reads, and shortens metric retention. |
+
+```bash
+uv run main.py --suite sregym-lite --agent stratus --model gpt-5 --profile svelte
+```
+
+`svelte` removes:
+
+- astronomy-shop's bundled **OpenSearch**, **Grafana** and **Jaeger**. Nothing in `sregym/`,
+  `mcp_server/` or `clients/` queries OpenSearch or Grafana; the bundled Jaeger is deleted
+  moments after deployment anyway, by `Jaeger.create_external_name_service()`.
+- Prometheus **Alertmanager** and **Pushgateway** (no alert rules are configured and nothing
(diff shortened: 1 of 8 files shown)


Saadmrp1038 (approved):

Saadmrp1038:
@niallrmurphy Thanks for the PR!

I think keeping this profile in-tree makes sense, and `--profile` works for me. Since `svelte` mode changes the environment visible to agents, it may affect agent pass rates. We will need to test that out.

niallrmurphy:
Thanks, this is great. My instinct is that it's hard to introduce two ways of running a piece of software with very different expectations and keep that going indefinitely - generally speaking there's pressure to consolidate in a way which has enough of the benefits of both - but I think in this specific case we should have this option to test and explore the performance/results manifold, potentially for quite a long time. Which was the motivation for the original patch :)

Looking forward to seeing your results!

PR #996, opened 2026-09-03
Separate evaluation and cleanup deadlines

Saadmrp1038:
## Problem

Two 63-attempt (21 faults x 3 attempts each) SREGym-Lite campaigns stopped at `kafka_poison_pill_hol_block`.

In each campaign, 18 attempts completed, one was marked incomplete, and 44 never started. The agent submitted both stages, and the mitigation evaluation succeeded.

However, the runner recorded `evaluation_timeout_after_agent_exit` and stopped the campaign.

The runner used one 300-second deadline for evaluation and cleanup. Cleanup includes fault recovery, application removal, and cluster reconciliation.

The Terra run spent 130 seconds in evaluation and 183 seconds in cleanup. The Luna run spent 131 seconds in evaluation and 184 seconds in cleanup.

Each phase finished within 300 seconds. However, their combined times were 313 and 315 seconds. The old shared deadline therefore stopped both campaigns incorrectly.

## Change

This PR gives evaluation and cleanup separate 300-second deadlines.

The Conductor returns control when evaluation finishes and cleanup begins. The runner then waits for the same cleanup task with a new deadline.

The runner does not start another attempt until cleanup finishes. It also records evaluation and cleanup timeouts separately.

### DIFF
--- a/main.py
+++ b/main.py
@@ -42,7 +42,8 @@ logger = logging.getLogger(__name__)
 _driver_results: list[dict] = []
 _driver_base_dir: Path | None = None
 _driver_error: BaseException | None = None
-SUBMISSION_DRAIN_TIMEOUT_SECONDS = float(os.getenv("SUBMISSION_DRAIN_TIMEOUT_SECONDS", "300"))
+EVALUATION_DRAIN_TIMEOUT_SECONDS = 300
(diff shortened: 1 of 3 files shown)


tianyin (approved):
LGTM

PR #997, opened 2026-09-03
Disable Codex web search in filtered mode

Saadmrp1038:
A Codex agent could use hosted web search to access the SREGym repository. This request runs outside the agent container, so the network proxy could not block it. This exposed the real fault name and caused some runs to be rejected (we auto reject runs if the agent traces contain unmasked fault id).

This change removes the hosted web-search tool from Codex when filtered internet access is active. Shell commands still use the existing network proxy. Open-mode runs remain unchanged.


tianyin (approved):
great work

PR #998, opened 2026-09-03
Block Codex hosted tools in filtered runs

Saadmrp1038:
During a filtered Codex benchmark run, the agent accessed the SREGym GitHub repository through a hosted GitHub tool. This exposed the fault information and invalidated the result.

The recent PR (#997) disabled native web search. However, Codex Apps and plugins remained enabled. These tools run through OpenAI services, so the local network proxy cannot block their requests.

This PR disables apps and plugins when `AGENT_INTERNET_ACCESS=filtered`. This prevents hosted tools from bypassing the network policy. Runs with open internet access remain unchanged.


tianyin (approved):
Sharp!

PR #1003, opened 2026-09-05
Restrict agent internet access

Saadmrp1038:
## Summary

The existing GitHub denylist still allowed agents to download upstream application code from other internet sources. This can reveal fault solutions and affect benchmark results.

This PR makes filtered runs block general internet access by default. Agents retain access to their model provider, required authentication endpoints, and internal SREGym and Kubernetes services. Users can allow extra destinations with `--allow-agent-endpoint`, or select `--internet-access open` for unrestricted access.

## Implementation

- **Agent network:** Agent containers use an internal Docker network. A separate HTTP/HTTPS proxy provides outbound access and permits only configured destinations. The runner installs the required proxy configuration and certificates before preflight.

- **Kubernetes pod network:** A Calico policy blocks external traffic from ordinary application and diagnostic pods. This also covers internet requests through `kubectl exec`. Internal traffic continues through existing network policies, so network-policy faults retain their controls. CoreDNS keeps the upstream DNS access it needs.

- **Kubernetes API controls:** The Kubernetes proxy protects the egress policy and rejects newly added host-network, privileged-container, and host-filesystem access. Ordinary temporary pods, Jobs, replacement workloads, and permitted network-policy changes remain available.

- **Provider configuration:** A shared resolver builds the allowlist from the selected agent, model, authentication method, and custom endpoint configuration. Explicit exceptions can restrict access by host, port, and URL path.

- **Runtime dependencies:** Affected applications and fault helpers use prebuilt AMD64 and ARM64 images instead of downloading dependencies during startup. These include Redis, Kafka, TLS helpers, Social Network assets, and the Grafana plugin. The new MCP image contains only its required runtime code, without fault definitions or oracle implementations.

The PR also fixes Kafka producer-leak timing. Injection waits for an actual OOM, and mitigation checks that the current Kafka pods remain healthy during an observation window.

## Setup

Filtered mode requires Calico 3.29 or later. The kind setup script installs the required version, and the documentation includes upgrade instructions for existing clusters.

---

Related PR: https://github.com/SREGym/SREGym-applications/pull/8

/validate-problem skip


HacksonClark (approved):
Tested, code is working well enough on my machine. It's technically not fully comprehensive in terms of stopping all traffic but it should stop any potential reward hacks. Fixing merge conflicts now, then if the smoke tests pass I will merge.

PR #1004, opened 2026-09-05
Make host-facing ports configurable, and stop killing foreign processes.

niallrmurphy:
Certain ports (API_PORT, MCP_SERVER_PORT, k8s proxy 16443) were effectively
hard-coded, meaning that if you happened to be running SREGym while something
else was sitting on them, bad things would happen. Specifically, the reclamation
procedure assumed that whatever was listening on that port according to `lsof`
could be killed since it was launched by SREGym, which is not the case.

Some minor tweaks to process reporting across multiple OSes.


Saadmrp1038:
LGTM

Saadmrp1038 (approved):

Issue #971, opened 2026-08-25
[Problem RFC] PostgreSQL identity sequence exhaustion blocks product-review writes


HacksonClark:
## Summary

Add an Astronomy Shop problem in which the PostgreSQL identity sequence backing `reviews.productreviews.id` reaches the `INTEGER` limit. Existing reads remain healthy, but new product-review writes fail when `nextval()` can no longer produce a representable ID.

Proposed public problem ID: `integer_overflow_primary_key_astronomy_shop`.

## Why this is a new problem class

This is a data-lifetime failure rather than a database outage, capacity limit, or connectivity fault. PostgreSQL remains Running and Ready, reads can continue, and only a write path fails. The agent must inspect the table type and identity sequence state, then restore safe ID headroom without corrupting existing rows.

Astronomy Shop is a natural target because `reviews.productreviews.id` is currently an `INTEGER GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY`.

## Proposed incident

Start from a healthy Astronomy Shop deployment with existing review data. Advance `reviews.productreviews_id_seq` to its maximum safe `int4` value, then exercise the normal product-review write path until PostgreSQL returns a sequence-maximum or integer-range error.

Expected observations:

- storefront and review reads can remain healthy;
- attempts to create a new product review fail;
- the PostgreSQL pod remains Ready;
- database logs or a direct sequence inspection show that the identity generator has exhausted its range.

The expected diagnosis must identify the exhausted sequence and the narrow `INTEGER` ID type, rather than reporting only that product-reviews or PostgreSQL is failing.

## Valid mitigations

The durable repair is a data-safe migration of the ID column and its sequence to `BIGINT`. A temporary sequence change may be accepted only when it proves that a collision-free unused range exists and leaves substantial headroom. Blindly resetting the sequence to a hard-coded low value is not safe because it can collide with existing primary keys.

## Evaluation requirements

- Verify both pre-existing reads and a new write through the application behavior.
- Require substantial, collision-free sequence headroom after mitigation, or verify a correct `BIGINT` migration.
- Preserve all existing review rows and primary-key uniqueness.
- Reject deleting/truncating the table, dropping constraints, disabling the writer, or modifying the oracle's probe data.
- Run more than one post-injection insert. In the current prototype, setting the sequence to `2,147,483,646` still permits the next value `2,147,483,647`; exhaustion is observed only on the following `nextval()`.
[...]

PR #1005, opened 2026-09-06
postgres_integer_overflow_primary_key problem (astronomy-shop)

saihariG:
Implements #971 

An `INTEGER` (int4) primary key or identity sequence has a hard ceiling of 2,147,483,647. On a high-write table the sequence eventually exhausts, and every new `INSERT` that allocates an id via `nextval()` begins failing with `reached maximum value of sequence`, while existing reads keep working, the classic "database is up but half-broken" outage that fools responders looking at pod/CPU health. 

Relevant PostgreSQL behavior is documented under [numeric types](https://www.postgresql.org/docs/current/datatype-numeric.html) and [sequence functions](https://www.postgresql.org/docs/current/functions-sequence.html).

App: **astronomy-shop** 
The `product-reviews` service writes to `reviews.productreviews`, whose `id` is an INTEGER GENERATED BY DEFAULT AS IDENTITY column backed by `reviews.productreviews_id_seq`.

- **`inject_fault()`:** pins the sequence at int4 max (setval(seq, 2147483647, true)) so the next nextval() overflows, then confirms that a write actually fails and raises if the fault did not take.
- **`recover_fault()`:** applies the durable real-world fix, a schema migration widening the column and sequence to `BIGINT (ALTER COLUMN id TYPE bigint + ALTER SEQUENCE ... AS bigint)`, then verifies writes succeed again.
- **`Mitigation oracle`:** a rollback-wrapped `INSERT` probe (leaves no data) that accepts the fix only when a write succeeds. The default pod-health oracle cannot detect this since every pod stays Running.

- New product-review writes fail with `ERROR: nextval: reached maximum value of sequence "productreviews_id_seq" (2147483647).`
- Reads of existing reviews stay healthy; `product-reviews`and `postgresql` pods stay Running with normal CPU and memory.
- The cause is only visible inside the database (the `id` column's type and the sequence's `last_value`), not from pod health.
- Naive fixes like restarting or scaling pods is **NOT a valid** mitigation.

<table>
  <tr>
    <td align="center">
      
    </td>
    <td align="center">
      
    </td>
  </tr>
</table>

Screenshot 2026-09-06 at 12 48 55 AM

## **Notes**
- Recovery uses the durable `BIGINT` migration (the real SRE fix). A sequence reset is only a safe band-aid when the id space is collision-free.
- `astronomy-shop` is redeployed fresh each run, so the exhausted sequence never leaks across runs.
- The write-probe deliberately triggers a failing `INSERT` to detect exhaustion, so an expected `ERROR` line appears in the validation log on the exhausted path.

/validate-problem integer_overflow_primary_key_astronomy_shop


HacksonClark:
/evaluate-problem integer_overflow_primary_key_astronomy_shop

HacksonClark:
/evaluate-problem integer_overflow_primary_key_astronomy_shop

HacksonClark:
/evaluate-problem integer_overflow_primary_key_astronomy_shop

HacksonClark:
/evaluate-problem integer_overflow_primary_key_astronomy_shop

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
```diff
@@ -0,0 +1,152 @@
...
+    def inject_fault(self) -> bool:
+        logger.info("Injecting integer-overflow (sequence exhaustion) fault...")
+
+        # Pin the identity sequence at INT4 max (is_called=true) so the very next
+        # nextval() overflows and product-review writes start failing
+        self._run_sql(f"SELECT setval('{self.SEQUENCE}', {self.INT4_MAX}, true);")
+
+        # Fail if the fault did not actually take
+        # A 'write' must now fail
+        status = self._review_write_status()
```
Saadmrp1038, line 73: The current `product-reviews` service has no review-creation operation. Its RPCs only read reviews or answer questions. The injector’s SQL probe produces an error, but normal application traffic never attempts this write. So, the agent sees probe-generated SQL errors, but normal application traffic does not generate any error. Is there any existing application write path that we can use for this fault?

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
```diff
@@ -0,0 +1,152 @@
...
+            "ok"        - the insert succeeded (sequence has headroom)
+            "exhausted" - the insert failed because the id sequence overflowed
+            "other"     - some other failure (postgres restarting, table missing, ...)
+        """
+
+        sql = (
+            "BEGIN; "
+            "INSERT INTO reviews.productreviews (product_id, username, description, score) "
+            "VALUES ('PROBE', 'health_probe', 'oracle smoke', 5.0); "
+            "ROLLBACK;"
```
Saadmrp1038, line 131: I set the sequence to `2147483646`, leaving one available ID. The oracle returned `success=True`, consumed that ID, and failed immediately on the next evaluation. Leaving two or ten IDs also passed. Please verify that the repair leaves substantial unused ID capacity, not only that one insert succeeds.

sregym/conductor/oracles/integer_overflow_primary_key_mitigation.py
```diff
@@ -0,0 +1,36 @@
...
+        super().__init__(problem)
+
+    def evaluate(self) -> dict:
+        print("--- Mitigation Evaluation (product-reviews insert) ---")
+
+        status = self.problem._review_write_status()
+
+        if status == "ok":
+            logger.info("review.productreviews accepts write again; mitigation accepted!")
+            return {"success": True}
```
Saadmrp1038: After `TRUNCATE reviews.productreviews RESTART IDENTITY`, the oracle returned success=True. All 50 original reviews were gone, and the frontend returned an empty list. We should also check that the fix does not delete the original data.

sregym/conductor/oracles/integer_overflow_primary_key_mitigation.py
```diff
@@ -0,0 +1,36 @@
+import logging
+
+from sregym.conductor.oracles.base import Oracle
+
+logger = logging.getLogger(__name__)
+
+
+class IntegerOverflowPrimaryKeyMitigationOracle(Oracle):
```
Saadmrp1038: After the BIGINT repair, I scaled `product-reviews` to zero. The oracle still returned `success=True`, but the frontend returned HTTP 500. Please inherit from `MitigationOracle` and require `super().evaluate()` to pass before the database checks. That should check that replica count is not 0.

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
```diff
@@ -0,0 +1,152 @@
...
+
+        self.app.create_workload()
+
+    @mark_fault_injected
+    def inject_fault(self) -> bool:
+        logger.info("Injecting integer-overflow (sequence exhaustion) fault...")
+
+        # Pin the identity sequence at INT4 max (is_called=true) so the very next
+        # nextval() overflows and product-review writes start failing
+        self._run_sql(f"SELECT setval('{self.SEQUENCE}', {self.INT4_MAX}, true);")
```
Saadmrp1038, line 69: The description says restarting pods does not help. However, deleting only the PostgreSQL pod removed the fault, and the oracle returned `success=True`. The replacement pod initialized a fresh database because the deployment has no persistent data volume.

Saadmrp1038 (changes requested):

HacksonClark:
@saihariG please resolve @Saadmrp1038's comments and request another review.

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
saihariG: Fixed! The oracle now requires the id sequence to retain substantial headroom

sregym/conductor/oracles/integer_overflow_primary_key_mitigation.py
saihariG: Fixed! The oracle now verifies the original seeded reviews are still present.

sregym/conductor/oracles/integer_overflow_primary_key_mitigation.py
saihariG: Fixed! The oracle now inherits `MitigationOracle` and requires `super().evaluate()` to pass before the database checks. Scaling `product-reviews` to zero fails now.

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
saihariG: I'll make the exhausted state survive a restart by giving the reviews data persistent storage, so deleting the pod no longer re-seeds init.sql and resets the sequence. How does that sound to you?

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
saihariG: yeah you are right, `product-reviews` has no review-creation RPC, so right now the failing writes only come from the oracle probe. I'll add a lightweight writer that periodically inserts a review, so failed writes appear continuously during the run rather than only at the evaluation time. Do you have anything specific in your mind?

Saadmrp1038: > Do you have anything specific in your mind?

Not really. As long as the errors come from actual application writes, rather than only the oracle probe, that is fine with me.

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
Saadmrp1038: Yup, that sounds good. Deleting existing database data must not count as a valid mitigation. So, the oracle needs to catch those actions.

HacksonClark:
/evaluate-problem integer_overflow_primary_key_astronomy_shop

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
saihariG: I added a light-weight `review-writer` deployment that keeps inserting reviews every 5 seconds as the application's database user (`otelu`) into `reviews.productreviews` (the same table application writes to). It starts when `inject_fault()` runs and is removed by `recover_fault()`

Without this, the integer-overflow issue only becomes visible when the oracle tries to insert a review. with the writer running, failed review submissions happen continuously as a part of normal application traffic, simulating the problem easy to observe in the system.

sregym/conductor/problems/integer_overflow_primary_key_astronomy_shop.py
saihariG: I ended up taking a different approach here rather than the persistent-volume idea. Instead of attaching a PVC to the shared `astronomy-shop` Postgres-instance, I updated the oracle to catch destructive fixes directly.

when `inject_fault()` runs, it inserts a sentinel row (`username = seq_overflow_sentinel`) while the sequence is still healthy. During evaluation, the oracle checks that the row is still present. If it's missing, the mitigation is rejected.

This essentially covers the cases that we wanted to prevent. Restarting the Postgres pod reinitializes the database, restoring the original 50 rows and resetting the sequence so the writes work again. However, it doesn't recreate the sentinel row. The same is true for either `TRUNCATE` or `DROP`. In all of these cases, the oracle now fails with _"The fault sentinel row is missing",_  which assures the thing without having to make the shared database persistent.

Saadmrp1038 (changes requested):
@saihariG Thanks for the updates! After some more testing I found some other issues. 

| Scenario | Expected | Actual |
| --- | --- | --- |
| Add new reviews, then delete all 50 original reviews | Fail: original data must survive | Pass: the row count and sentinel checks do not detect the lost reviews |
| Repair the sequence but remove the application user’s INSERT permission | Fail: application writes are still broken | Pass: the root probe succeeds while the writer repeatedly reports permission errors |
| Remove the primary key without equivalent uniqueness protection | Fail: review IDs must remain unique | Pass: the database also accepts an INSERT with an existing ID |
| Reset the sequence toward occupied IDs | Fail: the available range is not collision-free | Pass: the active writer starts reporting duplicate-ID errors 29 seconds later |
| Use a descending sequence through unused negative IDs | Pass: writes work and substantial unused capacity remains | Fail: the headroom calculation assumes an ascending sequence |
| Rename the identity sequence after a correct BIGINT repair | Pass: the identity column and writes still work | Fail: the oracle looks for the old sequence name |
| Read the injected sentinel row | No explicit fault or oracle labels | The agent can read `seq_overflow_sentinel` and `fault marker row; do not delete` (just make it look like a normal review) |

I think creating a foolproof mitigation oracle for this fault that guarantees all future writes will succeed indefinitely is not realistic. Fixing possible edge cases should be sufficient.

saihariG:
> @saihariG Thanks for the updates! After some more testing I found some other issues.
> 
> Scenario	Expected	Actual
> Add new reviews, then delete all 50 original reviews	Fail: original data must survive	Pass: the row count and sentinel checks do not detect the lost reviews
> Repair the sequence but remove the application user’s INSERT permission	Fail: application writes are still broken	Pass: the root probe succeeds while the writer repeatedly reports permission errors
> Remove the primary key without equivalent uniqueness protection	Fail: review IDs must remain unique	Pass: the database also accepts an INSERT with an existing ID
> Reset the sequence toward occupied IDs	Fail: the available range is not collision-free	Pass: the active writer starts reporting duplicate-ID errors 29 seconds later
> Use a descending sequence through unused negative IDs	Pass: writes work and substantial unused capacity remains	Fail: the headroom calculation assumes an ascending sequence
> Rename the identity sequence after a correct BIGINT repair	Pass: the identity column and writes still work	Fail: the oracle looks for the old sequence name
> Read the injected sentinel row	No explicit fault or oracle labels	The agent can read `seq_overflow_sentinel` and `fault marker row; do not delete` (just make it look like a normal review)
> I think creating a foolproof mitigation oracle for this fault that guarantees all future writes will succeed indefinitely is not realistic. Fixing possible edge cases should be sufficient.

Thanks for the thorough testing @Saadmrp1038 ! I reworked the oracle and stress-tested covering all the scenarios.

1. The write probe now runs as the application user
2. The oracle captures the original reviews at fault injection and requires them to survive, so deleting the originals and padding with new rows is rejected.
3. The oracle requires a single column PK/UNIQUE constraint on id to still exist
4. Added a collision free check (the sequence must be positioned past every existing id), so a reset that collides on later inserts fail immediately until the writer collides
5. Headroom is direction aware, so a descending sequence passes
6. The sequence is resolved via `pg_get_serial_sequence` 
7. The marker row is an ordinary-looking review matched by content

Also I verified locally that all valid fixes still pass, and every cheat is rejected.

HacksonClark:
/evaluate-problem integer_overflow_primary_key_astronomy_shop

Saadmrp1038:
LGTM.
Thanks for the work @saihariG

Saadmrp1038 (approved):

PR #1006, opened 2026-09-06
Support better failure classifications

niallrmurphy:
(The next four PRs are part of a larger project to make SREGym more
capable of running reliably, having more detailed results, and
making it easier to compare those results.)

Sometimes mitigation fails. An interesting question is whether if
fails because the diagnosis was incorrect and it successfully executed
the wrong plan, or the diagnosis was incorrect and it unsuccessfully
executed the right plan, or infrastructure was broken and no plan
would ever have worked.  For comparison purposes, it's useful to be
able to distinguish between these outcomes.

To do this, we create several new failure classes, machinery around
making these shared, implement it for selected Oracles, and fix a misattributed
infra failure to harness error. Other work to tune will follow.


Saadmrp1038:
LGTM

Saadmrp1038 (approved):

PR #1007, opened 2026-09-07
Implement stages - separation of diagnosis and mitigation.

niallrmurphy:
This PR is about adding the idea of 'stages' to runs - in particular,
to support separating diagnosis from remediation. (This enables more
efficient testing, since in some cases we actually only care about
diagnosis.) You can specify `--stages diagnosis mitigation` (the
default) or `--stages diagnosis`; obviously specifying mitigation
without diagnosis makes no sense.

The command line flag will take precedence, but if not otherwise done,
`tasklist.yml` will be parsed much as today. (There is an argument for
deprecating/removing it, but I won't do that in this PR.)


niallm-ciroos:
(Change 2/4.)

Saadmrp1038 (approved):

HacksonClark (approved):
LGTM

PR #1008, opened 2026-09-07
Export `traces.db` to plain CSV/TSV for activity reports, observability, etc.

niallrmurphy:
The main design question here is, why use this database as opposed to
any other approach? Well, there is an existing visualizer/ directory
with related code, but it is informally deprecated and has not been
updated recently. `traces.db` works the same way across all agents
(since the output is translated by each agent) and is decoupled
sufficient to support future agents. More subtly, though direct
OpenTelemetry export is possible (and may yet be achieved in other
ways) there is a potential leak were we to require transit of the data
through the existing infrastructure. Secondly, why have a phased
exporter like this, rather than a streamer or specific formats that
are read from the database by each program desiring to read them?
The answer is of course to reduce coupling, and avoiding streaming
for now is perfectly acceptable for what is mostly a batch process.

This PR creates the exporter program itself and some tests.


niallm-ciroos:
(Change 3/4 FWIW.)

Saadmrp1038 (approved):

Saadmrp1038:
Works Great!

PR #1014, opened 2026-09-08
Retrofit failure classification across all oracles

niallrmurphy:
The base idea is as before - that we are attempting to make the overall process
more instrumentable, more diagnosable, and more comparable. One significant
component of this is being able to attribute a failure: whether it's the agent,
the environment, the harness, or we don't know.

In this PR we establish support for being able to differentiate these failure
types and then retrofit this across all of the existing oracles.

In more detail, we change required_deployment_missing and required_deployment_scaled_to_zero,
fix the fact that subclass FAILURE_CLASSES tables overrode MitigationOracle's instead of merging,
amend valkey_auth_mitigation so that an exec failure is distinguishable from an auth
failure, shore up kafka_producer_leak_mitigation so that ApiException is no longer
swallowed, and tease apart some 21 other oracles that just had a bare failure instead of
environment-versus-harness.

Stacked on #1006; that PR should land first.


Saadmrp1038 (approved):

Saadmrp1038:
Works great! Thanks for the PR!

I made some changes to keep failure classification consistent and separate probe/API errors from mitigation failures.

PR #1016, opened 2026-09-08
fix(astronomy-shop): give kafka broker headroom

### LINKED ISSUE #1015: kafka_poison_pill_hol_block never injects: admin CLI exec OOM-kills the Kafka broker (exit 137)
## Summary

`kafka_poison_pill_hol_block` cannot inject its fault. Every attempt fails almost as soon as injection begins, all three deploy retries are exhausted, and the problem is skipped. We saw this on two separate runs of the lite suite — no successful injection either time, and no fault ever reaching the cluster.

The injector runs the Kafka admin CLI — a second JVM — inside the running broker container, whose cgroup has no memory headroom. The kernel OOM-kills it and takes the broker down with it.

**Problem ID:** `kafka_poison_pill_hol_block`

## Steps to reproduce

```bash
uv run python main.py --problem kafka_poison_pill_hol_block
```

Deploy astronomy-shop, let the app reach a healthy state, then let the problem inject. The first Kafka admin command the injector issues is enough to trigger it; no special conditions are needed.

Equivalently, the failing operation on its own, against a healthy astronomy-shop:

```bash
kubectl exec -n astronomy-shop deploy/kafka -- env 'KAFKA_HEAP_OPTS=-Xms32m -Xmx64m' KAFKA_OPTS= \
  /opt/kafka/bin/kafka-consumer-groups.sh --bootstrap-server kafka:9092 --list
```

Watch the broker while it runs:

```bash
kubectl get pod -n astronomy-shop -l app.kubernetes.io/component=kafka -w
```

## Expected vs actual

**Expected:** the admin command completes, the injector seeds `orders-fulfillment` with a poison record, and `orders-validator` stalls at that offset — the fault the problem is designed to produce.

**Actual:** the admin command is SIGKILLed with exit 137 and the broker container restarts. All three deploy retries fail the same way, the problem is skipped, and no fault is ever injected. `orders-validator` and `orders-fulfillment` never come into existence.

## Environment

- Kubernetes v1.31.14, `kubeadm` cluster on Ubuntu 22.04 nodes (EC2), Calico CNI, cri-dockerd runtime
- Python 3.12
- astronomy-shop from the vendored `opentelemetry-demo` chart (`SREGym-applications` → `astronomy-shop`), unmodified kafka values

We do not believe any of this is load-bearing: `limits.memory` is a cgroup limit enforced per-container by the kernel, so node size and distribution should not change the outcome. See the environment caveat below for the one difference that might.

## Symptoms

Injection fails almost as soon as it starts, in one of two ways:

```
ERROR all.infra.kubectl - Command failed (exit 137):
  kubectl exec -n astronomy-shop kafka-<hash> -- env 'KAFKA_HEAP_OPTS=-Xms32m -Xmx64m' KAFKA_OPTS= \
    /opt/kafka/bin/kafka-consumer-groups.sh --bootstrap-server kafka:9092 --delete --group orders-validator
  stderr: command terminated with exit code 137
❌ start_problem failed for 'kafka_poison_pill_hol_block' (deploy attempt 1/3)
```

and:

```
  stderr: error: Internal error occurred: unable to upgrade connection:
          container not found ("kafka")
❌ start_problem failed for 'kafka_poison_pill_hol_block' (deploy attempt 3/3)
⛔ All 3 deploy attempts failed for 'kafka_poison_pill_hol_block', skipping this attempt
```

Exit 137 is SIGKILL from the kernel's cgroup OOM killer. The second form is the same event observed one step later: the broker container was already gone when the next `kubectl exec` arrived.

Because the fault never lands, the run produces no observable symptom at all — no `orders-validator` deployment, no `orders-fulfillment` topic, no consumer lag. Anything consuming the run's telemetry sees a healthy application and then nothing.

## Evidence: the exec is the trigger

We tracked container-restart counts for the `kafka` container across one affected run. The broker records **zero restarts** for the whole time it is simply serving traffic, and then exactly one restart in each of the three windows containing an injection attempt — no restarts anywhere else in the run.

The broker is stable until something execs into it, so this is not background memory pressure — the exec is the trigger.

Note the broker is killed on **all three** attempts, not only the one that reports `container not found`. At the granularity we sampled we cannot resolve the ordering within an attempt (whether the CLI dies first and the broker follows, or the reverse), but the broker does not survive any of them.

**Environment caveat.** Our harness holds the app under load for an extended period before injecting, whereas the default flow injects much sooner after deploy. We do not think this changes the outcome, since `-Xms400M` commits the broker's heap at startup and the shortfall is roughly 100–150Mi, but it is the most plausible reason this could behave differently elsewhere and is worth stating rather than assuming.

## Root cause

`KafkaFaultInjector._run` (`sregym/generators/fault/inject_kafka.py:184-208`) execs the admin CLI inside the broker pod:

```python
command.extend([
    "-n", self.namespace,
    self._broker_pod_name(),
    "--",
    "env",
    "KAFKA_HEAP_OPTS=-Xms32m -Xmx64m",
    "KAFKA_OPTS=",
    f"{self._BIN_DIR}/{script}",
    *args,
])
```

`kubectl exec` runs the process in the target container's cgroup, so this JVM's memory counts against the broker's limit. In the vendored `opentelemetry-demo` chart, `components.kafka` is configured as:

```yaml
env:
  - name: KAFKA_HEAP_OPTS
    value: "-Xmx400M -Xms400M"
resources:
  limits:
    memory: 600Mi
```

There is **no `requests`** key. When a container declares limits but not requests, Kubernetes copies the limit into the request — so the broker is both capped and scheduled at 600Mi, holding a 400M heap that `-Xms400M` commits up front.

That leaves on the order of tens of megabytes free. A second JVM needs roughly 100–150Mi RSS even at `-Xmx64m`, because `-Xmx` bounds the heap and not metaspace, code cache, or thread stacks. It does not fit, and the kernel kills something.

This is deterministic rather than flaky: it failed on every attempt of every run we made, six for six. Both values are stock upstream chart config and have been unchanged for the problem's entire lifetime, so there is no regression window — we do not believe this problem has ever injected successfully.

## Why it is hard to see

Two things obscure the failure. Neither is the bug, but together they turn a deterministic OOM into a silent skip:

1. `KubeCtl.exec_command` (`sregym/service/kubectl.py:714-724`) catches `CalledProcessError` and **returns the stderr string** rather than raising, so the exit code never propagates. The injector compensates by appending `&& printf '\n__STREAM_ADMIN_OK__\n'` and testing for that marker (`inject_kafka.py:205-208`), which means an OOM kill is only ever detectable as a missing string. A stricter `exec_command_checked` exists at `sregym/service/kubectl.py:726` but is not used here.

2. After the third retry is exhausted, `main.py:393` logs and sets `result = None`, then cleanup runs and the driver continues to the next problem. The process exits 0. A caller that only checks the exit status sees success.

## Impact

- `kafka_poison_pill_hol_block` is in `PROBLEM_SETS` (`sregym/conductor/problem_sets.py:10`) and the registry (`sregym/conductor/problems/registry.py:228`), so the lite suite selects it and works through all three retries before skipping it.
- Each failed attempt tears the app down and redeploys it before the next try, so a full deploy/teardown cycle is spent three times over to reach a guaranteed skip.
- Any dataset or evaluation generated from these runs contains a labelled fault window with no fault in it.

## Suggested direction

Two options, not mutually exclusive:

1. **Give the broker's cgroup burst headroom.** `sregym/service/apps/values/astronomy-shop-fixes.yaml` is already applied by `AstronomyShop.deploy()`, so a `components.kafka.resources` override needs no Python change. Pinning `requests.memory` where it already effectively sits and raising only the limit leaves scheduling untouched.

2. **Stop exec-ing into the broker.** Run the admin CLI in a short-lived pod on the same Kafka image, pointed at `kafka:9092`. This removes the co-tenancy rather than budgeting around it, and makes the injector independent of the vendored chart's resource sizing. `inject_kafka.py` already creates its own Deployment for the pipeline workload, so the pattern exists in the same file.

## Related, filed separately if useful

- The same exec-a-heavyweight-process-into-an-app-container pattern appears in `inject_kafka.py:282-372` (CPython + `confluent_kafka`) and `sregym/generators/fault/inject_app.py:139-140` (several `mongo` shells). Neither is currently failing, but both depend on the target container having spare memory.
- `exec_command` swallowing exit codes (above) affects every caller, not just this injector.

dynomisto:
Fixes #1015

/validate-problem kafka_poison_pill_hol_block

## What

Adds a `components.kafka.resources` override to `sregym/service/apps/values/astronomy-shop-fixes.yaml`:

```yaml
components:
  kafka:
    resources:
      requests:
        memory: 600Mi
      limits:
        memory: 1Gi
```

One file, no Python change — `AstronomyShop.deploy()` already applies this values file via `-f` (`sregym/service/apps/astronomy_shop.py:16,56-57`).

## Why

`KafkaFaultInjector._run` (`inject_kafka.py:184-208`) execs `kafka-topics.sh` and `kafka-consumer-groups.sh` **inside the running broker container**, and `kubectl exec` runs them in that container's own cgroup.

The vendored `opentelemetry-demo` chart declares limits only for `components.kafka` (600Mi) with no `requests`, so Kubernetes copies the limit into the request: the broker is capped and scheduled at 600Mi while `-Xms400M` commits a 400M heap up front. A second JVM needs ~100–150Mi RSS even at `-Xmx64m`, because `-Xmx` bounds the heap and not metaspace, code cache or thread stacks. It does not fit, so the kernel OOM-kills it (exit 137) and takes the broker with it.

The result is that `kafka_poison_pill_hol_block` has never injected its fault. See the issue for the full diagnosis, including container-restart telemetry showing zero restarts while the broker serves traffic normally and then a restart in precisely the bucket of each of the three deploy attempts.

## Why this shape

- **Pin `requests`, raise only `limits`.** The container is already *effectively* requesting 600Mi via limit-copying, so declaring it explicitly changes nothing about scheduling. Raising only the cap makes the extra 424Mi burst capacity — occupied for the few seconds an admin CLI runs, not reserved. Bumping the limit alone would also raise the request and add 424Mi of scheduled memory per deployment.
- **Override at install time rather than edit the chart.** The chart lives two submodules down and both values are stock upstream config, so an in-place edit would mean a PR per repo and would be reverted by the next upstream merge.

## Verification

`helm template` with and without the override differs by exactly two lines, with the rendered object count unchanged:

```
 resources:                       resources:
   limits:                          limits:
     memory: 600Mi        ->          memory: 1Gi
                                    requests:
                                      memory: 600Mi
```

Re-running the problem on a live cluster:

| | Before | After |
|---|---|---|
| Injection | failed 3/3, exit 137 | succeeded on first attempt |
| `kafka` container restarts | 1 per attempt | **0 for the whole run** |
| Fault-window spans / logs | 0 / 0 | 17,501 / 47,790 |
| Quiet-gap spans / logs | — | 0 / 0 |
| Deploy attempts consumed | 3 (all failed) | 1 |

The injector reports success where it previously died:

```
== Fault Injection: Kafka poison-pill HOL block ==
Poison record at offset 20 of topic 'orders-fulfillment'; consumer group 'orders-validator' will stall there.
```

and the fault is observable for the first time:

- `orders-validator` logs `ERROR partition remains paused at offset=20` at a fixed interval — **123 messages spanning the full fault window**, with no self-recovery.
- `order-stream` logs **899 `published order_id=...`** records over the same window.

That pair is the head-of-line block made visible: the producer keeps publishing while the consumer is frozen at offset 20, so lag grows without bound. Both workloads were absent from every previous run, because injection never got far enough to create them.

This change does not touch `sregym/conductor/problems/`, so the validation gate does not require it, but the `/validate-problem kafka_poison_pill_hol_block` line above opts in anyway. Our evidence comes from a `kubeadm` cluster on EC2; running validation on the CI kind cluster confirms the fix independently and in a different environment.

## Follow-up, not in this PR

A more robust fix is to stop exec-ing into the broker at all and run the admin CLI in a short-lived pod on the same Kafka image, pointed at `kafka:9092`. That removes the co-tenancy rather than budgeting around it, and makes the injector independent of the vendored chart's resource sizing. `inject_kafka.py` already creates its own Deployment for the pipeline workload, so the pattern exists in the same file. Kept separate so this fix stays reviewable as a one-block change.

## Note on the file header

This PR also drops one sentence from the header of `astronomy-shop-fixes.yaml` — *"Nothing here changes what an agent can observe; each entry removes work the collector was doing that could never succeed."* That was accurate while the file held only the collector entry, but this change touches no collector work and does change what an agent can observe, since it is what makes the fault injectable. Happy to reinstate it with different wording if you would rather keep a summary line there.

### DIFF
--- a/sregym/conductor/oracles/kafka_producer_leak_mitigation.py
+++ b/sregym/conductor/oracles/kafka_producer_leak_mitigation.py
@@ -1,16 +1,25 @@
+import datetime
 import time
 
 from sregym.conductor.oracles.failure import FailureClass
 from sregym.conductor.oracles.mitigation import MitigationOracle
+from sregym.service.kafka_health import KafkaHealthCheck, broker_memory_failure, broker_pods
 from sregym.service.kubectl import ApiException, KubeCtl
 
 
 class KafkaProducerLeakOracle(MitigationOracle):
     FAILURE_CLASSES = {
         "kafka_still_restarting": FailureClass.AMBIGUOUS,
+        "kafka_not_serving": FailureClass.AMBIGUOUS,
     }
 
     def evaluate(self) -> dict:
+        try:
+            return self._evaluate()
+        except Exception as exc:
+            return self.fail_from_exception(exc)
+
+    def _evaluate(self) -> dict:
         self.rollout_time = 300
 
         results = super().evaluate()
@@ -36,20 +45,14 @@ class KafkaProducerLeakOracle(MitigationOracle):
 
             for c in kafka_deployment.spec.template.spec.containers:
                 if "kafka" in c.name:
-                    for e in c.env:
-                        if e.name == "KAFKA_HEAP_OPTS":
-                            if e.value != self.problem.heap_limit:
-                                print(f"❌ KAFKA_HEAP_OPTS is '{e.value}', expected '{self.problem.heap_limit}'")
-                                # Compared against the value the injector set,
-                                # so this is the fault observed directly.
-                                return self.fail(
-                                    "fault_still_present",
-                                    setting="KAFKA_HEAP_OPTS",
-                                    value=e.value,
-                                    expected=self.problem.heap_limit,
-                                )
-
-                            break
+                    heap = next((e.value for e in c.env or [] if e.name == "KAFKA_HEAP_OPTS"), None)
+                    if heap != self.problem.heap_limit:
+                        return self.fail(
+                            "fault_still_present",
+                            setting="KAFKA_HEAP_OPTS",
+                            value=heap,
+                            expected=self.problem.heap_limit,
+                        )
 
                     memory_limit = c.resources.limits.get("memory") if c.resources and c.resources.limits else None
                     if memory_limit != self.problem.memory_limit:
@@ -63,34 +66,44 @@ class KafkaProducerLeakOracle(MitigationOracle):
 
                     break
 
-            pods = kubectl.list_pods(self.problem.namespace)
-            rcnt_1 = None
-            for p in pods.items:
-                if "kafka" in p.metadata.name:
-                    for c in p.status.container_statuses:
-                        if "kafka" in c.name:
-                            rcnt_1 = c.restart_count
-                            break
-                    break
-
-            deadline = time.monotonic() + 120
-            while time.monotonic() < deadline:
-                pods = kubectl.list_pods(self.problem.namespace)
-                rcnt_2 = None
-                for p in pods.items:
-                    if "kafka" in p.metadata.name:
-                        for c in p.status.container_statuses:
-                            if "kafka" in c.name:
-                                rcnt_2 = c.restart_count
-                                break
+            with KafkaHealthCheck(kubectl, self.problem.namespace) as probe:
+                started = datetime.datetime.now(datetime.UTC)
+                # Kafka's container can be Ready before its protocol listener
+                # starts. Give a valid manual restart time to finish, then
+                # require the entire stability window below.
+                ready_deadline = time.monotonic() + 60
+                while True:
+                    if broker_memory_failure(kubectl, self.problem.namespace, started):
+                        return self.fail("fault_still_present", detail="Kafka has a new memory failure")
+                    if probe.check():
                         break
-
-                if rcnt_1 is None or rcnt_2 is None or rcnt_2 > rcnt_1:
-                    print(f"❌ kafka container restarted during the watch window ({rcnt_1} -> {rcnt_2})")
-                    # The broker is still OOMing even with the limits restored,
-                    # so the configuration is right but the outcome is not.
-                    return self.fail("kafka_still_restarting", before=rcnt_1, after=rcnt_2)
-
-                time.sleep(5)
+                    if time.monotonic() >= ready_deadline:
+                        return self.fail("kafka_not_serving")
+                    time.sleep(2)
+                before = self._broker_restarts()
+                deadline = time.monotonic() + 120
+                next_probe = 0
+                while True:
+                    after = self._broker_restarts()
+                    if not before or before != after:
+                        return self.fail("kafka_still_restarting", before=before, after=after)
+                    if broker_memory_failure(kubectl, self.problem.namespace, started):
+                        return self.fail("fault_still_present", detail="Kafka has a new memory failure")
+                    now = time.monotonic()
+                    if now >= next_probe or now >= deadline:
+                        if not probe.check():
+                            return self.fail("kafka_not_serving")
+                        next_probe = time.monotonic() + 20
+                    if now >= deadline:
+                        break
+                    time.sleep(min(5, max(0, deadline - time.monotonic())))
 
         return results
+
+    def _broker_restarts(self):
+        return {
+            str(pod.metadata.uid): status.restart_count
+            for pod in broker_pods(self.problem.kubectl, self.problem.namespace)
+            for status in pod.status.container_statuses or []
+            if status.name == "kafka"
+        }
--- a/sregym/generators/fault/inject_app.py
+++ b/sregym/generators/fault/inject_app.py
@@ -1,6 +1,8 @@
 """Inject faults at the application layer: Code, MongoDB, Redis, etc."""
 
 import base64
+import datetime
+import shlex
 import textwrap
 import time
 
@@ -8,6 +10,7 @@ from kubernetes import client
 
 from sregym.generators.fault.base import FaultInjector
 from sregym.service.apps.hotel_reservation import HOTEL_RESERVATION_APPLICATION_IMAGE
+from sregym.service.kafka_health import KafkaHealthCheck, broker_memory_failure
 from sregym.service.kubectl import KubeCtl
 
 FEATURE_FLAG_EXPERIMENTAL_ROUTING_IMAGE = HOTEL_RESERVATION_APPLICATION_IMAGE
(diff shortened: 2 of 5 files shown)


Saadmrp1038 (approved):

Saadmrp1038:
Interesting.
Even in our recent runs, it seemed that `600Mi` worked fine. Might be due to a setup difference(?)

dynomisto:
Maybe - I am currently working on AWS EC2 machines and I think you use Cloudlab. I'll investigate where that could introduce this issue

Saadmrp1038:
Yeah. It's a bit weird. Funnily with 1Gi limit the CloudLab setup gets timeout before memory limit is reached.
For now I think I will just increase the payload size so that it works on both setup.

dynomisto:
Ah - in my private implementation i have a baseline/warmup phase - i.e. the application runs for e.g. 1h without any problem injected. Then when the problem gets injected it crashes:


`The mechanism follows from the chart: KAFKA_HEAP_OPTS: -Xmx400M -Xms400M commits 400 MB at startup, but pages only become resident as they're touched, and a fixed-size heap never returns them. Cold, the broker has plenty of slack under the 600Mi cap and the admin CLI fits. After an hour of traffic the touched high-water mark is near the full heap, plus non-heap and log-segment page cache, so the cgroup sits at its limit and a second JVM can't start.`

Just a theory by claude code, not verified, but sounds plausible

dynomisto:
Another variant that would solve this would be to move the admin CLI into a short-lived pod and let the limit go back to 600Mi. This would make both problems work without tuning payloads etc

Saadmrp1038:
@dynomisto With the `1Gi` limit the `kafka_producer_leak` fault was getting timed out without producing the expected failure. I pushed fixes for that fault. I kept your memory-limit changes, so you should not face any issues anymore.

Saadmrp1038:
Thanks for reporting this!

Issue #1000, opened 2026-09-04
Harden agent container isolation? (unconditional ~/.aws mount, no cap-drop/non-root)


johannes-engler-mw:
Was reading through the agent container setup and it looks like the isolation is aimed at benchmark integrity (keeping the agent from finding the answers) rather than at protecting the host it runs on. #494 was about agents editing problem code, #911 stayed on the egress layer. Seems deliberate, so I wanted to ask before writing anything.

Two things stood out.

**`~/.aws` goes into every container.** `container_runner.py:558`:

```python
aws_dir = Path.home() / ".aws"
if aws_dir.is_dir():
    args.extend(["-v", f"{aws_dir.resolve()}:/root/.aws:ro"])
```

The only check is whether the directory exists, not whether the run uses Bedrock, not whether any AWS var is set. Run against OpenAI or a local Ollama and the agent still gets the whole directory, `sso/cache/` and `cli/cache/` included, which after an `aws sso login` means live tokens.

I get why it's there: the README points people at `AWS_PROFILE=bedrock`, and profile resolution needs the real files, not just the env var. It's the unconditional part that seems wider than you meant. Gating it on the run actually using AWS would be a few lines.

**No `--cap-drop`, no `--user`, no `no-new-privileges`.** Agents run as root with default caps. #911's evidence section actually mentions the missing `--cap-drop`, but that issue ended up scoped to networking and it never got picked up separately. Not asking for gVisor here, just the cheap flags.

So:

1. Is host-side hardening something you want, or is "untrusted about the benchmark, trusted about the machine" the intended model? Fine either way, just worth writing down somewhere, since people run this on laptops with real cloud creds.
2. Would you take a PR narrowing the `~/.aws` mount? That one seems worth doing on its own.
3. Any reason `--cap-drop=ALL` or non-root would break the agent installers or kubectl? The install scripts run at container start, so non-root might mean moving that into the image build. Happy to test if you haven't already ruled it out.

Glad to do either one if it's useful.

tianyin:
@johannes-engler-mw thanks for the comment!

You are precise about our current trust model ("keeping the agent from finding the answers"). 

Most of our experiments were not done on a personal laptop so we didn't put too many thoughts there.

I would love to tighten the interface with the host. With efforts like SREGym-Lite, it may be the case that people are trying it out on laptops or PCs. As an OS person, it feels bad to hear --cap-drop and --no-new-priv were ignored. 

[...]

PR #1017, opened 2026-09-08
Mount ~/.aws only when the run uses AWS

johannes-engler-mw:
Refs #1000

## Problem

`container_runner.py` mounted `~/.aws` into every agent container whenever the
directory existed. The only check was `aws_dir.is_dir()`, never whether the run
used AWS. Run against OpenAI or a local Ollama and the agent still got the whole
directory, including `sso/cache/`, which after an `aws sso login` means live
tokens.

## Change

Mount only when the run actually resolves AWS credentials: a credential-selecting
`AWS_*` variable (profile, access key, role, web identity, Bedrock bearer token),
or an AWS provider in `AGENT_MODEL_ID` (`bedrock`, `amazon-bedrock`, `sagemaker`).

Region variables don't count. They say where to call, not which identity to call
with. The model-id check matters on its own, since a default profile in
`~/.aws/config` sets no `AWS_*` variable.

`_build_base_docker_args` now takes `extra_env` so variables arriving through
`ExecInput` are visible to the check, the same way `_build_env_flags` sees them.

## What this doesn't fix

For runs that do use Bedrock, `sso/cache/` is still exposed. Profile resolution
needs it, so trimming the mount to `config` + `credentials` would break SSO.

`AWS_ACCESS_KEY_ID` still triggers the mount even though a complete key pair
authenticates on its own. I kept it because `~/.aws/config` can still hold
`role_arn` for assume-role chains, but I can tighten it if you prefer.

## Testing

I don't have a Bedrock account for a live call, which is why I checked it via tests.
If you want a real run before merging, that needs someone with credentials.

---

https://claude.ai/code/session_01AY4TR63izhxgG85b5Lxzqj


HacksonClark (approved):
I was able to test the AWS changes and it worked great!

PR #1018, opened 2026-09-08
Drop container capabilities and set no-new-privileges

johannes-engler-mw:
Refs #1000

## Problem

Agent containers run as root with Docker's default 14 capabilities and no
`no-new-privileges`. #911's evidence section mentioned the missing `--cap-drop`,
but that issue stayed scoped to networking and this never got picked up.

## Change

```python
HARDENING_FLAGS = (
    "--cap-drop=ALL",
    "--cap-add=DAC_OVERRIDE",
    "--security-opt=no-new-privileges",
)
```

Applied in `_build_base_docker_args`, behind `--container-hardening {on,off}`
(default `on`). That leaves `cap_dac_override` alone in place of the default 14,
dropping `setuid`, `setgid`, `net_raw`, `mknod`, `sys_chroot`, `setfcap` and the
rest. The flag is plumbed through `AgentLauncher` the way `--internet-access` is,
and the chosen mode shows up in the logged config line.

Choices rather than a `--no-` flag for two reasons: the mode is then visible in
the run's config log, and a stricter mode can be added later without renaming
the flag once agents run non-root.

## Why DAC_OVERRIDE stays

`--cap-drop=ALL` by itself breaks the harness. Measured on Ubuntu 24.04 /
x86_64 against `sregym-agent-base`:

```
cap-drop=ALL                  FAIL  tee: /logs/install.log: Permission denied
cap-drop=ALL +DAC_OVERRIDE    PASS
```

`/logs` and `/workspace` are bind mounts owned by the host user (mode 775).
Container root matches neither owner nor group, and `other` has no write bit, so
the first thing `build_composite_command` does fails without it.

Keeping it is a downside worth being clear about: for a root process
`DAC_OVERRIDE` bypasses all file permission checks, so this drops 13 of the 14
rather than all of them. Removing it means running non-root.

Nothing else needed a capability. Both npm installers finish with `CapEff 0`,
kubectl works under the full set, and `NET_RAW` is irrelevant because the image
ships no `ping`, `traceroute`, `nc` or `dig`.

## What `--container-hardening off` is for

Agents that install tooling mid-run. Measured on `ubuntu:24.04` / arm64:

```
cap-drop=ALL +DAC_OVERRIDE                 apt-get update  exit 100
  E: setgroups 65534 failed (Operation not permitted)
  E: seteuid 42 failed (Operation not permitted)
cap-drop=ALL +DAC_OVERRIDE +SETUID +SETGID apt-get update  exit 0
unhardened                                 apt-get update  exit 0
```

So it is `CAP_SETUID`/`CAP_SETGID` that `apt-get` needs in order to drop to
`_apt`, not `no-new-privileges`, and it fails at `update` rather than at
`install`.

## Gaps

- `_ensure_filtered_egress` does its own `docker run` for mitmproxy and doesn't
  go through `_build_base_docker_args`, so under `--internet-access filtered`
  that container is still unhardened. This PR doesn't fix it.
- Non-root would remove the need for `DAC_OVERRIDE`, but it means moving the
  agent installers into the image build and fixing ownership on `/logs` and
  `/workspace`. Not doing that here.

## Testing

`tests/service/test_container_hardening.py`, 10 cases: flags present by default,
absent when disabled, applied under both internet policies, placed before the
image argument (Docker ignores run options that come after it), and the launcher
plumbing — default on, both flag values reaching `ContainerConfig`, and no
change once the runner exists.

I also ran the real generated command on a Linux box, built through
`ContainerRunner` itself with the mounts created by the actual code path. It
passes with the flags on and with them off. Not covered: a full problem run end
to end, so this checks the container mechanics rather than an agent solving
something.

---

https://claude.ai/code/session_01AY4TR63izhxgG85b5Lxzqj
https://claude.ai/code/session_011oVzeQyEcLCP5eYM6gDL9Y


Saadmrp1038:
@johannes-engler-mw Thanks for the PR!

> An agent that tries to `apt-get` install something mid-run may now fail. I think that's fine for a benchmark, but it's a behaviour change, so your call.

I feel this hardening should be controllable through a flag for this reason. If some users want to use agents that can install tooling during runtime they might want `apt-get` to work. They can just set the param according to their need. Default should be hardened container.

> that container is still unhardened. This PR doesn't fix it.

It should be fine.

> Non-root would remove the need for `DAC_OVERRIDE`

I don't think it's necessary at this moment.

johannes-engler-mw:
Good catch, the knob existed as `ContainerConfig.harden_container`, but nothing wired it to the CLI, so you could only reach it by editing the code. Fixed in a5a03dbd: `--container-hardening {on,off}`, default `on`, plumbed through `AgentLauncher` the same way `--internet-access` is.

Writing the help text made me check what actually breaks, and my "may fail" above was too soft. On `ubuntu:24.04`:

```
cap-drop=ALL +DAC_OVERRIDE                  apt-get update  exit 100
cap-drop=ALL +DAC_OVERRIDE +SETUID +SETGID  apt-get update  exit 0
```

It's `CAP_SETUID`/`CAP_SETGID` that `apt-get` needs to drop to `_apt`, and it dies at `update`, not at `install`. That's in the flag help and the README now.

Left mitmproxy and non-root alone, per your call on both.

Saadmrp1038:
LGTM
Had to add `TAR_OPTIONS=--no-same-owner` in curson installation command. Otherwise, it was failing.

Saadmrp1038 (approved):

PR #1019, opened 2026-09-08
Write the local provider config for OpenCode preflight

johannes-engler-mw:
## Problem

OpenCode has no built-in `local` provider, so a `local/<model>` run needs a
generated config declaring one. `OpenCodeAgent` writes that config,
`run_preflight` didn't. So `--model local/...` failed preflight with an opaque
server error even though the run itself would have worked.

## Change

Pull the config writing into `write_local_provider_config` and call it from
both. The preflight has no logs dir, so it writes to a temp dir and cleans up in
a `finally`. Nothing changes for the agent run.

## Testing

Ran a full `local/` model run on a Linux VM.

Also added `tests/clients/test_opencode_local_provider.py`, since there was no
coverage of this before. Five cases, including that the API key is written as an
`{env:...}` reference rather than inlined, because the config ends up in the
logs dir.


Saadmrp1038 (approved):

Saadmrp1038:
LGTM
Thanks for the fix!

PR #1020, opened 2026-09-08
Detect architecture when installing kubectl

johannes-engler-mw:
## Problem

The kubectl download hardcoded `linux/amd64`, so building the image on ARM gave
you an x86 kubectl. ARM is otherwise supported, `install-codex.sh` already picks
its platform package based on `Linux/aarch64`.

## Change

Use `dpkg --print-architecture`, which prints the same names kubectl publishes
under.

## Testing

Checked both ends: `ubuntu:22.04` reports `amd64` and `arm64` under the matching
platforms, and the v1.37.0 kubectl URLs for both return 200.


Saadmrp1038:
Works great! Thanks for fixing this.

Saadmrp1038 (approved):

PR #1024, opened 2026-09-09
Improve deployment rollout checkes

Saadmrp1038:
Some mitigation oracles accepted incomplete Deployment rollouts. For example, one new pod was Ready while an extra old pod remained. The generic oracle also accepted stalled rollouts after its wait timed out because its final check only counted Ready replicas.

This PR replaces duplicated rollout checks with one shared function across mitigation grading, setup, recovery, and infrastructure readiness. It requires the controller to observe the latest Deployment generation. Total, updated, Ready, and Available replica counts must match the desired count, with no unavailable replicas.

The final oracle checks enforce this rule even after a rollout timeout (previously it would just continue).

/validate-problem skip


tianyin (approved):
LGTM (need to resolve conflicts)

PR #1025, opened 2026-09-10
Add isolated Docker-in-Docker runtime for parallel SREGym runs

HacksonClark:
SREGym's fixed cluster names, ports and Docker state prevent independent problems from sharing a host safely. This adds an experimental Docker-in-Docker runtime: each outer container owns a Docker daemon, the existing four-node KIND/Calico cluster, nested agent containers and its own results directory. Separate launcher invocations can run problems concurrently.

- Add a build/run launcher, cgroup delegation, readiness checks, bounded shutdown, startup diagnostics and a live smoke test. Use a newer KIND node base for DinD while preserving the existing host setup's default.
- Persist results through a bind mount. Stage temporary CSVs and opaque run artifacts on that same filesystem so atomic publication avoids cross-device rename failures.
- Skip the unused diagnosis-judge preflight for mitigation-only campaigns, enabling credential-free autosubmit/noop runs.
- Document parallel execution, resource requirements, optional memory-backed Docker storage, and the integration boundary for Harbor. This supplies the runtime building block; a complete Harbor task adapter remains separate work.

Validation:

- Nine launcher tests and six campaign/publication/preflight tests passed; shell syntax, Ruff and diff checks passed.
- Concurrent environments each reached four Ready nodes, with distinct Docker daemon IDs and Kubernetes namespace UIDs. Nested mounts, networking, Kubernetes service connectivity and workload recovery passed smoke checks.
- Real fault/recovery oracle lifecycles passed for `wrong_service_selector_hotel_reservation` (disk-backed Docker) and `network_policy_block` (memory-backed Docker).
- A complete `autosubmit --stages mitigation` campaign finished with `run_status=complete`, expected `Mitigation.success=False`, published agent artifacts and per-problem/aggregate CSVs, without API keys.

Validation used an ARM64 Ubuntu host with QEMU for AMD64 application images. Initial cold deployments on its slow disk timed out; successful retries and the memory-backed run are described in the guide. Privileged Linux containers share the host kernel and retain existing KIND problem limitations. Default tmpfs etcd and optional memory-backed Docker change storage behavior; disk/persistence fault evaluations need disk-backed settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,9 @@ SREGym runs on a self-managed Kubernetes cluster that you provision on Linux hos
 ### b) Emulated cluster
 SREGym can be run on an emulated cluster using [kind](https://kind.sigs.k8s.io/) on your local machine. However, not all problems are supported.
 
+For an experimental Docker-in-Docker environment with a private cluster per run,
+including parallel problem execution, see the [DinD guide](./docker/dind/README.md).
+
 **Note:** If you run into pod crashes or "too many open files" errors, see the [kind README](./kind/README.md) for required host kernel settings and troubleshooting.
 
 ```bash
--- a/clients/claudecode/driver.py
+++ b/clients/claudecode/driver.py
@@ -46,13 +46,23 @@ def run_preflight() -> None:
     if reasoning_effort:
         command.extend(["--effort", reasoning_effort])
 
-    r = subprocess.run(
-        command,
-        capture_output=True,
-        text=True,
-        timeout=60,
-        env=env,
-    )
+    timeout_seconds = 150
+    try:
+        r = subprocess.run(
+            command,
+            capture_output=True,
+            text=True,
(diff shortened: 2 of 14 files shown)


Saadmrp1038 (approved):

Saadmrp1038:
Works great!
> This adds an experimental Docker-in-Docker runtime

Yup. I would treat this as an experimental feature right now. During testing found some non-blocking issues:
1. Nested node reports the host's CPU and Memory (56 CPUs, 252 GB) instead of actual environment limits (8 CPUs, 16 GB).
2. Not really a bug but the cold start is really slow. Application deployment took approximately 30 minutes for the first faults of runs. We can possibility of make this faster (needs some rnd) later on.

The containers also share host level kernel. But I think all kind-compatible faults should work fine in DinD settings. 

---

Doing some full runs should help surface any hidden bugs.

PR #1027, opened 2026-09-10
Align token metrics across client results and ATIF exports

Saadmrp1038:
The same run reported different token totals in its client results and ATIF export. The two implementations counted cached input and reasoning differently.

This PR gives both outputs consistent definitions. Input totals include cache reads and writes. Output totals include reasoning. Separate fields retain these breakdowns without counting them twice.

It also corrects duplicate usage records and reads usage from additional native session formats. Missing usage remains unknown instead of appearing as zero.


tianyin (approved):
LGTM

PR #1028, opened 2026-09-10
Apply deployment profiles and validate infrastructure reuse

Saadmrp1038:
This PR fixes some issues in infrastructure setup, reuse, and cleanup.

Previously, switching between `full` and `svelte` could leave Prometheus on the previous profile. Prometheus now checks its installed configuration before reuse and redeploys when the profile differs. Svelte also removes `openebs-device` when it is absent from the original cluster baseline. Results and startup logs now record the selected profile.

Startup now checks the metrics API before reusing metrics-server. Full mode also requires a complete, nonempty OpenEBS NDM rollout. Cleanup preserves infrastructure by exact resource type and name instead of matching name fragments.

Helm now reads local dependency entries correctly and stops installation when a dependency update fails. Complete local charts avoid unnecessary dependency downloads.

### DIFF
--- a/main.py
+++ b/main.py
@@ -32,7 +32,7 @@ from sregym.conductor.constants import StartProblemResult
 from sregym.conductor.problem_sets import PROBLEM_SETS
 from sregym.phases import read_ledger as read_phase_ledger
 from sregym.phases import results_columns as phase_results_columns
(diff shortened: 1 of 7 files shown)


tianyin (approved):
LGTM

PR #1031, opened 2026-09-12
Make SREGym ARM compatible

### LINKED ISSUE #1001: ARM64: setup_kind_cluster.sh arm is documented but most app images are amd64-only
The repo documents an ARM64 path (`kind/kind-config-arm.yaml`, `bash kind/setup_kind_cluster.sh arm` in `docs/SREGym-Lite.md`, `aarch64` detection in `install-codex.sh`), but most application images are amd64-only, so a run on ARM gets a long way in and then fails with nothing to go on.

On an ARM64 VM the cluster comes up fine, all four nodes Ready. Then:

```
❌ start_problem failed for 'network_policy_block' (deploy attempt 1/3):
Timeout: Not all pods in namespace 'hotel-reservation' reached the Ready state within 600 seconds.
```

Nothing there points at architecture, and each attempt burns the full 600s before the retry tears the namespace down and removes the evidence. What is actually happening:

```
$ docker image inspect ghcr.io/sregym/hotel-reservation:latest --format '{{.Os}}/{{.Architecture}}'
linux/amd64

$ docker run --rm --entrypoint /bin/sh ghcr.io/sregym/hotel-reservation:latest -c 'echo ok'
exec /bin/sh: exec format error
```

The image is a single-platform manifest, so the pull succeeds with no platform check and the failure only surfaces when a container execs. Kind nodes are containers themselves, so there is no binfmt emulation to fall back on. It fails the same way on an ARM Mac.

Same for `deathstarbench/social-network-microservices` and `ghcr.io/sregym/train-ticket-deploy`. Astronomy Shop is the exception, since upstream OpenTelemetry publishes multi-arch, so ARM today is effectively astronomy-shop only. #584 reported this for the train-ticket image and was closed, but the other images have the same problem.

Separately, `docker/agents/Dockerfile:42` hardcodes `bin/linux/amd64/kubectl`, so even the agent image gets an unrunnable kubectl on ARM. Small fix (`dpkg --print-architecture` returns exactly the tokens k8s uses and is a no-op on x86), happy to send it as a PR.

Is publishing arm64 variants of `ghcr.io/sregym/*` on the table? Those are your own builds, unlike the deathstarbench ones. If not, could the docs say plainly that the ARM path runs astronomy-shop only? Right now `setup_kind_cluster.sh arm` reads as full support. Even an image-architecture check at deploy time would turn this from a confusing timeout into a clear error. Glad to help either way.

Saadmrp1038:
closes #1001 

### Summary

SREGym used several AMD64-only images and architecture-specific setup paths. These dependencies prevented the benchmark from operating correctly on ARM64 hosts.

This PR publishes the required custom images for both `linux/amd64` and `linux/arm64`. Application manifests now use public, digest-pinned images from SREGym GitHub Packages.

### Changes

Added native AMD64 and ARM64 support for the agent environment, application services, workloads, storage components, and fault images. These images are published through SREGym GitHub Packages and pinned by digest for reproducible runs.

Updated application manifests, workloads, noise generators, and fault injectors to use the new multi-architecture images. Image-replacement recovery now restores the image captured before injection instead of using a hard-coded image.

The related `SREGym-applications` changes (https://github.com/SREGym/SREGym-applications/pull/9) add multi-architecture support for Hotel Reservation, Social Network, Blueprint, FleetCast, FlightTicket, TrainTicket, and OpenWhisk. The FleetCast branch pins compatible backend dependencies. The FlightTicket branch adds multi-architecture OpenWhisk components, packages actions during image builds, and fixes Redis, gateway, and event-provider compatibility.

## Testing

I ran the lite problem suite along with 7 additional faults, for a total of 28 faults, across real clusters and KIND ones. I also ran the applications on an ARM machine and found no issues.

---

/validate-problem skip


tianyin (approved):
This is a huge PR. It is very well tested. I approve it.

PR #1032, opened 2026-09-15
Support existing subscriptions for the LLM judge

### LINKED ISSUE #1030: Support existing agent subscriptions for the LLM judge
SREGym already supports subscription access for agent runs through Codex, Claude Code, and GitHub Copilot. I think it would be useful to support those subscriptions for the LLM judge too, alongside the current API backend.

This would let users run evaluations without setting up separate API billing and help us developers use our existing subscriptions for the judge during development and testing.

Does this make sense? I can work on it if others think it would be useful.

munimthahmid:
People who already have a subscription can use it for the judge without setting up separate API billing.

**What changed**

- Add `--judge-backend` for Codex, Claude Code, Copilot, and Cursor, independently of the agent.
- Keep API as the default, with the same judge rubric and scoring.
- Reuse the existing container runner and Cursor bridge, keeping the old Cursor command working.
- Use `CODEX_HOME` to select the Codex login.

**Testing**

- Live Docker judge checks passed with Codex, Copilot, and Cursor on correct and incorrect diagnoses.
- Claude Code still needs an authenticated live check because I don’t have subscription access.

Closes #1030.


HacksonClark:
This is a very nice QoL improvement :)

Saadmrp1038:
Works great!

Saadmrp1038 (approved):

